Issue #17, opened 2025-02-21
Resources and prompts cannot be added after server start


marianogonzalez:
**Bug description**
The request handlers for resources, templates and prompts are only added if any are provided at the moment of building the server. Although methods like addResource() do function as expected, request with methods such as `resources/list` or `resources/read` fail with no such method errors because the handlers are not there, even if the server does support the capability. 

**Environment**
Java SDK, async server

**Steps to reproduce**

Create a server like this:
```
return McpServer.async(createTransport())
          .serverInfo(config.getServerName(), config.getServerVersion())
          .capabilities(new ServerCapabilities(
                                               emptyMap(),
                                               null,
                                               null,
                                               new ResourceCapabilities(
                                                                        config.getResourceCapabilities().isSubscribe(),
                                                                        config.getResourceCapabilities().isListChanged()),
                                               new ServerCapabilities.ToolCapabilities(config.getToolCapabilities()
                                                   .isListChanged())))
          .build();
```

Afterwards, add a resource like this:

```
server.addResource(new AsyncResourceRegistration(
new McpSchema.Resource(uri, name, description, mimeType, null), mono -> {....});
```

**Expected behavior**
The resource can be accessed and listed after being added

jwflicker:
is there a workaround for this?

marianogonzalez:
@jwflicker I sent PR #18 that fixes the problem. You can apply that fix locally while waiting for it to be merged

marianogonzalez:
Fixed in PR #18

PR #18, opened 2025-02-21
Resources and prompts cannot be added after server start 

marianogonzalez:
fixes issue #17

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
@@ -138,18 +138,14 @@ public class McpAsyncServer {
 		}
 
 		// Add resources API handlers if provided
-		if (!Utils.isEmpty(this.resources)) {
+		if (serverCapabilities.resources() != null) {
 			requestHandlers.put(McpSchema.METHOD_RESOURCES_LIST, resourcesListRequestHandler());
 			requestHandlers.put(McpSchema.METHOD_RESOURCES_READ, resourcesReadRequestHandler());
-		}
-
-		// Add resource templates API handlers if provided.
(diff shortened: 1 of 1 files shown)


mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -138,18 +138,14 @@ public class McpAsyncServer {
 		}
 
 		// Add resources API handlers if provided
-		if (!Utils.isEmpty(this.resources)) {
+		if (serverCapabilities.resources() != null) {
```
tzolov, line 141: please add `this.` prefix.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -138,18 +138,14 @@ public class McpAsyncServer {
...
-		}
-
-		// Add resource templates API handlers if provided.
-		if (!Utils.isEmpty(this.resourceTemplates)) {
 			requestHandlers.put(McpSchema.METHOD_RESOURCES_TEMPLATES_LIST, resourceTemplateListRequestHandler());
 		}
 
 		// Add prompts API handlers if provider exists
-		if (!Utils.isEmpty(this.prompts)) {
+		if (serverCapabilities.prompts() != null) {
```
tzolov, line 148: please add `this.` prefix.

tzolov (changes requested):
@marianogonzalez, thank you catching this issue. 
Could you please fix the prefixes as requested.
Also can you add the `serverCapabilities.resources() != null` for the resource templates as well (line 147).

marianogonzalez:
Thank you for reviewing @tzolov . Added the requested changes.

PR #27, opened 2025-03-01
refactor(mcp): remove redundant type field from Content implementations

### LINKED ISSUE #26: Improve MchSchema JSON consistency
## Bug description
Redundant type handling in McpSchema Content implementations causes unnecessary code complexity and potential maintenance issues. 
The Content interface and its implementations (TextContent, ImageContent, EmbeddedResource) currently maintain explicit type fields and methods, despite Jackson already handling type information through @JsonSubTypes annotation. 
This redundancy increases code complexity and creates potential for inconsistencies if the explicit type values don't match the expected polymorphic types.

## Environment

Java SDK version: 17
MCP Java SDK version: 0.8.0-SNAPSHOT

## Steps to reproduce

```java
	@Test
	void testTextContent() throws Exception {
		McpSchema.TextContent test = new McpSchema.TextContent("XXX");
		String value = mapper.writeValueAsString(test);
		assertEquals("""
				{"type":"text","text":"XXX"}""", value);
	}
```

## Solution:

Lets Jackson handle the type information entirely and removes the redundant field and method:

- Remove the `type()` method from `Content` interface
- Removing the `type` field, constructor initialization, and getter method from all implementations
- Adjusting constructors to no longer require the type parameter
- Ensuring tests pass, confirming that JSON serialization still includes the correct type information

tzolov:
- The type field and associated methods were redundant in Content implementations (TextContent, ImageContent, EmbeddedResource) as the type information is already handled by Jackson's polymorphic type handling via @JsonSubTypes annotation.
- Added comprehensive unit tests for McpSchema to validate serialization/deserialization behavior of all schema components.

Resolves #26
Resolve https://github.com/spring-projects/spring-ai/issues/2350

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -9,6 +9,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
@@ -964,7 +965,7 @@ public final class McpSchema {
 			@JsonSubTypes.Type(value = EmbeddedResource.class, name = "resource") })
 	public sealed interface Content permits TextContent, ImageContent, EmbeddedResource {
 
-		String type();
+		// String type();
 
 	}
 
@@ -972,19 +973,10 @@ public final class McpSchema {
 	public record TextContent( // @formatter:off
 		@JsonProperty("audience") List<Role> audience,
 		@JsonProperty("priority") Double priority,
-		@JsonProperty("type") String type,
 		@JsonProperty("text") String text) implements Content { // @formatter:on
 
-		public TextContent {
-			type = "text";
-		}
-
-		public String type() {
-			return type;
-		}
-
 		public TextContent(String content) {
-			this(null, null, "text", content);
+			this(null, null, content);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -964,61 +965,34 @@ public record CompleteCompletion(// @formatter:off
 			@JsonSubTypes.Type(value = EmbeddedResource.class, name = "resource") })
 	public sealed interface Content permits TextContent, ImageContent, EmbeddedResource {
 
-		String type();
+		// String type();
```
chemicL, line 968: It should be removed I think.

chemicL (changes requested):
Despite the triviality of the fix, the tests are comprehensive and useful so I reviewed them carefully. I left a few comments.

mcp/src/test/java/io/modelcontextprotocol/spec/McpSchemaTests.java
```diff
@@ -0,0 +1,458 @@
...
+
+		McpSchema.Implementation clientInfo = new McpSchema.Implementation("test-client", "1.0.0");
+
+		McpSchema.InitializeRequest request = new McpSchema.InitializeRequest(McpSchema.LATEST_PROTOCOL_VERSION,
+				capabilities, clientInfo);
+
+		String value = mapper.writeValueAsString(request);
+		assertEquals(
+				"""
+						{"protocolVersion":"2024-11-05","capabilities":{"roots":{"listChanged":true},"sampling":{}},"clientInfo":{"name":"test-client","version":"1.0.0"}}""",
```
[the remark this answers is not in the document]
tzolov: you're right. I will remove the constant from the fixture.

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
tzolov: Actually I've decided to add a `default String type()` implementation to comply with the MCP Schema, were the type in the content exist.

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -964,61 +964,45 @@ public record CompleteCompletion(// @formatter:off
 			@JsonSubTypes.Type(value = EmbeddedResource.class, name = "resource") })
 	public sealed interface Content permits TextContent, ImageContent, EmbeddedResource {
 
-		String type();
+		default String type() {
```
chemicL, line 967: Since you are dealing with a sealed interface, you can use a [pattern switch](https://docs.oracle.com/en/java/javase/17/language/pattern-matching-switch.html) on `this` and have a complete coverage without the default case.

tzolov: pattern matching is available in Java 21+. The MCP project at the moment is configured for 17.

tzolov:
Addressed review comments

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
chemicL: Oh, it was a preview feature back then. My bad, sorry!

chemicL (approved):

PR #31, opened 2025-03-05
Refactor server side to handle multiple clients

chemicL:
This PR introduces an internal concept of a server session maintained with a particular client alongside with a user-facing concept of an exchange, which provides means to communicate with the client as part of a server handler.

We are trying to avoid breaking changes by introducing deprecations with occasional breakages for those enhancing the core SDK. These are listed in the migration document.


tzolov:
squashed and merged at f50f88ce60fad913cf62db7ce772a6073d4fc973

PR #39, opened 2025-03-11
feat(client): Improve initialization state handling in McpAsyncClient

tzolov:
- Add proper initialization state tracking using AtomicBoolean and Sinks
- Implement timeout handling for requests requiring initialization
- Ensure all client methods verify initialization state before proceeding
- Fix rootsListChangedNotification to check initialization state
- Improve error messages for uninitialized client operations

### DIFF
--- a/mcp-test/src/main/java/io/modelcontextprotocol/client/AbstractMcpAsyncClientTests.java
+++ b/mcp-test/src/main/java/io/modelcontextprotocol/client/AbstractMcpAsyncClientTests.java
@@ -10,6 +10,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 
 import io.modelcontextprotocol.spec.ClientMcpTransport;
+import io.modelcontextprotocol.spec.McpError;
 import io.modelcontextprotocol.spec.McpSchema;
 import io.modelcontextprotocol.spec.McpSchema.CallToolRequest;
 import io.modelcontextprotocol.spec.McpSchema.ClientCapabilities;
@@ -47,7 +48,7 @@ public abstract class AbstractMcpAsyncClientTests {
 
 	protected ClientMcpTransport mcpTransport;
 
-	private static final Duration TIMEOUT = Duration.ofSeconds(20);
+	// private static final Duration TIMEOUT = Duration.ofMillis(1000);
 
 	private static final String ECHO_TEST_MESSAGE = "Hello MCP Spring AI!";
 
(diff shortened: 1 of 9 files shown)


mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -335,6 +346,26 @@ public Mono<Void> closeGracefully() {
...
+
+	/**
+	 * Utility method to handle the common pattern of checking initialization before
+	 * executing an operation.
+	 * @param <T> The type of the result Mono
+	 * @param errorMessage The error message to use if the client is not initialized
+	 * @param operation The operation to execute if the client is initialized
+	 * @return A Mono that completes with the result of the operation
+	 */
+	private <T> Mono<T> withInitializationCheck(String errorMessage,
```
chemicL, line 361: "Client must be initialized before " is repeated in each use, consider replacing `errorMessage` with `action` and just specify what's about to happen, e.g. `action="pinging the server"`.

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -335,6 +346,26 @@ public Mono<Void> closeGracefully() {
 		return this.mcpSession.closeGracefully();
 	}
 
+	// --------------------------
+	// Utility Methods
+	// --------------------------
```
chemicL, line 351: It's a not some "utility" but a fundamental prerequisite to almost any operation, perhaps we can group it under "initialization methods"?

chemicL (commented):
Left a couple suggestions, but overall this is a useful improvement :)

tzolov:
Thanks for the feedback @chemicL . I've tried to address the review comments and to improve the documentation.

mcp-test/src/main/java/io/modelcontextprotocol/client/AbstractMcpAsyncClientTests.java
```diff
@@ -92,8 +96,16 @@ void testConstructorWithInvalidArguments() {
 			.hasMessage("Request timeout must not be null");
 	}
 
+	@Test
+	void testListToolsWithoutInitialization() {
+		assertThatThrownBy(() -> mcpAsyncClient.listTools(null).block()).isInstanceOf(McpError.class)
```
chemicL, line 99: The test will last at least 600ms in case of `WebClient` and 4s for JDK `HttpClient` before the exception is observed, correct? If I am correct, it would be useful to consider using StepVerifier.withVirtualTime to emulate time passing by instead of adding more seconds to the time w validate the behaviour.

PR #51, opened 2025-03-19
refactor: improve MCP client timeout handling and reactive testing

tzolov:
- Add configurable initialization timeout separate from request timeout
- Rename ServletSse* test classes to HttpSse* for better naming consistency
- Replace direct .block() calls with StepVerifier for better reactive testing
- Improve error handling and reactive programming patterns throughout tests
- Chain reactive operations for cleaner test flow


chemicL (changes requested):
Thanks, this will help a bit. Next step is to incorporate the VirtualTimeScheduler to avoid actually timing out after 1s but immediately doing so. I noticed some breaking changes, perhaps it's best to introduce them in a separate PR to highlight in the release notes easier?

tzolov:
Thanks for the suggestions @chemicL . Last commit addresses them all.

chemicL (approved):

PR #52, opened 2025-03-19
Improve client test reliability and execution time

chemicL:
This change uses `VirtualTimeScheduler` and pretends enough time has passed to trigger a timeout on the initialization.

Another problem with reliability of the tests was that the used testcontainer for the SSE server does not support multiple clients and the existence of both the global client for the entire suite and some customized local clients in some tests caused responses to be delivered to the other client at some racing situations. Now each test creates a dedicated client and performs cleanup locally.

While these tests were improved, two other issues were found and fixed:
* The first one is that the `closeGracefully` of `DefaultMcpSession` was not lazy and would trigger connection disposal before the returned `Mono` was subscribed. 
* The second one was dealing with closing the StdIo client before the process was started. In such a case there should not be an error but rather a warning and successful completion.


tzolov (approved):
Looks great! Thanks @chemicL 
Nice catch for the closeGracefully and the StdIo closing.

PR #66, opened 2025-03-22
refactor: remove deprecated 0.7.0 code

tzolov:
These changes are part of the planned deprecation cycle announced in 0.8.0, with the deprecated classes scheduled for removal in 0.9.0

- Delete WebFluxSseServerTransport, WebMvcSseServerTransport, StdioServerTransport, and HttpServletSseServerTransport
- Remove deprecated interfaces: ServerMcpTransport, ClientMcpTransport
- Delete DefaultMcpSession implementation
- Remove all deprecated test classes for the removed implementations
- Update references to use McpServerTransport and McpClientTransport interfaces


mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
```diff
@@ -71,7 +71,7 @@ public class McpSyncClient implements AutoCloseable {
 	 */
 	@Deprecated
```
tzolov, line 72: Looks like this deprecation annotation is incorrect

mcp/src/test/java/io/modelcontextprotocol/MockMcpTransport.java
```diff
@@ -13,17 +13,17 @@
...
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.Sinks;
 
 /**
- * A mock implementation of the {@link McpClientTransport} and {@link ServerMcpTransport}
+ * A mock implementation of the {@link McpClientTransport} and {@link McpServerTransport}
  * interfaces.
  */
-public class MockMcpTransport implements McpClientTransport, ServerMcpTransport {
+public class MockMcpTransport implements McpClientTransport, McpServerTransport {
```
tzolov, line 26: Looks like we need to split the MockMcpTransport to client and server mocks.

mcp-test/src/main/java/io/modelcontextprotocol/MockMcpTransport.java
```diff
@@ -15,15 +15,15 @@
...
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.Sinks;
 
 /**
- * A mock implementation of the {@link McpClientTransport} and {@link ServerMcpTransport}
+ * A mock implementation of the {@link McpClientTransport} and {@link McpServerTransport}
  * interfaces.
  */
-public class MockMcpTransport implements McpClientTransport, ServerMcpTransport {
+public class MockMcpTransport implements McpClientTransport, McpServerTransport {
```
tzolov, line 26: Looks like we need to split the MockMcpTransport to client and server mocks.

mcp/src/test/java/io/modelcontextprotocol/server/McpServerProtocolVersionTests.java
```diff
@@ -27,93 +27,100 @@ private McpSchema.JSONRPCRequest jsonRpcInitializeRequest(String requestId, Stri
...
-		assertThat(response).isInstanceOf(McpSchema.JSONRPCResponse.class);
-		McpSchema.JSONRPCResponse jsonResponse = (McpSchema.JSONRPCResponse) response;
-		assertThat(jsonResponse.id()).isEqualTo(requestId);
-		assertThat(jsonResponse.result()).isInstanceOf(McpSchema.InitializeResult.class);
-		McpSchema.InitializeResult result = (McpSchema.InitializeResult) jsonResponse.result();
-		assertThat(result.protocolVersion()).isEqualTo(latestVersion);
-
-		server.closeGracefully().subscribe();
-	}
+	// @Test
```
tzolov, line 30: Needs to be refactored after the MockMcpTransport is split

mcp-test/src/main/java/io/modelcontextprotocol/MockMcpTransport.java
tzolov: Made the mcp-test's MockMcpTransport deprecated as no one is using it.

mcp/src/test/java/io/modelcontextprotocol/MockMcpTransport.java
tzolov: Resolved in the last commit

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
chemicL: It is correct. The intention was to allow us to hide the visibility in 0.9.0. We can also remove the deprecated javadoc note above now.

mcp/src/test/java/io/modelcontextprotocol/MockMcpServerTransport.java
```diff
@@ -0,0 +1,68 @@
...
+	public McpSchema.JSONRPCMessage getLastSentMessage() {
+		return !sent.isEmpty() ? sent.get(sent.size() - 1) : null;
+	}
+
+	@Override
+	public Mono<Void> closeGracefully() {
+		return Mono.defer(() -> {
+			return Mono.empty();
+		});
+	}
```
chemicL, line 61: No need for `defer`. `return Mono.empty()` is enough.

chemicL (changes requested):
Thanks! Overall all good, just two comments to address 👍

chemicL (approved):

tzolov:
rebased, squashed and merged at 25f3bad68d83367833f81da81714c3b0dcc7dcbd

PR #95, opened 2025-04-01
chore: Add AsyncCloseable interface.

He-Pin:
## Motivation and Context
Introduce a common interface for autocloseable.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
+++ b/mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
@@ -121,7 +121,7 @@ class WebFluxSseIntegrationTests {
 					.hasMessage("Client must be configured with sampling capabilities");
 			}
 		}
-		server.close();
+		server.closeGracefully().block();
 	}
 
 	@ParameterizedTest(name = "{0} : {displayName} ")
@@ -189,7 +189,7 @@ class WebFluxSseIntegrationTests {
 				assertThat(result.stopReason()).isEqualTo(CreateMessageResult.StopReason.STOP_SEQUENCE);
 			});
 		}
-		mcpServer.closeGracefully().block();
+		mcpServer.closeGracefully();
 	}
 
 	@ParameterizedTest(name = "{0} : {displayName} ")
@@ -267,7 +267,7 @@ class WebFluxSseIntegrationTests {
 			});
 		}
 
-		mcpServer.closeGracefully().block();
+		mcpServer.closeGracefully();
 	}
 
 	@ParameterizedTest(name = "{0} : {displayName} ")
@@ -326,7 +326,7 @@ class WebFluxSseIntegrationTests {
 
 		}
 
-		mcpServer.closeGracefully().block();
+		mcpServer.closeGracefully();
 	}
 
 	// ---------------------------------------
@@ -376,7 +376,7 @@ class WebFluxSseIntegrationTests {
 			});
 		}
 
-		mcpServer.close();
+		mcpServer.closeGracefully();
 	}
 
 	@ParameterizedTest(name = "{0} : {displayName} ")
@@ -410,7 +410,7 @@ class WebFluxSseIntegrationTests {
 			}
 		}
 
-		mcpServer.close();
+		mcpServer.closeGracefully();
 	}
 
 	@ParameterizedTest(name = "{0} : {displayName} ")
@@ -437,7 +437,7 @@ class WebFluxSseIntegrationTests {
 			});
 		}
 
-		mcpServer.close();
+		mcpServer.closeGracefully();
 	}
 
 	@ParameterizedTest(name = "{0} : {displayName} ")
@@ -471,7 +471,7 @@ class WebFluxSseIntegrationTests {
 			});
 		}
 
-		mcpServer.close();
+		mcpServer.closeGracefully();
 	}
 
 	@ParameterizedTest(name = "{0} : {displayName} ")
@@ -502,7 +502,7 @@ class WebFluxSseIntegrationTests {
 			});
 		}
 
-		mcpServer.close();
+		mcpServer.closeGracefully();
 	}
 
 	// ---------------------------------------
@@ -554,7 +554,7 @@ class WebFluxSseIntegrationTests {
 			assertThat(response).isEqualTo(callResponse);
 		}
 
-		mcpServer.close();
+		mcpServer.closeGracefully();
 	}
 
 	@ParameterizedTest(name = "{0} : {displayName} ")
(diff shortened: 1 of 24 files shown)


mcp/src/main/java/io/modelcontextprotocol/spec/AsyncCloseable.java
```diff
@@ -0,0 +1,24 @@
...
+     *
+     * @return A {@link Mono} that completes when the resource has been closed.
+     */
+    Mono<Void> closeGracefully();
+
+    /**
+     * Immediately closes the resource gracefully.
+     */
+    default void close() {
+        this.closeGracefully().subscribe();
```
He-Pin, line 22: Maybe the default implementation should call `block`

mcp/src/main/java/io/modelcontextprotocol/spec/AsyncCloseable.java
```diff
@@ -0,0 +1,24 @@
...
+import reactor.core.publisher.Mono;
+
+/**
+ * Interface for close operations that are asynchronous.
+ */
+public interface AsyncCloseable {
+
+	/**
+	 * Begins the process of closing the resource gracefully, if there is one in progress,
+	 * return the existing one.
```
chemicL, line 12: I don't think that's accurate? We'd need to keep track of closing-in-progress and I am not certain we do that currently.

mcp/src/main/java/io/modelcontextprotocol/spec/AsyncCloseable.java
```diff
@@ -0,0 +1,24 @@
...
+	 * @return A {@link Mono} that completes when the resource has been closed.
+	 */
+	Mono<Void> closeGracefully();
+
+	/**
+	 * Immediately closes the resource gracefully.
```
chemicL, line 18: The graceful close is not immediate, it kicks off asynchronously. In the cases of `void` return type it is a disconnected chain.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerTransportProvider.java
```diff
@@ -52,6 +52,7 @@ public interface McpServerTransportProvider {
 	 * Immediately closes all the transports with connected clients and releases any
 	 * associated resources.
 	 */
+	@Override
```
chemicL, line 55: This method can be removed.

mcp/src/main/java/io/modelcontextprotocol/spec/AsyncCloseable.java
```diff
@@ -0,0 +1,24 @@
...
+	 */
```
chemicL, line 19: An `@see #closeGracfully()` javadoc entry would be useful.

mcp/src/main/java/io/modelcontextprotocol/spec/McpSession.java
```diff
@@ -72,11 +72,13 @@ default Mono<Void> sendNotification(String method) {
 	 * Closes the session and releases any associated resources asynchronously.
 	 * @return a {@link Mono<Void>} that completes when the session has been closed.
 	 */
+	@Override
 	Mono<Void> closeGracefully();
 
 	/**
 	 * Closes the session and releases any associated resources.
 	 */
+	@Override
```
chemicL, line 81: This method can be removed.

mcp/src/main/java/io/modelcontextprotocol/spec/McpTransport.java
```diff
@@ -45,6 +45,7 @@ public interface McpTransport {
 	 * needed. It should handle the graceful shutdown of any active connections.
 	 * </p>
 	 */
+	@Override
```
chemicL, line 48: This method can be removed

chemicL (changes requested):
Thank you for the PR. I think with a few tweaks it will be a useful addition.

mcp/src/main/java/io/modelcontextprotocol/spec/AsyncCloseable.java
He-Pin: So yo suggest we drop the :
```
if there is one in progress, return the existing one.
```
?

I think we should always return the same instance.

He-Pin: I checked the code, you are right, not all implementation return the same `Mono`, will drop this description.

mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
```diff
@@ -189,7 +189,7 @@ void testCreateMessageSuccess(String clientType) {
 				assertThat(result.stopReason()).isEqualTo(CreateMessageResult.StopReason.STOP_SEQUENCE);
 			});
 		}
-		mcpServer.closeGracefully().block();
+		mcpServer.closeGracefully();
```
chemicL, line 192: This won't close the server since the Mono is not subscribed to. As this is an async server it is required to subscribe. I think the same issue has been introduced in multiple other places.

1. Please review all the code changes for this error.
2. Please do not force-push as it makes reviewing difficult.

Thanks!

chemicL (changes requested):

mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
He-Pin: 1. We should always return the same instance.
2. We should do the close/subscribe when invoking. 

But , these are not consistent across the codebase.

In akka/pekko, there is a `whenTerminate`, which will return the same instance too.

He-Pin: wdyt about this @chemicL , I'm a little busy at work, but I can update this pr with your input.

chemicL: I'm not sure - is this in reference to my comment above? I only wanted to indicate that the `mcpServer` instance is an `McpAsyncServer`, which means `closeGracefully()` returns a `Mono`, which needs to be subscribed to in order to trigger closing, otherwise nothing happens.

That comment does not touch upon returning the same instance of `Mono`, just the fact that it won't close anything.

For returning the same instance - I am not sure whether we need that - do you think we do? I imagine in most cases the double closing should be idempotent so perhaps there is no need. If we do have issues with double closing, let's open a new issue to address this and think of solutions, WDYT?

Kehrlann:
Closing due to inactivity

Issue #114, opened 2025-04-07
Clean up the HttpClientSseClientTransport creation


tzolov:
- Create the transport via Builder only. Reduce the constructor access to protected/private.
- Provide HttpClient.Builder and HttpRequest.Builder customizers to modify the builders before creating the transport.

tzolov:
To address this https://github.com/modelcontextprotocol/java-sdk/pull/101 as well

tzolov:
Resolved by fab434c088e7e90ad4cbbedd55b28c553536c7de

PR #117, opened 2025-04-07
refactor: adds Client/Reqeust customizer for HttpClientSseClientTransport

Aliaksie:
…

This PR improves the construction pattern for HttpClientSseClientTransport by enforcing usage via a builder. It also introduces support for customizing HttpClient.Builder and HttpRequest.Builder before the transport is created. This provides a cleaner, more extensible way to configure HTTP behavior. Related to [issue](https://github.com/modelcontextprotocol/java-sdk/issues/114)

## Motivation and Context

I came across this project and really liked its design and goals, so I wanted to start contributing.

## Breaking Changes

 Yes — this change makes the constructor for HttpClientSseClientTransport protected/private, which is a breaking change for users relying on direct instantiation.

To migrate, users should use the new builder-based approach for constructing HttpClientSseClientTransport.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

## Additional context

let me know if i missed something and need to add or if these changes don't make sense no problem just close this pr :-)

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
@@ -13,6 +13,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
 import com.fasterxml.jackson.core.type.TypeReference;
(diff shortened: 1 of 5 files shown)


mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
```diff
@@ -176,17 +145,20 @@ public static class Builder {
...
-		private HttpRequest.Builder requestBuilder = HttpRequest.newBuilder();
+		private HttpRequest.Builder requestBuilder = HttpRequest.newBuilder()
+			.header("Content-Type", "application/json");
 
 		/**
 		 * Creates a new builder with the specified base URI.
 		 * @param baseUri the base URI of the MCP server
 		 */
-		public Builder(String baseUri) {
+		Builder(String baseUri) {
```
chemicL, line 161: We need to deprecate it first in order to avoid breaking the users. It is currently public.

chemicL (changes requested):
Left some comments around backwards compatibility, but this is a useful change, thanks!

mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
Aliaksie: You're absolutely right — to avoid breaking existing users, I've marked all currently public constructors as @Deprecated and added clear Javadoc indicating that they will be removed in a future release.
In addition, I introduced an alternative package-private constructor, which will be used in the new builder-based approach.

mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
```diff
@@ -100,59 +101,27 @@ public class HttpClientSseClientTransport implements McpClientTransport {
...
+	 * @param httpClient the HTTP client to use
 	 * @param requestBuilder the HTTP request builder to use
 	 * @param baseUri the base URI of the MCP server
 	 * @param sseEndpoint the SSE endpoint path
 	 * @param objectMapper the object mapper for JSON serialization/deserialization
 	 * @throws IllegalArgumentException if objectMapper, clientBuilder, or headers is null
 	 */
-	public HttpClientSseClientTransport(HttpClient.Builder clientBuilder, HttpRequest.Builder requestBuilder,
-			String baseUri, String sseEndpoint, ObjectMapper objectMapper) {
+	HttpClientSseClientTransport(HttpClient httpClient, HttpRequest.Builder requestBuilder, String baseUri,
```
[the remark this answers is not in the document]
Aliaksie: You're absolutely right — to avoid breaking existing users, I've marked all currently public constructors as @Deprecated and added clear Javadoc indicating that they will be removed in a future release.
In addition, I introduced an alternative package-private constructor, which will be used in the new builder-based approach.

tzolov:
Thanks @Aliaksie 

Rebased, added tests and merged at fab434c088e7e90ad4cbbedd55b28c553536c7de

PR #132, opened 2025-04-09
feat(mcp): refactor logging to use exchange for targeted client notifications

### LINKED ISSUE #131: Refactor logging system to use exchange for targeted client notifications
**Bug description**
The current MCP logging implementation has several limitations:

- Logging notifications are broadcast to all connected clients regardless of their interest level
- There's a single global minimum logging level for the entire server
- Clients cannot individually configure which log levels they want to receive
- The current approach is inefficient as it sends unnecessary messages to clients

This design doesn't align with the per-session approach used for other MCP features like sampling and roots, where each client can have its own configuration.

**Expected behavior**

- Use the exchange mechanism to send logging notifications only to specific client sessions
- Track minimum logging level per client session rather than globally
- Filter logging notifications at the server side based on each client's configured level
- Change setLoggingLevel from a notification to a request/response pattern for proper acknowledgment
- Deprecate the old broadcasting behavior.

Acceptance Criteria
- Each client can set its own minimum logging level
- Logging notifications below a client's minimum level are filtered out
- Logging notifications at or above a client's minimum level are delivered
- Ddd integration tests

**Environment**
Affects all MCP versions including 0.9.0-SN

Related to https://github.com/modelcontextprotocol/java-sdk/issues/97

tzolov:
Refactors the MCP logging system to use the exchange mechanism for sending logging notifications only to specific client sessions rather than broadcasting to all clients.

- Move logging notification delivery from server-wide broadcast to per-session exchange
- Implement per-session minimum logging level tracking and filtering
- Change setLoggingLevel from notification to request/response pattern
- Deprecate global server.loggingNotification in favor of exchange.loggingNotification
- Add integration test demonstrating filtered logging notifications

Resolves #131


mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
```diff
@@ -125,6 +127,9 @@ void testCreateMessageWithoutSamplingCapabilities(String clientType) {
 			assertThat(e).isInstanceOf(McpError.class)
 				.hasMessage("Client must be configured with sampling capabilities");
 		}
+
+		server.close();
+		client.close();
```
chemicL, line 132: Can apply try-with-resources around these.

mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
```diff
@@ -339,8 +344,8 @@ void testRootsWithMultipleHandlers(String clientType) {
 		mcpServer.close();
 	}
 
-	@ParameterizedTest(name = "{0} : {displayName} ")
-	@ValueSource(strings = { "httpclient", "webflux" })
+	// @ParameterizedTest(name = "{0} : {displayName} ")
+	// @ValueSource(strings = { "httpclient", "webflux" })
```
chemicL, line 348: Commented out probably during testing.

mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
```diff
@@ -378,9 +380,9 @@ void testRootsServerCloseWithActiveSubscription(String clientType) {
 
 	String emptyJsonSchema = """
 			{
-				"$schema": "http://json-schema.org/draft-07/schema#",
```
chemicL, line 381: What happened here?

mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
```diff
@@ -396,7 +398,7 @@ void testToolCallSuccess(String clientType) {
 					// perform a blocking call to a remote service
 					String response = RestClient.create()
 						.get()
-						.uri("https://github.com/modelcontextprotocol/specification/blob/main/README.md")
+						.uri("https://raw.githubusercontent.com/modelcontextprotocol/modelcontextprotocol/refs/heads/main/README.md")
```
chemicL, line 401: Would be nice if we ran something locally instead of reaching out to a real url.

mcp-test/src/main/java/io/modelcontextprotocol/client/AbstractMcpAsyncClientTests.java
```diff
@@ -453,15 +453,15 @@ void testLoggingLevelsWithoutInitialization() {
...
+			Mono<Object> testAllLevels = mcpAsyncClient.initialize().then(Mono.defer(() -> {
+				Mono<Object> chain = Mono.empty();
 				for (McpSchema.LoggingLevel level : McpSchema.LoggingLevel.values()) {
 					chain = chain.then(mcpAsyncClient.setLoggingLevel(level));
 				}
 				return chain;
 			}));
 
-			StepVerifier.create(testAllLevels).verifyComplete();
+			StepVerifier.create(testAllLevels).expectNextCount(1).verifyComplete();
```
chemicL, line 464: What is the item? Does it have significance?

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -780,16 +780,15 @@ private NotificationHandler asyncLoggingNotificationHandler(
 	 * @return A Mono that completes when the logging level is set.
 	 * @see McpSchema.LoggingLevel
 	 */
-	public Mono<Void> setLoggingLevel(LoggingLevel loggingLevel) {
+	public Mono<Object> setLoggingLevel(LoggingLevel loggingLevel) {
```
chemicL, line 783: This is a backwards incompatible change. I understand we can classify this as a bug?

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
```diff
@@ -322,8 +322,8 @@ public GetPromptResult getPrompt(GetPromptRequest getPromptRequest) {
 	 * Client can set the minimum logging level it wants to receive from the server.
 	 * @param loggingLevel the min logging level
 	 */
-	public void setLoggingLevel(McpSchema.LoggingLevel loggingLevel) {
-		this.delegate.setLoggingLevel(loggingLevel).block();
+	public Object setLoggingLevel(McpSchema.LoggingLevel loggingLevel) {
```
chemicL, line 325: Why `Object`?

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
chemicL, line 783: Why `Object` and not an actual meaningful type?

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
```diff
@@ -322,8 +322,8 @@ public GetPromptResult getPrompt(GetPromptRequest getPromptRequest) {
 	 * Client can set the minimum logging level it wants to receive from the server.
 	 * @param loggingLevel the min logging level
```
chemicL, line 323: Missing `@return` javadoc

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -780,16 +780,15 @@ private NotificationHandler asyncLoggingNotificationHandler(
 	 * @return A Mono that completes when the logging level is set.
```
chemicL, line 780: The Javadoc needs updating.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -257,6 +258,9 @@ private static class AsyncServerImpl extends McpAsyncServer {
 
 		private final ConcurrentHashMap<String, McpServerFeatures.AsyncPromptSpecification> prompts = new ConcurrentHashMap<>();
 
+		// TODO: this field is deprecated and should be remvoed together with the
+		// broadcasting loggingNotification.
+		@Deprecated
```
chemicL, line 263: There is no need to deprecate a private field

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -680,12 +694,23 @@ public Mono<Void> loggingNotification(LoggingMessageNotification loggingMessageN
 			return this.mcpTransportProvider.notifyClients(McpSchema.METHOD_NOTIFICATION_MESSAGE, params);
 		}
 
-		private McpServerSession.RequestHandler<Void> setLoggerRequestHandler() {
+		private McpServerSession.RequestHandler<Object> setLoggerRequestHandler() {
```
chemicL, line 697: Shouldn't it be an actual type?

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -680,12 +694,23 @@ public Mono<Void> loggingNotification(LoggingMessageNotification loggingMessageN
...
-				this.minLoggingLevel = objectMapper.convertValue(params, new TypeReference<LoggingLevel>() {
-				});
+				return Mono.defer(() -> {
 
-				return Mono.empty();
+					SetLevelRequest newMinLoggingLevel = objectMapper.convertValue(params,
+							new TypeReference<SetLevelRequest>() {
+							});
+
+					exchange.setMinLoggingLevel(newMinLoggingLevel.level());
```
chemicL, line 705: I didn't find an actual implementation of this in the python reference SDK. Have you checked others? Is there a consensus here? I am wondering if it's the right way to implement it, as at least in the Python SDK the handler delegates to a user-provided function to handle it.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
```diff
@@ -101,4 +109,30 @@ public Mono<McpSchema.ListRootsResult> listRoots(String cursor) {
...
+		return Mono.defer(() -> {
+			if (loggingMessageNotification.level().level() < this.session.getMinLoggingLevel().level()) {
+				return Mono.empty();
+			}
+
+			return this.session.sendNotification(McpSchema.METHOD_NOTIFICATION_MESSAGE, loggingMessageNotification);
+		});
+	}
+
+	public void setMinLoggingLevel(LoggingLevel minLoggingLevel) {
```
chemicL, line 133: Public method requires a javadoc. Also, should this be exposed to the user? On one hand perhaps this should be in the user control, on the other it is an implementation of the spec if I'm correct, so maybe it should be hidden.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
```diff
@@ -101,4 +109,30 @@ public Mono<McpSchema.ListRootsResult> listRoots(String cursor) {
...
+	 * @return A Mono that completes when the notification has been sent
+	 */
+	public Mono<Void> loggingNotification(LoggingMessageNotification loggingMessageNotification) {
+
+		if (loggingMessageNotification == null) {
+			return Mono.error(new McpError("Logging message must not be null"));
+		}
+
+		return Mono.defer(() -> {
+			if (loggingMessageNotification.level().level() < this.session.getMinLoggingLevel().level()) {
```
chemicL, line 125: Thie logic of checking should be internal to the entity that maintains the state and not leak into the caller (session should control and check it if the session has the set level, not the exchange which is just the user).

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -53,6 +54,8 @@ public class McpServerSession implements McpSession {
 
 	private final AtomicInteger state = new AtomicInteger(STATE_UNINITIALIZED);
 
+	private final AtomicReference<LoggingLevel> minLoggingLevel = new AtomicReference<>(LoggingLevel.INFO);
```
chemicL, line 57: We probably just need a volatile field for an enum that we just set.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -138,6 +149,13 @@ public Mono<Void> sendNotification(String method, Map<String, Object> params) {
 		return this.transport.sendMessage(jsonrpcNotification);
 	}
 
+	public Mono<Void> sendNotification(String method, Object params) {
```
chemicL, line 152: Why do we need to add another public method to the `McpServerSession`? We need to maintain consistency here and make sure the surface is as minimal as we can.

mcp/src/test/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProviderIntegrationTests.java
```diff
@@ -44,7 +45,7 @@
 
 public class HttpServletSseServerTransportProviderIntegrationTests {
 
-	private static final int PORT = 8185;
+	private static final int PORT = 8189;
```
chemicL, line 48: Why change? Perhaps we need a simple utility to allocate a random port instead of relying on a constant? https://stackoverflow.com/questions/2675362/how-to-find-an-available-port

chemicL (changes requested):
I haven't gone through the tests yet as I have some concerns regarding the design and public API changes, let's address these first.

mcp-test/src/main/java/io/modelcontextprotocol/client/AbstractMcpAsyncClientTests.java
tzolov: it is null same as the ping

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
tzolov: because we don't have ObjectMapper outside the session and can't convert the object to map as expected by the schema.  Perhaps we need to make this as the public apis and remove the one that takes explicit map.

mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
tzolov: yep, this is a mistake.

mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
tzolov: A mistake while testing. Btw the $schema filed is optional for the JSON Schema, which explains why it was working.

mcp-test/src/main/java/io/modelcontextprotocol/client/AbstractMcpAsyncClientTests.java
tzolov: This PR makes the setLevel call a `request` instead of `notification`. It should be a request according to the spec.
But a void request.

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
tzolov: yes it is a bug

tzolov: because there is not response. It is the same request as the ping.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
tzolov: this is more for us not to forget to delete it. But i can remove the annotation.

mcp/src/test/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProviderIntegrationTests.java
tzolov: This makes sense. Spring Boot used to have such utility but discarded it for some reason. 
Lets handle this in a dedicated issue/pr

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
tzolov: Because the McpAsyncServerExchange doesn't have access to transport and can not convert the `LoggingMessageNotification` input into  `Map<String, Object> params`. 
I guess we need to relax the sendNotification signature to accept Object params instead of the Map. but this is a breaking change.  I will add  a comment on the method.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
tzolov: sure

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
tzolov: it can

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
tzolov: this request returns empty response. Same like Ping

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
tzolov: anything particular?

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
tzolov: I agree we can make it a protected so that only the server cha use it. We do not allow the users to register custom METHOD_LOGGING_SET_LEVEL. So this should remain spec implementation detail

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
chemicL: Ok, it's an empty response. Got it.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
chemicL: Ok thanks! Having consulted other impls, I think it's correct for us to store the state in the exchange and not expose this to the users at all.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
chemicL: Just to keep the record of where this ended up - #137 addressed an earlier mistake that triggered this problematic situation and we are now in a cleaner state :heart:

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -134,7 +134,8 @@ public <T> Mono<T> sendRequest(String method, Object requestParams, TypeReferenc
 	@Override
 	public Mono<Void> sendNotification(String method, Object params) {
 		McpSchema.JSONRPCNotification jsonrpcNotification = new McpSchema.JSONRPCNotification(McpSchema.JSONRPC_VERSION,
-				method, params);
+				method, this.transport.unmarshalFrom(params, new TypeReference<Map<String, Object>>() {
```
chemicL, line 137: I think we don't need
```java
this.transport.unmarshalFrom(params, new TypeReference<Map<String, Object>>() {})
```

and we can just provide `params` directly.

I'll commit this change.

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
chemicL: Actually, I think we should not change this signature. Internally there is an empty result, agreed. But I think we can leave the API as the empty response is not useful to the user.

tzolov:
rebased, squashed and merged at c88ac937f3e195c7e767c61e5024737c3417ad72

PR #137, opened 2025-04-10
refactor: change notification params type from Map<String,Object> to Object

tzolov:
This change generalizes the parameter type for notification methods across the MCP framework, allowing for more flexible parameter passing. Instead of requiring parameters to be structured as a Map<String,Object>, the API now accepts any Object as parameters.

The primary motivation is to simplify client usage by allowing direct passing of strongly-typed
 objects without requiring conversion to a Map first, as demonstrated in the McpAsyncServer
logging notification implementation.

Affected components:

- McpSession interface and implementations
- McpServerTransportProvider interface and implementations
- McpSchema JSONRPCNotification record

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
@@ -669,15 +669,17 @@ public class McpAsyncServer {
 				return Mono.error(new McpError("Logging message must not be null"));
 			}
 
-			Map<String, Object> params = this.objectMapper.convertValue(loggingMessageNotification,
-					new TypeReference<Map<String, Object>>() {
-					});
+			// Map<String, Object> params =
+			// this.objectMapper.convertValue(loggingMessageNotification,
+			// new TypeReference<Map<String, Object>>() {
+			// });
 
 			if (loggingMessageNotification.level().level() < minLoggingLevel.level()) {
 				return Mono.empty();
 			}
 
-			return this.mcpTransportProvider.notifyClients(McpSchema.METHOD_NOTIFICATION_MESSAGE, params);
+			return this.mcpTransportProvider.notifyClients(McpSchema.METHOD_NOTIFICATION_MESSAGE,
+					loggingMessageNotification);
 		}
 
 		private McpServerSession.RequestHandler<Void> setLoggerRequestHandler() {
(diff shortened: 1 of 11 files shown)


mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -669,15 +669,17 @@ public Mono<Void> loggingNotification(LoggingMessageNotification loggingMessageN
 				return Mono.error(new McpError("Logging message must not be null"));
 			}
 
-			Map<String, Object> params = this.objectMapper.convertValue(loggingMessageNotification,
-					new TypeReference<Map<String, Object>>() {
-					});
+			// Map<String, Object> params =
+			// this.objectMapper.convertValue(loggingMessageNotification,
+			// new TypeReference<Map<String, Object>>() {
+			// });
```
chemicL, line 676:
```suggestion
```

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerTransportProvider.java
```diff
@@ -42,11 +42,11 @@ public interface McpServerTransportProvider {
 	/**
 	 * Sends a notification to all connected clients.
 	 * @param method the name of the notification method to be called on the clients
-	 * @param params a map of parameters to be sent with the notification
+	 * @param params parameters to be sent with the notification
 	 * @return a Mono that completes when the notification has been broadcast
 	 * @see McpSession#sendNotification(String, Map)
 	 */
-	Mono<Void> notifyClients(String method, Map<String, Object> params);
+	Mono<Void> notifyClients(String method, Object params);
```
chemicL, line 49: Let's keep in mind to take note of the breaking change in the release docs. I think it's worth the effort though.

mcp/src/main/java/io/modelcontextprotocol/spec/McpSession.java
```diff
@@ -63,10 +63,10 @@ default Mono<Void> sendNotification(String method) {
 	 * parameters with the notification.
 	 * </p>
 	 * @param method the name of the notification method to be sent to the counterparty
-	 * @param params a map of parameters to be sent with the notification
+	 * @param params parameters to be sent with the notification
 	 * @return a Mono that completes when the notification has been sent
 	 */
-	Mono<Void> sendNotification(String method, Map<String, Object> params);
+	Mono<Void> sendNotification(String method, Object params);
```
chemicL, line 69: Let's keep in mind to take note of the breaking change in the release docs. I think it's worth the effort though.

chemicL (approved):
Breaking change alert, but I think it's absolutely worth it.

PR #140, opened 2025-04-10
feat: Add Pagination for server requests

### LINKED ISSUE #967: Server should handle paginated requests
**Expected Behavior**

The server should handle paginated requests per [specification](https://modelcontextprotocol.io/specification/2025-11-25/server). 

**Current Behavior**

Currently, the whole list is returned on list requests.

**Context**

When a server contains big lists of tools, prompts, etc. the requests may put too much load on the server or take too long.
The code has TODOs at some of the affected places.

denniskawurek:
Adds the Pagination feature to the `prompts/list` feature to fulfil the specification.

## Motivation and Context

This adds the Pagination feature to the `prompts/list` feature as described in the [specification](https://modelcontextprotocol.io/specification/2025-03-26/server/utilities/pagination).

To make this possible mainly two changes are made:

1. The logic for cursor handling is added.
2. Handling for invalid parameters (MCP error code `-32602 (Invalid params)`) is added to the `McpServerSession`.

Decisions made:

1. For now the cursor is the integer value of the start index encoded as a `base64` string.
2. The page size is set to `10`.
3. When parameters are found to be invalid the newly introduced `McpParamsValidationError` is returned to handle it properly in the `McpServerSession`.

If this change is ok, I'm happy to create a follow-up PR for the other MCP features. Maybe some parts of the cursor handling can be moved out to separate methods.

## How Has This Been Tested?

Integration tests are added for the ` prompts/list`  feature.

## Types of changes

## Checklist

Resolves #967


chemicL (changes requested):
Thank you for the contribution! It is a great start, would you like to take this further and unify the logic for other listing requests?

denniskawurek:
Hi,

thanks for your review and your feedback!

> Thank you for the contribution! It is a great start, would you like to take this further and unify the logic for other listing requests?

Yes I will do this. I should be able to update the pull request by the end of the week.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -225,10 +225,20 @@ private Mono<McpSchema.JSONRPCResponse> handleIncomingRequest(McpSchema.JSONRPCR
...
 				.map(result -> new McpSchema.JSONRPCResponse(McpSchema.JSONRPC_VERSION, request.id(), result, null))
-				.onErrorResume(error -> Mono.just(new McpSchema.JSONRPCResponse(McpSchema.JSONRPC_VERSION, request.id(),
-						null, new McpSchema.JSONRPCResponse.JSONRPCError(McpSchema.ErrorCodes.INTERNAL_ERROR,
-								error.getMessage(), null)))); // TODO: add error message
-																// through the data field
+				.onErrorResume(error -> {
+
+					var errorCode = McpSchema.ErrorCodes.INTERNAL_ERROR;
+
+					if (error instanceof McpParamsValidationError) {
```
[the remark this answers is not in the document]
denniskawurek: Sounds good! I think, too that a refactoring here should belong to a separate merge request.

mcp/src/main/java/io/modelcontextprotocol/spec/McpParamsValidationError.java
```diff
@@ -0,0 +1,13 @@
+package io.modelcontextprotocol.spec;
+
+public class McpParamsValidationError extends McpError {
+
+	public McpParamsValidationError(McpSchema.JSONRPCResponse.JSONRPCError jsonRpcError) {
```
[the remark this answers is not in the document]
denniskawurek: Done.

mcp/src/main/java/io/modelcontextprotocol/spec/McpParamsValidationError.java
```diff
@@ -0,0 +1,13 @@
...
+		super(jsonRpcError.message());
+	}
+
+	public McpParamsValidationError(Object error) {
```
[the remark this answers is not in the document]
denniskawurek: Done.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -647,20 +651,67 @@ public Mono<Void> notifyPromptsListChanged() {
...
+						});
+
+				if (!isCursorValid(request.cursor(), this.prompts.size())) {
+					return Mono.error(new McpParamsValidationError("Invalid cursor"));
+				}
+
+				int requestedStartIndex = 0;
+
+				if (request.cursor() != null) {
+					requestedStartIndex = decodeCursor(request.cursor());
```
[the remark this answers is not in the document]
denniskawurek: I've done that. The cursor now contains the index and the hash of the collection. The values are separated by a colon. I found it is the easiest format for encoding/decoding, parsing etc.

denniskawurek:
I did another push with small changes in the commit title and message and have also changed a few comments and names in tests. No business logic was affected.

pantanurag555:
@denniskawurek Is this change still being worked on? I am making some changes to handle pagination on the client side and just saw that the server side changes are yet to be merged. I can pick up the remaining work on this PR if you need some help.

denniskawurek:
Hi @pantanurag555 

I've just merged main into the branch, so that it's up to date, again.

Regarding your question I haven't started anything in regards of the client, yet.
The idea was here to have only the implementation for the server and implement the client in a separate commit/merge request, to keep the change as small as possible like it's stated in the contribution guidelines.
However that's the way I'd prefer to do.

Maybe @chemicL can give some information of the maintainer point of view and whether this change is interesting.

Regards
Dennis

denniskawurek:
Looks like the test is flaky somehow. However, I'm not allowed to rerun the job.

lwilli:
Hi, thanks for putting this work together. I am very interested in seeing PR merged as it's blocking some of my work right now. What needs to be done to move this along? Thanks in advance!

denniskawurek:
Resolves #967

mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
```diff
@@ -400,11 +405,27 @@ public Mono<Void> removeTool(String toolName) {
...
-			List<Tool> tools = this.tools.stream()
-				.map(McpStatelessServerFeatures.AsyncToolSpecification::tool)
-				.toList();
-			return Mono.just(McpSchema.ListToolsResult.builder(tools).build());
+		return (exchange, params) -> {
+			var paginatedRequest = jsonMapper.convertValue(params, PAGINATED_REQUEST_TYPE_REF);
+			var cursor = paginatedRequest != null ? paginatedRequest.cursor() : null;
+
+			var mapSize = this.tools.size();
+			var mapHash = this.tools.hashCode();
```
Kehrlann: This is likely wrong across horizontally-scaled instances -  `hashCode` hashes the `*ToolSpecification`, which contain handlers, and the hash of those will be tied to object _identity_.

Additionally, that's hashing a lot of things, including input and output schema strings, without ever caching them.

We should probably hash across keys (tool names), but I guess we have no guarantee of tool names being unique so there's something to do about this too

mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
```diff
@@ -563,22 +584,54 @@ public Mono<Void> removeResourceTemplate(String uriTemplate) {
...
+			var mapHash = this.resources.hashCode();
+
+			return handleCursor(cursor, mapSize, mapHash).map(requestedStartIndex -> {
+				var startIndex = requestedStartIndex != null ? requestedStartIndex : 0;
+				var endIndex = Math.min(startIndex + PAGE_SIZE, mapSize);
+
+				var nextCursor = getCursor(endIndex, mapSize, mapHash);
+
+				var resultList = this.resources.values()
+					.stream()
```
Kehrlann: I think the ordering is not guaranteed to be stable in a `ConcurrentHashMap`. I think in practice it's stable, BUT if you start adding elements dynamically and grow the underlying map, you may end up with different orders. I don't think it's a massive problem in this specific case.

If we do order by key as suggested in tools, we can sort the keys and get a stability guarantee (unless you modify the collection during a pagination run, but that's a corner case.)

mcp-test/src/main/java/io/modelcontextprotocol/AbstractMcpClientServerIntegrationTests.java
```diff
@@ -1427,6 +1431,158 @@ void testToolListChangeHandlingSuccess() {
...
+						.build(), HttpResponse.BodyHandlers.ofString());
+				String responseBody = response.body();
+				assertThat(responseBody).isNotBlank();
+			}
+			catch (Exception e) {
+				e.printStackTrace();
+			}
+			return callResponse;
+		};
+	}
```
Kehrlann, line 1717: Is there a reason for reaching out to a remote service in tests? I'd rather avoid that and just return a result.

Kehrlann (commented):
Hey @denniskawurek , picking this up from #967 

A few concerns here:

1. It's a breaking change ; we should default to non-paginated results
2. Page size should be configurable
3. We've introduced tool filtering as part of #1108 ; so now the view is request dependent. If that PR gets merged first, you'll need to adapt the pagination to happen after the filtering (shouldn't be too hard)

mcp-test/src/main/java/io/modelcontextprotocol/AbstractMcpClientServerIntegrationTests.java
denniskawurek: Hi @Kehrlann ,
thanks for your feedback. I can apply it, but have a question about the first point: Can you specify please how to default to non-paginated results?

If I understand correctly: the first request has no cursor. That's  the starting request, not a special signal. If the server includes a nextCursor in its response, that tells the client more results exist and it should paginate. If nextCursor is absent, the client has everything.

But the spec doesn't give the client a way to ask for non-paginated results, and doesn't give the server a signal for when to skip pagination - that's entirely the server's own choice, made per response, with nothing in the request driving it.

So effectively the server would always return the whole list which makes the code for pagination never be run.

denniskawurek: > Is there a reason for reaching out to a remote service in tests? I'd rather avoid that and just return a result.

I have done it for consistency reasons as there are other tests which do the same (e.g. `testToolListChangeHandlingSuccess` ). I can change it for my specific case, but just so that you know that there are other tests which do the same.

Kehrlann:
> I have done it for consistency reasons as there are other tests which do the same (e.g. testToolListChangeHandlingSuccess ). I can change it for my specific case, but just so that you know that there are other tests which do the same.

Yes, please use a local computation. Thanks for flagging the other tests, I was not aware. I'll fix them separately.

> Can you specify please how to default to non-paginated results?

When a user creates an MCP server with the SDK, the default behavior should be the current behavior, and return the full lists by default. If they wish, they _can_ opt in to pagination for their server, but only through an explicit configuration. I don't have strong opinions on how to do this, e.g. return everything if the server configuration has `page size <= 0` (but I'm open to all suggestions).

denniskawurek:
Hi @Kehrlann ,

thank you for your answers on my previous questions.

Based on this I published some changes to the PR:


> 1. It's a breaking change ; we should default to non-paginated results
> 2. Page size should be configurable

See ` McpServer.java`. It makes the page size configurable. The default value is 0 encoded behind the constant ` NO_PAGING`.

Request handlers process the cursor attribute only if a page size on the server is configured. Otherwise it will default to returning everything (for backwards compatibility).

> 3. We've introduced tool filtering as part of #1108 ; so now the view is request dependent. If that PR gets merged first, you'll need to adapt the pagination to happen after the filtering (shouldn't be too hard)

This is implemented as requested.

mcp-test/src/main/java/io/modelcontextprotocol/AbstractMcpClientServerIntegrationTests.java
denniskawurek: I have removed the call to an external resource as discussed.

mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
denniskawurek: I have changed the request handler in a way that they apply sorting before reading the elements.

mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
denniskawurek: I've added ` computeToolsHash`  method which is used for calculating the map hash.

PR #144, opened 2025-04-11
feat(client): adds StreamableHttpClientTransport

Aliaksie:
adds StreamableHttpClientTransport 
## How Has This Been Tested?

WIP

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

WIP

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/client/transport/StreamableHttpClientTransport.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/transport/StreamableHttpClientTransport.java
@@ -1,0 +1,319 @@
+/*
+ * Copyright 2024 - 2024 the original author or authors.
+ */
+
+package io.modelcontextprotocol.client.transport;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.modelcontextprotocol.spec.McpClientTransport;
+import io.modelcontextprotocol.spec.McpSchema;
+import io.modelcontextprotocol.util.Assert;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.util.retry.Retry;
+
+/**
+ * A transport implementation for the Model Context Protocol (MCP) using JSON streaming.
+ *
+ * @author Aliaksei Darafeyeu
+ */
+public class StreamableHttpClientTransport implements McpClientTransport {
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(StreamableHttpClientTransport.class);
+
+	private final HttpClientSseClientTransport sseClientTransport;
+
+	private final HttpClient httpClient;
+
+	private final HttpRequest.Builder requestBuilder;
+
+	private final ObjectMapper objectMapper;
+
+	private final URI uri;
+
+	private final AtomicReference<TransportState> state = new AtomicReference<>(TransportState.DISCONNECTED);
+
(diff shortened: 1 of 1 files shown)


mcp/src/main/java/io/modelcontextprotocol/client/transport/StreamableHttpClientTransport.java
```diff
@@ -0,0 +1,396 @@
...
+						data += line.substring(6) + "\n";
+					else if (line.startsWith("id: "))
+						id = line.substring(4).trim();
+				}
+
+				if (data.endsWith("\n")) {
+					data = data.substring(0, data.length() - 1);
+				}
+
+				return new FlowSseClient.SseEvent(event, data, id);
```
chickenlj, line 340: The right argument sequence here should be `FlowSseClient.SseEvent(id, event, data);`

chickenlj (changes requested):

mcp/src/main/java/io/modelcontextprotocol/client/transport/StreamableHttpClientTransport.java
Aliaksie: Good catch, thanks!

mcp/src/main/java/io/modelcontextprotocol/client/transport/StreamableHttpClientTransport.java
```diff
@@ -0,0 +1,396 @@
...
+	 * A builder for creating instances of WebSocketClientTransport.
+	 */
+	public static class Builder {
+
+		private final HttpClient.Builder clientBuilder = HttpClient.newBuilder()
+			.version(HttpClient.Version.HTTP_1_1)
+			.connectTimeout(Duration.ofSeconds(10));
+
+		private final HttpRequest.Builder requestBuilder = HttpRequest.newBuilder()
+			.header("Accept", "application/json, text/event-stream");
```
chickenlj, line 101: I think we should put "Content-type: application/json" request header here. Some servers might not be able to process the request without this header.

chickenlj:
Great job @Aliaksie

Here are some of my thoughts after reading this PR and the SDK source code.

I think just adding the McpClientTransport implementation cannot give full support for Streamable HTTP because how it works is inconsistent with HTTP+SSE in many places, which might require some redesign or refactoring of the current SDK implementation:

1. The initialization process is not necessary in Streamable HTTP, it is only required when implementing stateful management. After initialization, all subsequent requests are required to have a  `mcp-session-id header`. However, the current Java SDK implementation enforces checking of initialization status and does not manage the `mcp-session-id` after initialization.
2. `/mcp GET` requests are constrained in the protocol to be used when the client wants to initiate a SSE request proactively, while the current implementation initiates a GET request and establishe SSE session every time on transport connect. Subsequent POST requests also rely on this to send back responses, which can be seen through the operation of the `pendingResponses` field.

Aliaksie:
> Great job @Aliaksie
> 
> Here are some of my thoughts after reading this PR and the SDK source code.
> 
> I think just adding the McpClientTransport implementation cannot give full support for Streamable HTTP because how it works is inconsistent with HTTP+SSE in many places, which might require some redesign or refactoring of the current SDK implementation:
> 
> 1. The initialization process is not necessary in Streamable HTTP, it is only required when implementing stateful management. After initialization, all subsequent requests are required to have a  `mcp-session-id header`. However, the current Java SDK implementation enforces checking of initialization status and does not manage the `mcp-session-id` after initialization.
> 2. `/mcp GET` requests are constrained in the protocol to be used when the client wants to initiate a SSE request proactively, while the current implementation initiates a GET request and establishe SSE session every time on transport connect. Subsequent POST requests also rely on this to send back responses, which can be seen through the operation of the `pendingResponses` field.

Hey @chickenlj Thanks a lot for the thoughtful feedback!

1 I added to Streamable HTTP transport code to manage `mcp-session-id`, and I checked documentation (transports and lifecycle ) and it seems to me that initialization request is still required, even in Streamable HTTP  but the server may or may not return a `mcp-session-id`
2 Yes — that’s a really important. I added the logic to make `connect` lazy and optional  with default connect on init for for backward compatibility...

I hope all this makes the SDK compliant with Streamable HTTP's flexibility while still maintaining support for HTTP+SSE fallback

Aliaksie:
Hey @tzolov, hope you're doing well! Just wanted to check in — have you had a chance to review this PR? I’d appreciate any feedback when you get a moment. Looking forward to your thoughts!

shaohzhangebay:
Hi team, could you please share the timeline to have streamable http for mcp java sdk ?

skyrdis:
when this feature could be ready to be parity with typescript and python?

Aliaksie:
Hi  @tzolov👋 Just checking in to see if there’s been a chance to review this PR 

Would love to hear if this makes sense for the SDK or if I should revise or break it down further.
Thanks a lot for your time!

PR #147, opened 2025-04-12
feat: Support cancellation notification while request timeout

CrazyHZM:
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/spec/McpClientSession.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/spec/McpClientSession.java
@@ -8,6 +8,7 @@ import java.time.Duration;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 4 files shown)


chemicL:
I am not sure we should send a cancellation notification for timeouts. Can you point me to the specification where it is mentioned timeouts should issue cancellation notifications? I agree cancellations should be supported, but I'm uncertain whether it should be automatically issued for timing out of requests.

CrazyHZM:
@chemicL 
The specification does not indicate that a requirement notification needs to be sent when the timeout occurs, but I think this is a reasonable scenario for canceling the request because the client no longer cares about the response. I think it should be proposed as a specification when handling timeouts.
I created an issue to discuss this matter: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/418

chemicL:
@CrazyHZM thanks. I suggest we close the PR for now and feel free to reopen it if the issue for the spec update is resolved positively.

Issue #103, opened 2025-04-02
Expected Behavior


a67793581:
**Expected Behavior**

 The implementation should properly handle cases where `messageEndpoint` already contains a protocol and host to avoid incorrect concatenation.

**Current Behavior**

Currently, the Java MCP SDK does not correctly handle `messageEndpoint`. Instead of properly resolving the full URL, it simply concatenates `messageEndpoint` with the base URL as a raw string. This results in malformed URLs, such as: http://localhost/mcphttp://localhost/mcp/message

**Context**
https://github.com/mark3labs/mcp-go/issues/76#issuecomment-2756278292
Here are the references:  

https://github.com/modelcontextprotocol/java-sdk/releases/tag/v0.8.1
```java
public Mono<Void> sendMessage(McpSchema.JSONRPCMessage message) {
    if (this.isClosing) {
        return Mono.empty();
    } else {
        try {
            if (!this.closeLatch.await(10L, TimeUnit.SECONDS)) {
                return Mono.error(new McpError("Failed to wait for the message endpoint"));
            }
        } catch (InterruptedException var5) {
            return Mono.error(new McpError("Failed to wait for the message endpoint"));
        }

        String endpoint = (String)this.messageEndpoint.get();
        if (endpoint == null) {
            return Mono.error(new McpError("No message endpoint available"));
        } else {
            try {
                String jsonText = this.objectMapper.writeValueAsString(message);
                HttpRequest request = HttpRequest.newBuilder().uri(URI.create(this.baseUri + endpoint)).header("Content-Type", "application/json").POST(BodyPublishers.ofString(jsonText)).build();
                return Mono.fromFuture(this.httpClient.sendAsync(request, BodyHandlers.discarding()).thenAccept((response) -> {
                    if (response.statusCode() != 200 && response.statusCode() != 201 && response.statusCode() != 202 && response.statusCode() != 206) {
                        logger.error("Error sending message: {}", response.statusCode());
                    }

                }));
            } catch (IOException var6) {
                return !this.isClosing ? Mono.error(new RuntimeException("Failed to serialize message", var6)) : Mono.empty();
            }
        }
    }
}```

jpfortier:
Hi, do you know if there is a workaround? Is this due to a misconfiguration on my part? (seeing this error on a call that worked yesterday)

a67793581:
[...]

PR #150, opened 2025-04-13
fix: resolve absolute and relative message endpoint uri

ashakirin:
fixes java mcp message endpoint error #103


tzolov:
Thanks @ashakirin !

cherry-picked, edited, squashed, rebased and merged at 344f1b838340efbcf0cfcd428358bdb77e9a533c

CaelumF:
Should the default sse endpoint not be `./sse` instead of `/sse` ? when running with localhost/something/something in the basePath (multiple servers are intended to be reachable in the same program), the change from the latest dev snapshot is resolving to just localhost/sse @tzolov

PR #155, opened 2025-04-14
Fix flaky test running blocking code in event loop

chemicL:
Deadlock caused by blocking in an async handler is now eliminated and tests are no longer flaky.


tzolov:
Rebase, fix previously introduced bogus tests configurations, merge at f7f8ccd0acb6d39558b65ceb1ae4e5f71619a37c

PR #178, opened 2025-04-21
feat: adds mcp-spi

Aliaksie:
- adds `mcp-spi` module 
- renamed `mcp` to `mcp-reactor`
- `mcp-reactor` use first version of `mcp-spi` 

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

let me know if i missed something and need to add or if these changes don't make sense no problem just close this pr :-)


Aliaksie:
Hey @tzolov, hope you're doing well! Just wanted to check in — have you had a chance to review this PR? I’d appreciate any feedback when you get a moment. Looking forward to your thoughts!

Aliaksie:
Hi  @tzolov👋 Just checking in to see if there’s been a chance to review this PR 

I’d love to get your thoughts on whether this is a direction the project is open to, or if there’s anything I should revise to help move it forward. Happy to discuss further or break it into smaller pieces if that helps with review.
Thanks a lot for your time!

mcp-bom/pom.xml
```diff
@@ -26,10 +26,24 @@
...
-            <!-- Core MCP -->
+            <!-- MCP Spec -->
             <dependency>
                 <groupId>io.modelcontextprotocol.sdk</groupId>
-                <artifactId>mcp</artifactId>
+                <artifactId>mcp-spi</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+
+            <!-- MCP Jacson Schema -->
```
134130, line 36:
```suggestion
            <!-- MCP Jackson Schema -->
```

A nit typo!

mcp-spi/src/main/java/io/modelcontextprotocol/schema/McpSchemaCodec.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Codec interface for encoding and decoding JSON-RPC messages.
+ *
+ * @author Aliaksei Darafeyeu
+ */
+public interface McpSchemaCodec {
+
+	byte[] encode(McpSchema.JSONRPCMessage message);
+
+	String encodeAsString(Object message) throws IOException;
```
DRayX, line 18: This method is only ever used for `McpSchema.JSONRPCMessage`  and `McpError` instances. It may make more sense to change the signature to `String encodeAsString(McpSchea.JSONRPCMessage message)` (which would match decodeFromString) and add a separate method or overload for encoding `McpError` instances.

mcp-spi/src/main/java/io/modelcontextprotocol/schema/McpType.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package io.modelcontextprotocol.schema;
+
+import java.lang.reflect.Type;
+
+/**
+ * @author Aliaksei Darafeyeu
+ */
+public interface McpType<T> {
```
DRayX, line 12: This interface doesn't quite make sense.  It's always used to wrap a Class instance, never a ParameterizedType or GenericArray.  Unlike Jackson `TypeReference`, this class doesn't support obtaining full type information by sub-classing, so I don't really see the value in having this wrapper.  It seems like it would make more sense to just use `Class<T>` in it's place.

mcp-spi/src/main/java/io/modelcontextprotocol/logger/McpLogger.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2025-2025 the original author or authors.
+ */
+
+package io.modelcontextprotocol.logger;
+
+/**
+ * @author Aliaksei Darafeyeu
+ */
+public interface McpLogger {
```
DRayX, line 10: This logging interface doesn't seem ideal; it feels like all logging levels should support attaching a `Throwable` cause, the number of methods seems excessive, and there's no way to lazily evaluate string parameters if the message is omitted.  Maybe a single `void log(McpSchema.LoggingLevel level, Supplier<String> messageSupplier, @Nullable Throwable cause)` method would make more sense.

mcp-spi/src/main/java/io/modelcontextprotocol/schema/McpSchemaCodec.java
DRayX: Actually, on second thought, I'm not sure why this is used with `McpError` at all; maybe those lines could just avoid going through the codec.

mcp-spi/src/main/java/io/modelcontextprotocol/schema/McpSchemaCodec.java
```diff
@@ -0,0 +1,28 @@
...
+package io.modelcontextprotocol.schema;
+
+import java.io.IOException;
+
+/**
+ * Codec interface for encoding and decoding JSON-RPC messages.
+ *
+ * @author Aliaksei Darafeyeu
+ */
+public interface McpSchemaCodec {
```
DRayX, line 14: The byte[] base encode and decode as well as `decodeBytes` never seem to be used.  The actual encoded type of the data is likely dependent on the specific transport being used.  I wonder if it might make more sense to change this to something like:

```java
public interface McpSchemaCodec<T> {
  T encode(McpSchema.JSONRPCMessage message) throws IOException;
  McpSchema.JSONRPCMessage decode(T message) throws IOException;
  <U> U unmarshalFrom(Object data, Class<T> clazz);
}
```

Then, each transport would be capable of using the codecs for the encoded data types it understands instead of each codec needing to support all possible encodings.

Issue #187, opened 2025-04-23
Fix Mockito inline mocking for Java 21+


chemicL:
Running `mvn clean verify` using Java 21, the following is printed:

Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation: https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3
WARNING: A Java agent has been loaded dynamically (/Users/jdariusz/.m2/repository/net/bytebuddy/byte-buddy-agent/1.15.11/byte-buddy-agent-1.15.11.jar)
WARNING: If a serviceability tool is in use, please run with -XX:+EnableDynamicAgentLoading to hide this warning
WARNING: If a serviceability tool is not in use, please run with -Djdk.instrument.traceUsage for more information
WARNING: Dynamic loading of agents will be disallowed by default in a future release
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended

When run with Java 24 the build just fails.

PR #207, opened 2025-04-25
fix: Mockito inline mocking for Java 21+

denniskawurek:
Before this fix the execution of the maven surefire plugin with Java 21 logged warnings that mockito should be added as a java agent, because the self-attaching won't be supported in future java releases.
In Java 24 the test just broke.

This problem is solved by modifying the `pom.xml` of the parent and doing this changes:

* Adding mockito as a java agent.
* Removing the `surefireArgLine` from the properties. This can be added back when it's needed (for example if JaCoCo will be used).

Furthermore, the `pom.xml` in the `mcp-spring-*` modules now have the byte-buddy dependency included, as the test would otherwise break when trying to mock `McpSchema#CreateMessageRequest`.

## Motivation and Context
This is a fix for #187 

## How Has This Been Tested?

I executed `mvn clean package`  with Java 17, 21 and 24 and checked for the warning and whether the modules can be built.

## Types of changes

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/pom.xml
+++ b/mcp-spring/mcp-spring-webflux/pom.xml
@@ -83,6 +83,12 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>net.bytebuddy</groupId>
+			<artifactId>byte-buddy</artifactId>
+			<version>${byte-buddy.version}</version>
+			<scope>test</scope>
+		</dependency>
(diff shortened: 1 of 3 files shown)


chemicL:
Thank you @denniskawurek 👏  This has now been merged as eae3840e7d44932c60c131cb7a346b5367b788ff with a small improvement regarding the `surefireArgLine`.

PR #228, opened 2025-05-09
Remove temporary delegate impl from McpAsyncServer

chemicL:
In [0.8.0](https://github.com/modelcontextprotocol/java-sdk/releases/tag/v0.8.0) we introduced the exchange concept and required temporary existence of a dual stack API - prior one without the exchange in server features and the new one that has it. Since then we removed the deprecated functionality but we left the temporary `McpAsyncServer.AsyncServerImpl` delegate used by `McpAsyncServer`. This PR removes this temporary code.


tzolov:
Thanks,  @chemicL 
This was a left over we missed for the 0.9.0. LGTM

tzolov (commented):
LGTM

Issue #115, opened 2025-04-07
Improve the HttpClientSseClientTransport internals


tzolov:
Ensure a consistent concurrency implementation that does not mix different paradigms.

chemicL:
I think this was addressed in #337

PR #234, opened 2025-05-13
fix: Preserve thread interrupt status after catching InterruptedException

rohitdutt-04:
## Motivation and Context
When the sendMessage method catches an InterruptedException, it previously swallowed the interrupt without re-asserting the thread’s interrupt status. This can break cooperative interruption in multi-threaded environments and was flagged by SonarQube as a violation of best practices. Re-asserting the interrupt ensures that higher-level code can still detect and respond to the interruption.

## How Has This Been Tested?
The change has been verified by running a full SonarQube scan on the project after the update. No new issues or warnings were reported, confirming that the change resolves the previously flagged issue related to interrupt handling without introducing regressions.

## Breaking Changes
No. This change only affects internal error handling in the sendMessage method and does not modify any public APIs or require changes in user code or configurations.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
This change makes the SDK more robust by properly handling InterruptedException. According to standard Java practices, when an InterruptedException is caught, the thread’s interrupt status should be preserved — which we now do by calling Thread.currentThread().interrupt(). This also resolves a SonarQube warning related to rule java:S2142.

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
@@ -404,6 +404,7 @@ public class HttpClientSseClientTransport implements McpClientTransport {
 			}
 		}
 		catch (InterruptedException e) {
+			Thread.currentThread().interrupt(); // Preserve interrupt status
 			return Mono.error(new McpError("Failed to wait for the message endpoint"));
 		}
 


chemicL:
Although in principle the recommendation is right, I view this situation rather as a problem with the implementation of `HttpClientSsseClientTransport`. It is making a blocking call that can get interrupted inside a non-blocking, reactive method. That is something we should address and it has been raised as #115.

Issue #224, opened 2025-05-07
Please add Tool/ToolParam annotations to java sdk


scottslewis:
**Expected Behavior**

It would be ideal to have Tool annotation classes (specifically Tool/ToolParam) be added to the mcp-java-sdk.

**Current Behavior**

Multiple, tool-specific annotations exist, the most prominent being spring's:

https://github.com/spring-projects/spring-ai/tree/main/spring-ai-model/src/main/java/org/springframework/ai/tool/annotation

These annotations (and the one interface dependency) should be a part of the java mcp-sdk itself, rather than require spring dependencies (or other emerging implementations).

**Context**

I would like to create a dynamic MCP server using only OSGi services/declarative services, and use Tool/ToolParam annotations for binding OSGi services to MCP tools.  Having Tool/ToolParam annotations as part of the mcp itself rather than spring or other vendor-specific framework would aid in that effort.

scottslewis:
Please see pr #270

scottslewis:
I've use the classes described to demonstrate a MCP Server <-> Tool Group service example described [here](https://eclipseecf.blogspot.com/2025/06/remote-tools-for-model-context-protocol.html).   A consistent set of annotation types allows developers to not be tied to any particular server <-> server framework (or transport protocol).

scottslewis:
Could someone on the committer team review the associated pr #270 ?

tzolov:
Hi @scottslewis,

Have you tried the https://github.com/spring-ai-community/mcp-annotations project? 

It provides annotation-based method handling for MCP servers in Java, built on the [MCP Java SDK](https://github.com/modelcontextprotocol/sdk-java).

**Current annotations:**
- `@McpComplete` - completion functionality 
- `@McpPrompt` - prompt message generation
- `@McpResource` - resource access
- `@McpTool` - tool implementation with automatic JSON schema generation
- `@McpToolParam` - parameter descriptions and requirements
- `@McpLoggingConsumer` - logging message notifications
- `@McpSampling` - sampling requests
- `@McpArg` - method parameters as MCP arguments

(Related [example](https://github.com/spring-projects/spring-ai-examples/tree/main/model-context-protocol/mcp-annotations-server) that uses the various annotations)

The core `mcp-annotations` module only depends on the MCP Java SDK and common Java libraries. `mcp-annotations-spring` adds Spring integration features.

**Goal:** Graduate `mcp-annotations` as an optional module(s) in the MCP Java SDK. 

[...]

PR #235, opened 2025-05-13
Annotation classes to support variety of frameworks/tools

scottslewis:
This pr defines two annotation classes (Tool and ToolParam) and one support interface class ToolCallResultConverter.   

As per issue #224 it would be helpful if the mcp java sdk had these annotation classes so that tools builders didn't need to create their own/duplicate separate annotations for each tool suite.  As things are now,  java tooling and framework builders are creating their own Tool and ToolParam annotation classes, making it difficult to be tooling vendor independent wrt building mcp servers.

## Motivation and Context
Having these classes as part of the mcp sdk would allow vendor neutral mcp servers to be more easily created.

## Breaking Changes
This should supersede the tooling and framework-specific annotation classes each framework vendor is using.

- [ X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Additional context

Obviously, these annotation classes are very much like the vendor-specific versions currently being deployed.   I would suggest that this change be made before/as part of 1.0.0 release of mcp sdk.

I've intentionally left off any copyright notice, license, or documentation.  Any appropriate for this project is fine with me.  I would expect that wrt documentation links to the specification fields (e.g. description) could be provided.   I would not object to changing package structure or other naming.  This is just a suggestion

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/annotations/Tool.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/annotations/Tool.java
@@ -1,0 +1,15 @@
+package io.modelcontextprotocol.annotations;
+
+import io.modelcontextprotocol.util.ToolCallResultConverter;
+
+public @interface Tool {
+
+	String name() default "";
+
+	String description() default "";
+
+	boolean returnDirect() default false;
+
+	Class<? extends ToolCallResultConverter> resultConverter();
+
+}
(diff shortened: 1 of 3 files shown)


scottslewis:
I've significantly simplified the design of the annotation classes.   Now, they only provide the McpSchema meta-data as annotation properties.   e.g. for Tool interface:

```

@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.METHOD)
public @interface Tool {

	String name() default "";

	String description() default "";

	ToolAnnotation[] annotations() default {};

}

```

Also similarly simplified the ToolParam annotation:

```
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.PARAMETER)
public @interface ToolParam {

	String name() default "";

	String description() default "";

	boolean required() default true;

}

```

and added optional ToolResult annotation

Also created classes:  io.modelcontextprotocol.tools.util.*Description classes that use reflection to create instances with annotation content at runtime.

Issue #261, opened 2025-05-23
Incorrect URI resolution


abhimoondra:
**Bug description**
The resolved request url isn't correct for SSE endpoint

**Environment**
MCP Java SDK Version : 0.10.0
Java : 24

**Steps to reproduce**
The following function returns a value which is incorrect: https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp/src/main/java/io/modelcontextprotocol/util/Utils.java#L71-L79

So for base url as : `http://localhost:5555/sse`
and endpoint url as : `?sessionId=f1decbff-7929-40e8-b872-887425beb6eb`

The result is : `http://localhost:5555/?sessionId=f1decbff-7929-40e8-b872-887425beb6eb`

**Expected behavior**
I expect the result should be `http://localhost:5555/sse?sessionId=f1decbff-7929-40e8-b872-887425beb6eb`

**Minimal Complete Reproducible example**
```java
URI base = new URI("http://localhost:5555/sse");
URI relative = new URI("?sessionId=f1decbff-7929-40e8-b872-887425beb6eb");
URI resolved = base.resolve(relative.toString());
System.out.println(resolved);
```

denniskawurek:
I created a first draft for this and appreciate any feedback whether this can be followed this way. I know there's work ongoing for the Streamable HTTP support and want to avoid any duplicated work  / conflicts in the code.

chemicL:
I think this is a duplicate of #257 and in order to have a single issue I propose this one be closed. If I am mistaken, please reopen and provide feedback.

PR #262, opened 2025-05-23
fix: Resolve URIs

denniskawurek:
## Motivation and Context

Currently, the URL resolving is broken under different circumstances when adding a base URL to the path.

There are quite a few issues regarding this in this repo, but also in the Spring AI repos. Example: https://github.com/modelcontextprotocol/java-sdk/issues/261

Basically, the main issue is that the `resolveUri` method works for the message endpoint, but not for the SSE endpoint. So an URL resolver for the SSE endpoint is added.

## How Has This Been Tested?
See `WebMvcSseCustomPathIntegrationTests`

## Types of changes
- [ ] Documentation update

## Checklist

## Open questions

This PR is currently **WIP**, because for now it is a little bit unclear for how to handle the context path.
I guess this is _not_ the same as the `baseUrl`, but the main path of the servlet to reach? As an example, for SSE we would get:

```
http://localhost + PORT + CONTEXT_PATH + BASE_URL + SSE
```

If this is correct, then some additional changes need to be made to support this.

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java
+++ b/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java
@@ -154,6 +154,10 @@ public class WebFluxSseServerTransportProvider implements McpServerTransportProv
 		Assert.notNull(messageEndpoint, "Message endpoint must not be null");
 		Assert.notNull(sseEndpoint, "SSE endpoint must not be null");
 
+		if (baseUrl.endsWith("/")) {
+			baseUrl = baseUrl.substring(0, baseUrl.length() - 1);
+		}
+
 		this.objectMapper = objectMapper;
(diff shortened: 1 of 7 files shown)


denniskawurek:
I found some time to dig deeper into it and pushed changes which fix the behaviour for the Webflux and WebMvc transport providers.

One of the main issues is that the sse connection returns the message endpoint to the client. This endpoints needs to be the _full path_ with the context under which the server is running.
The downside is that another parameter is needed in the constructors to take this endpoint.

Unfortunately this can't be part of the `baseUrl`, because it is not really part of the URL and will break the mapping.

However, I tested running a server with Spring-AI with configurations like:

```yml
spring.ai.mcp.server:
  type: ASYNC
  base-url: /mcp
  sse-endpoint: /connect

spring:
  webflux:
    base-path: /v1/api
```

and instead of `spring.webflux` for webmvc:

```yml
servlet:
 context-path: /v1/api
```

With this solution I can connect succesfully to such routes after applying small changes in the Spring - AI code.

A big clean up of the code is still needed, though.

Edit: Cleanup done and PR marked as ready for review.

PR #271, opened 2025-05-27
feat: implement support for elicitation

LucaButBoring:
Implementation for the new elicitation spec (https://github.com/modelcontextprotocol/modelcontextprotocol/pull/382).

## Motivation and Context
Supports elicitation, which is now merged into the spec.

## How Has This Been Tested?
Thorough unit and integration tests (matched sampling test cases).

## Breaking Changes
None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
The Python SDK also has a PR that we can compare against (https://github.com/modelcontextprotocol/python-sdk/pull/625).

The SDK data models for schema objects (used in `ElicitRequest`) are a bit complicated here - similar to `Content`, schema objects are polymorphic and have a `type` property, but this doesn't always allow us to directly select a subtype, requiring us to use a custom deserializer for this. The main issue is `EnumSchema`, which shares the `string` type with `StringSchema`, but has an `enum` field to distinguish it, which can't be modeled by the `JsonSubType` annotation. Tests have been implement for both serialization and deserialization for this type to make sure we can handle this correctly.


tzolov:
Thank you @LucaButBoring, 

I will try to review this PR 

In the future, please consider the [contributing guidelines](https://github.com/modelcontextprotocol/java-sdk/blob/main/CONTRIBUTING.md) before implementing a major feature. 
> For non-trivial changes, please clarify with the maintainers in an issue whether you can contribute the change and the desired scope of the change.

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -10,14 +10,12 @@
 import java.util.List;
 import java.util.Map;
 
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.annotation.JsonSubTypes;
-import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.annotation.*;
```
tzolov: No wildcard imports

tzolov (changes requested):
Nice work @LucaButBoring,
There are minor formatting issues with the import to fix. 

I'm wondering if we should enforce PrimitiveSchemaDefinition types in the schema or just use Map<String, Object> instead?
We don't use these schemas for argument validation and the schema builders can still be used as external utilities, what do you think?

I guess we can also implement a dynamic schema generation from Java Class or records. But this will be task for another PR

mcp/src/test/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProviderIntegrationTests.java
```diff
@@ -339,6 +341,220 @@ void testCreateMessageWithRequestTimeoutFail() throws InterruptedException {
...
+
+					exchange.createElicitation(mock(ElicitRequest.class)).block();
+
+					return Mono.just(mock(CallToolResult.class));
+				});
+
+		var server = McpServer.async(mcpServerTransportProvider).serverInfo("test-server", "1.0.0").tools(tool).build();
+
+		try (
+				// Create client without sampling capabilities
```
tzolov: samping -> elicitation

LucaButBoring:
Thanks for reviewing, will fix the formatting issues within the next couple of hours once I get a chance.

---

Regarding the schema data model, the reason I went with a well-typed representation was because unlike tool input and output schemas, which are simply typed as arbitrary JSONSchema data in the specification, the kind of schema used in elicitation is actually a subset of JSONSchema, and is well-defined in the spec.

In particular, the spec's elicitation schema is strictly one level deep, disallowing `object` or `array` properties and having no concept of `$defs` or `$ref`, among other limitations. Modeling these restrictions in the SDK directly makes sense IMO, as opposed to full JSONSchema objects which are more or less arbitrary. If you think it still makes sense to drop down to `Map` in spite of that, I'm happy to change it, though.

We could also potentially reuse the `Schema` hierarchy (or a subset of it if we add `object`/`array`) for modeling JSONSchema generally in the future, too.

tzolov:
Thanks for the update @LucaButBoring , 
I understand your intention for type-safe elicitation schema (as json schema subset). But my point is that no-one is enforcing argument validation against this format! 
Because of this the McpSchema::Elicitation feels more like a builder utility rather an a schema.

LucaButBoring:
Fair enough, it can be revisited as part of some sort of unified validation support in the future, then — will swap it out with a Map for now.

LucaButBoring:
Updated to remove, ready for re-review.

tzolov:
Thank you very match @LucaButBoring  
Great contribution!

tzolov:
Squashed and merged at 8a5a591d39256ba3947003ec4477e1722363eb35

PR #272, opened 2025-05-28
Fix: Handle SSE ping messages in WebFluxSseClientTransport

goecho:
- Add check for SSE ping messages from fastMCP server
- Log debug message when ping received instead of throwing exception
- Complete the stream on ping events
- Matches behavior of Python SDK which treats pings as keep-alive

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebFluxSseClientTransport.java
+++ b/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebFluxSseClientTransport.java
@@ -3,10 +3,6 @@
  */
 package io.modelcontextprotocol.client.transport;
 
-import java.io.IOException;
-import java.util.function.BiConsumer;
-import java.util.function.Function;
-
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.modelcontextprotocol.spec.McpClientTransport;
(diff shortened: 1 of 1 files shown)


tzolov:
Hi @goecho , can you please elaborate on this. I'm not able to find `ping` event type mentioned by the specification?

goecho:
Hi @tzolov this behavior is not defined in the MCP protocol itself—it's specific to the Python implementation using the fastmcp library. The behavior comes from the keep-alive mechanism in its dependency, the sse-starlette library. You can refer to the code here: https://github.com/sysid/sse-starlette/blob/main/sse_starlette/sse.py#L214

YunKuiLu:
Hi @tzolov @goecho , I had the same issue and opened PR #223  to dealt with it by adding an `sseErrorHandler` parameter.
The SSE comment message `: ping` isn't part of the MCP spec, so I'm thinking of just logging it as a warning by default. How does that sound?

goecho:
Hi @YunKuiLu ,I don't think this is a good idea — we shouldn't change the original logic. The original approach was better: throwing an exception for unknown types helps expose issues, which is better than just logging them.

YunKuiLu:
@goecho Thank you for your reply.

I actually considered both options: throwing an exception to stop the client or just logging it. But since the issue is only logged in `HttpClientSseClientTransport`, I think the client should keep the same default behavior for consistency.

https://github.com/modelcontextprotocol/java-sdk/blob/d9006ecb78a40b1c551bad08cacaa483eb83fcca/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java#L363-L365

osenx:
I think we can combine two ways. First, expand the ping type, and then use the error handler to support any errors

schengyi:
Hi @ tzolov, ping is not an event, but a comment in the sse protocol that starts with ":". The official handling method is: it should be ignored.
However, the buildEvent method in org.springframework.cttp.codec.ServerSetEventHttpMessage Reader intercepts the content after ":" and puts it in the comment, but the event in ServerSentEvent is empty, which can cause an exception to be thrown.

I think PR should be merged immediately because sending ping messages is present in many frameworks.

[event-stream-interpretation](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation)
image

exception
image

tzolov:
@chengyi3192 thanks for clarifying the semantics of the comment SSE (non)events.

tzolov:
@goecho I guess we can strengthen the check to verify that the `id`, `event` and `data` fields are empty as well?  And perhaps generalize to `isCommentLine`

goecho:
@tzolov Great point — I've now added the checks for id, event, and data, and pulled the logic into an isCommentLine() helper. Thanks!

PR #284, opened 2025-05-30
Add support for DNS rebinding protections

ddworken:
## Motivation and Context
This implements the mitigations described [here](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#security-warning). To avoid breaking existing applications this doesn't enable any changes by-default, but enabling this feature is heavily encouraged for any local MCP servers using the SSE transport. 

## How Has This Been Tested?
Tested via unit tests. 

## Breaking Changes
To avoid introducing any breaking changes, the DNS rebinding protections are disabled by default. Ideally we should find a way to enable them by default. But for now, adding them as disabled is a good first step. 

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java
+++ b/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java
@@ -1,6 +1,7 @@
 package io.modelcontextprotocol.server.transport;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -111,6 +112,11 @@ public class WebFluxSseServerTransportProvider implements McpServerTransportProv
 	private volatile boolean isClosing = false;
 
 	/**
+	 * DNS rebinding protection configuration.
+	 */
+	private final DnsRebindingProtectionConfig dnsRebindingProtectionConfig;
+
+	/**
 	 * Constructs a new WebFlux SSE server transport provider instance with the default
 	 * SSE endpoint.
 	 * @param objectMapper The ObjectMapper to use for JSON serialization/deserialization
@@ -134,7 +140,7 @@ public class WebFluxSseServerTransportProvider implements McpServerTransportProv
 	 * @throws IllegalArgumentException if either parameter is null
 	 */
 	public WebFluxSseServerTransportProvider(ObjectMapper objectMapper, String messageEndpoint, String sseEndpoint) {
-		this(objectMapper, DEFAULT_BASE_URL, messageEndpoint, sseEndpoint);
+		this(objectMapper, DEFAULT_BASE_URL, messageEndpoint, sseEndpoint, null);
 	}
 
 	/**
@@ -149,6 +155,24 @@ public class WebFluxSseServerTransportProvider implements McpServerTransportProv
 	 */
 	public WebFluxSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
 			String sseEndpoint) {
+		this(objectMapper, baseUrl, messageEndpoint, sseEndpoint, null);
+	}
+
+	/**
+	 * Constructs a new WebFlux SSE server transport provider instance with optional DNS
+	 * rebinding protection.
+	 * @param objectMapper The ObjectMapper to use for JSON serialization/deserialization
+	 * of MCP messages. Must not be null.
+	 * @param baseUrl webflux message base path
+	 * @param messageEndpoint The endpoint URI where clients should send their JSON-RPC
+	 * messages. This endpoint will be communicated to clients during SSE connection
+	 * setup. Must not be null.
+	 * @param sseEndpoint The endpoint URI where clients establish their SSE connections.
+	 * @param dnsRebindingProtectionConfig The DNS rebinding protection configuration (may be null).
+	 * @throws IllegalArgumentException if required parameters are null
+	 */
+	public WebFluxSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
+			String sseEndpoint, DnsRebindingProtectionConfig dnsRebindingProtectionConfig) {
 		Assert.notNull(objectMapper, "ObjectMapper must not be null");
 		Assert.notNull(baseUrl, "Message base path must not be null");
 		Assert.notNull(messageEndpoint, "Message endpoint must not be null");
@@ -158,6 +182,7 @@ public class WebFluxSseServerTransportProvider implements McpServerTransportProv
 		this.baseUrl = baseUrl;
 		this.messageEndpoint = messageEndpoint;
 		this.sseEndpoint = sseEndpoint;
+		this.dnsRebindingProtectionConfig = dnsRebindingProtectionConfig;
 		this.routerFunction = RouterFunctions.route()
 			.GET(this.sseEndpoint, this::handleSseConnection)
 			.POST(this.messageEndpoint, this::handleMessage)
@@ -256,6 +281,16 @@ public class WebFluxSseServerTransportProvider implements McpServerTransportProv
 			return ServerResponse.status(HttpStatus.SERVICE_UNAVAILABLE).bodyValue("Server is shutting down");
 		}
 
+		// Validate headers
+		if (dnsRebindingProtectionConfig != null) {
+			String hostHeader = request.headers().asHttpHeaders().getFirst("Host");
+			String originHeader = request.headers().asHttpHeaders().getFirst("Origin");
+			if (!dnsRebindingProtectionConfig.validate(hostHeader, originHeader)) {
+				logger.warn("DNS rebinding protection validation failed - Host: '{}', Origin: '{}'", hostHeader, originHeader);
+				return ServerResponse.status(HttpStatus.FORBIDDEN).bodyValue("DNS rebinding protection validation failed");
+			}
+		}
+
 		return ServerResponse.ok()
 			.contentType(MediaType.TEXT_EVENT_STREAM)
 			.body(Flux.<ServerSentEvent<?>>create(sink -> {
@@ -298,6 +333,25 @@ public class WebFluxSseServerTransportProvider implements McpServerTransportProv
 	private Mono<ServerResponse> handleMessage(ServerRequest request) {
 		if (isClosing) {
 			return ServerResponse.status(HttpStatus.SERVICE_UNAVAILABLE).bodyValue("Server is shutting down");
+		}
+
+		// Always validate Content-Type for POST requests
+		String contentType = request.headers().contentType()
+			.map(MediaType::toString)
+			.orElse(null);
+		if (contentType == null || !contentType.toLowerCase().startsWith("application/json")) {
+			logger.warn("Invalid Content-Type header: '{}'", contentType);
+			return ServerResponse.badRequest().bodyValue(new McpError("Content-Type must be application/json"));
+		}
+
+		// Validate headers for POST requests if DNS rebinding protection is configured
+		if (dnsRebindingProtectionConfig != null) {
+			String hostHeader = request.headers().asHttpHeaders().getFirst("Host");
+			String originHeader = request.headers().asHttpHeaders().getFirst("Origin");
+			if (!dnsRebindingProtectionConfig.validate(hostHeader, originHeader)) {
+				logger.warn("DNS rebinding protection validation failed - Host: '{}', Origin: '{}'", hostHeader, originHeader);
+				return ServerResponse.status(HttpStatus.FORBIDDEN).bodyValue("DNS rebinding protection validation failed");
+			}
 		}
 
 		if (request.queryParam("sessionId").isEmpty()) {
@@ -397,6 +451,8 @@ public class WebFluxSseServerTransportProvider implements McpServerTransportProv
 
 		private String sseEndpoint = DEFAULT_SSE_ENDPOINT;
 
+		private DnsRebindingProtectionConfig dnsRebindingProtectionConfig;
+
 		/**
 		 * Sets the ObjectMapper to use for JSON serialization/deserialization of MCP
 		 * messages.
@@ -447,6 +503,23 @@ public class WebFluxSseServerTransportProvider implements McpServerTransportProv
 			return this;
 		}
 
+
+		/**
+		 * Sets the DNS rebinding protection configuration.
+		 * <p>
+		 * When set, this configuration will be used to create a header validator that
+		 * enforces DNS rebinding protection rules. This will override any previously set
+		 * header validator.
+		 * @param config The DNS rebinding protection configuration
+		 * @return this builder instance
+		 * @throws IllegalArgumentException if config is null
+		 */
+		public Builder dnsRebindingProtectionConfig(DnsRebindingProtectionConfig config) {
+			Assert.notNull(config, "DNS rebinding protection config must not be null");
+			this.dnsRebindingProtectionConfig = config;
+			return this;
+		}
+
 		/**
 		 * Builds a new instance of {@link WebFluxSseServerTransportProvider} with the
 		 * configured settings.
@@ -457,7 +530,8 @@ public class WebFluxSseServerTransportProvider implements McpServerTransportProv
 			Assert.notNull(objectMapper, "ObjectMapper must be set");
 			Assert.notNull(messageEndpoint, "Message endpoint must be set");
 
-			return new WebFluxSseServerTransportProvider(objectMapper, baseUrl, messageEndpoint, sseEndpoint);
+			return new WebFluxSseServerTransportProvider(objectMapper, baseUrl, messageEndpoint, sseEndpoint,
+					dnsRebindingProtectionConfig);
 		}
 
 	}
--- a/mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcSseServerTransportProvider.java
+++ b/mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcSseServerTransportProvider.java
@@ -6,6 +6,7 @@ package io.modelcontextprotocol.server.transport;
 
 import java.io.IOException;
 import java.time.Duration;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
@@ -108,6 +109,11 @@ public class WebMvcSseServerTransportProvider implements McpServerTransportProvi
 	private volatile boolean isClosing = false;
 
 	/**
+	 * DNS rebinding protection configuration.
+	 */
+	private final DnsRebindingProtectionConfig dnsRebindingProtectionConfig;
+
+	/**
 	 * Constructs a new WebMvcSseServerTransportProvider instance with the default SSE
 	 * endpoint.
 	 * @param objectMapper The ObjectMapper to use for JSON serialization/deserialization
@@ -132,7 +138,7 @@ public class WebMvcSseServerTransportProvider implements McpServerTransportProvi
 	 * @throws IllegalArgumentException if any parameter is null
 	 */
 	public WebMvcSseServerTransportProvider(ObjectMapper objectMapper, String messageEndpoint, String sseEndpoint) {
-		this(objectMapper, "", messageEndpoint, sseEndpoint);
+		this(objectMapper, "", messageEndpoint, sseEndpoint, null);
 	}
 
 	/**
@@ -149,6 +155,24 @@ public class WebMvcSseServerTransportProvider implements McpServerTransportProvi
 	 */
 	public WebMvcSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
 			String sseEndpoint) {
+		this(objectMapper, baseUrl, messageEndpoint, sseEndpoint, null);
+	}
+
+	/**
+	 * Constructs a new WebMvcSseServerTransportProvider instance with DNS rebinding protection.
+	 * @param objectMapper The ObjectMapper to use for JSON serialization/deserialization
+	 * of messages.
+	 * @param baseUrl The base URL for the message endpoint, used to construct the full
+	 * endpoint URL for clients.
+	 * @param messageEndpoint The endpoint URI where clients should send their JSON-RPC
+	 * messages via HTTP POST. This endpoint will be communicated to clients through the
+	 * SSE connection's initial endpoint event.
+	 * @param sseEndpoint The endpoint URI where clients establish their SSE connections.
+	 * @param dnsRebindingProtectionConfig The DNS rebinding protection configuration (may be null).
+	 * @throws IllegalArgumentException if any required parameter is null
+	 */
+	public WebMvcSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
+			String sseEndpoint, DnsRebindingProtectionConfig dnsRebindingProtectionConfig) {
 		Assert.notNull(objectMapper, "ObjectMapper must not be null");
 		Assert.notNull(baseUrl, "Message base URL must not be null");
 		Assert.notNull(messageEndpoint, "Message endpoint must not be null");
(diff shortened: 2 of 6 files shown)


mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
```diff
@@ -0,0 +1,108 @@
...
+ * Configuration for DNS rebinding protection in SSE server transports. Provides
+ * validation for Host and Origin headers to prevent DNS rebinding attacks.
+ */
+public class DnsRebindingProtectionConfig {
+
+	private final Set<String> allowedHosts;
+
+	private final Set<String> allowedOrigins;
+
+	private final boolean enableDnsRebindingProtection;
```
chemicL:
```suggestion
	private final boolean enabled;
```

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
```diff
@@ -0,0 +1,108 @@
...
+
+	private DnsRebindingProtectionConfig(Builder builder) {
```
chemicL: Consider passing explicit arguments instead of the `builder`. Should the `builder` change, and it is mutable, this constructor could create instances that are in inconsistent state. The pattern present throughout the codebase is rather to pass explicit parameters instead of this cyclic dependency.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		return true;
+	}
+
+	public static Builder builder() {
+		return new Builder();
+	}
+
+	public static class Builder {
```
chemicL: In order to avoid having two means to instantiate the `Builder`, the constructor should be private. A static factory exists and it is the pattern that's used by this project.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
```diff
@@ -0,0 +1,108 @@
...
+			String lowerOrigin = originHeader.toLowerCase();
+			if (!allowedOrigins.contains(lowerOrigin)) {
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	public static Builder builder() {
```
chemicL: Missing javadoc

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
chemicL: Please add javadocs to the public class and its methods.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		public Builder allowedOrigins(Set<String> origins) {
+			if (origins != null) {
+				origins.forEach(this::allowedOrigin);
+			}
+			return this;
+		}
+
+		public Builder enableDnsRebindingProtection(boolean enable) {
```
chemicL: This method name is useful, but internally the field can be just named `enabled` as we know the context here.

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
```diff
@@ -127,10 +130,27 @@ public HttpServletSseServerTransportProvider(ObjectMapper objectMapper, String m
...
+	 * rebinding protection.
+	 * @param objectMapper The JSON object mapper to use for message
+	 * serialization/deserialization
+	 * @param baseUrl The base URL for the server transport
+	 * @param messageEndpoint The endpoint path where clients will send their messages
+	 * @param sseEndpoint The endpoint path where clients will establish SSE connections
+	 * @param dnsRebindingProtectionConfig The DNS rebinding protection configuration (may
+	 * be null)
+	 */
+	public HttpServletSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
```
chemicL: Instead of adding a new public constructor, I'd suggest we
1. Deprecate the existing ones,
2. Suggest using the `Builder` in their javadocs,
3. Have this constructor private.

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
```diff
@@ -252,6 +284,26 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
+		// Validate headers for POST requests if DNS rebinding protection is configured
+		if (dnsRebindingProtectionConfig != null) {
+			String hostHeader = request.getHeader("Host");
+			String originHeader = request.getHeader("Origin");
+			if (!dnsRebindingProtectionConfig.validate(hostHeader, originHeader)) {
+				logger.warn("DNS rebinding protection validation failed - Host: '{}', Origin: '{}'", hostHeader,
+						originHeader);
+				response.sendError(HttpServletResponse.SC_FORBIDDEN, "DNS rebinding protection validation failed");
+				return;
+			}
```
chemicL: Suppose this can be extracted to a method as it's also duplicated in the GET handling.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
```diff
@@ -0,0 +1,108 @@
...
+		return new Builder();
+	}
+
+	public static class Builder {
+
+		private final Set<String> allowedHosts = new HashSet<>();
+
+		private final Set<String> allowedOrigins = new HashSet<>();
+
+		private boolean enableDnsRebindingProtection = true;
```
chemicL: This ought to be documented in the builder and I believe it should also be the default value in the actual config class.

mcp/src/test/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfigTests.java
```diff
@@ -0,0 +1,157 @@
...
+ * Unit tests for DNS rebinding protection configuration.
+ */
+public class DnsRebindingProtectionConfigTests {
+
+	@Test
+	void testDefaultConfiguration() {
+		DnsRebindingProtectionConfig config = DnsRebindingProtectionConfig.builder().build();
+
+		// Test default behavior - when allowed lists are empty and headers are provided,
+		// validation fails because the headers are not in the (empty) allowed lists
```
chemicL: This needs to be documented in the class javadoc.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
```diff
@@ -0,0 +1,108 @@
+package io.modelcontextprotocol.server.transport;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Configuration for DNS rebinding protection in SSE server transports. Provides
+ * validation for Host and Origin headers to prevent DNS rebinding attacks.
```
chemicL: The description should be setting some expectations as to how it can be used and when it will fail to validate and when it will succeed.

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
```diff
@@ -475,6 +527,8 @@ public static class Builder {
 
 		private String sseEndpoint = DEFAULT_SSE_ENDPOINT;
 
+		private DnsRebindingProtectionConfig dnsRebindingProtectionConfig;
```
chemicL: Shouldn't the default behaviour be injected always unless the user specifically overrides it?

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
```diff
@@ -202,6 +222,18 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response)
 			return;
 		}
 
+		// Validate headers if DNS rebinding protection is configured
+		if (dnsRebindingProtectionConfig != null) {
+			String hostHeader = request.getHeader("Host");
```
chemicL: To make it work with HTTP2, shouldn't https://jakarta.ee/specifications/servlet/6.0/apidocs/jakarta.servlet/jakarta/servlet/servletrequest#getServerName() be used instead?

mcp/src/test/java/io/modelcontextprotocol/server/transport/HttpServletSseHeaderValidationTests.java
chemicL, line 1: Please check `AbstractMcpAsyncServerTests` and the implementations. Do you think we could be able to have a generic test suite with just the transport-specific settings in the inheriting classes in the same fashion?

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+public class DnsRebindingProtectionConfig {
```
chemicL: `DnsRebindingProtectionConfig` -> `DnsRebindingProtection`.
I view this more as an entity that has behaviour via the `validate()` method, not just a set of configurations, so I'd skip the "Config" part.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	/**
+	 * Validates Host and Origin headers for DNS rebinding protection. Returns true if the
+	 * headers are valid, false otherwise.
+	 * @param hostHeader The value of the Host header (may be null)
+	 * @param originHeader The value of the Origin header (may be null)
+	 * @return true if the headers are valid, false otherwise
+	 */
+	public boolean validate(String hostHeader, String originHeader) {
```
chemicL:
```suggestion
	public boolean isValid(String hostHeader, String originHeader) {
```

chemicL: I would expect a `validate` method to throw an exception or have some effect instead. Since the return is a boolean, I think it's more reasonable to call the method `isValid`.

chemicL (changes requested):
Thank you for the PR. I did the first round by going through the code in the mcp module. I believe the comments would apply also to the mcp-spring modules. Some questions, some suggestions here and there, but the most important themes are:
1. HTTP(2) protocol compliance - will how to make it work when there is `:authority` pseudo-header instead of `Host`?
2. MCP Spec compliance - does `Host` require checking? The spec only mentions `Origin`.
3. Since validation is enforced by the spec as a MUST i do feel we should enable validation with a breaking change. @tzolov looking forward to your input here as well.
4. Builder pattern consistency.
5. Javadocs improvements and filling the gaps.
6. Removing code duplication.

Thanks in advance for a follow-up and the work so far!

chemicL:
Ah, please also remember to run `spring-javaformat:apply` before commiting to ensure the formatting checks pass.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
jgrandja: I'm not sure the name should be `DnsRebindingProtection` because the attributes provided in it would be limited only to the DNS Rebinding Protection use case.

For example, the following:

```java
public class McpServerTransportSettings {

    private Set<String> allowedHostnames;

}
```

is more general and can be reused in other security checks, since host name is often used in other appsec mitigations.

I also don't feel the behaviour (`validate()`) and data (`allowedHostnames`) should be mixed. IMO, it's ok to duplicate the `validate()` logic internally in the `McpServerTransportProvider` implementations and simply reuse the data, e.g. `McpServerTransportSettings`.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
```diff
@@ -0,0 +1,108 @@
...
+
+	public static Builder builder() {
+		return new Builder();
+	}
+
+	public static class Builder {
+
+		private final Set<String> allowedHosts = new HashSet<>();
+
+		private final Set<String> allowedOrigins = new HashSet<>();
```
jgrandja: Even though the spec states:

> Servers **MUST** validate the `Origin` header on all incoming connections to prevent DNS rebinding attacks

It doesn't make sense to me because `Origin` is NOT required and therefore can be empty depending on the type of request and browser behaviour.

I don't think validating the `Origin` header is valid. We should only have to validate the `Host` header.

DNS rebinding is all about changing the underlying IP address associated to a domain name (`Host`). For example, http://example.com initially points to `34.192.228.43` but can be dynamically changed to a private internal address of `192.168.1.77` enabled the attack on the internal network. For a more detailed description, check out [this article](https://medium.com/@brannondorsey/attacking-private-networks-from-the-internet-with-dns-rebinding-ea7098a2d325).

In summary, the MCP server that receives the request **should validate that it’s own host matches the host being requested.**

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
ddworken: Done, I renamed it to `DnsRebindingProtection`. Personally I agree with @chemicL that it is cleaner to keep this focused on DNS rebinding and keep the data and `isValid()` in the same place.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
ddworken: Done.

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
ddworken: Done.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
ddworken: Done.

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
ddworken: Done.

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
ddworken: Done.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
ddworken: Done.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
ddworken: Done.

ddworken: Done.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
ddworken: Added docs on it. I'm not sure there is any point in having a default value in the parent class though since it is always set in the constructor?

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
ddworken: That would make this a breaking change. I do see the argument for shipping this breaking change but IMO that is best saved for another PR to make this PR a standalone non-breaking change.

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
ddworken: > the MCP server that receives the request should validate that it’s own host matches the host being requested.

I agree that this is the most critical part of DNS rebinding protection, though it is pretty common to also check the Origin header since it provides another layer of defense and also mitigates a variety of other web-based attacks. 

> It doesn't make sense to me because `Origin` is NOT required and therefore can be empty depending on the type of request and browser behaviour.
> 
> I don't think validating the `Origin` header is valid. We should only have to validate the `Host` header.

Modern browsers send the `Origin` header for ~all cross-origin requests with the main exceptions being simple GET requests and navigational requests. So for the purposes of MCP, I believe it is valid and should always be set.

ddworken:
> Since validation is enforced by the spec as a MUST i do feel we should enable validation with a breaking change. @tzolov looking forward to your input here as well.

I'll say I think there is a pretty good argument for enabling validation by default, but since that is a breaking change I think it ought to:

1. Be deferred to a separate PR 
2. Be done in a unified way across the MCP ecosystem

mcp/src/main/java/io/modelcontextprotocol/server/transport/DnsRebindingProtectionConfig.java
jgrandja: @ddworken 

> though it is pretty common to also check the Origin header

Can you please provide references to other MCP implementations that validate the `Origin` header so I can further investigate.

> and also mitigates a variety of other web-based attacks

What other web-based attacks are you referring to?

ddworken: > Can you please provide references to other MCP implementations that validate the `Origin` header so I can further investigate.

* https://github.com/modelcontextprotocol/typescript-sdk/pull/565
* https://github.com/modelcontextprotocol/python-sdk/pull/861

We also are hoping to enable this for other MCP implementations too. 

> > and also mitigates a variety of other web-based attacks
> 
> What other web-based attacks are you referring to?

E.g. CSRF attacks. If an MCP server accepts [simple requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#simple_requests) then it can be CSRF attacked, and validating the Origin header blocks this.

jgrandja: The 2x references you provided are PR's submitted by you and only merged last month. Do you have other references where the `Origin` header is validated to mitigate DNS Rebinding attack?

> E.g. CSRF attacks. If an MCP server accepts [simple requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#simple_requests) then it can be CSRF attacked, and validating the Origin header blocks this.

This is not correct. The recommended mitigation for CSRF attack are the use of tokens. See [token-based mitigation](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#token-based-mitigation).

ddworken: I definitely agree that checking the `Origin` header is not the primary defense against DNS rebinding. But it does offer a second layer of security that helps add defense in depth against DNS rebinding, in addition to also offering mitigations against CSRF (see below). So I do think this is something worth adding to this SDK (and is also worth adding to other MCP SDKs). IMO the main question here is maybe we should rename `DnsRebindingProtection` to `WebSecurityProtection` to make it clear that the origin checking also benefits web security more broadly. WDYT? 

> This is not correct. The recommended mitigation for CSRF attack are the use of tokens. See [token-based mitigation](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#token-based-mitigation).

Nowadays, checking the `Origin` header is a widely relied upon security feature. While tokens do offer a number of appealing advantages that can make them more versatile than solely checking the `Origin` header, the `Origin` header is unspoofable and is safe to rely upon in this way. OWASP talks about this some [here](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#using-standard-headers-to-verify-origin) as do [these Mozilla docs](https://wiki.mozilla.org/Security/Origin).

ddworken:
@chemicL could I get your help with landing this PR? What things are remaining here?

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
```diff
@@ -327,6 +371,37 @@ public Mono<Void> closeGracefully() {
...
+	 * Uses {@link HttpServletRequest#getServerName()} and
+	 * {@link HttpServletRequest#getServerPort()} instead of the Host header for HTTP/2
+	 * compatibility.
+	 * @param request The HTTP servlet request
+	 * @param response The HTTP servlet response
+	 * @return true if validation passed (or no validation needed), false if validation
+	 * failed and error was sent
+	 * @throws IOException If an I/O error occurs while sending error response
+	 */
+	private boolean validateDnsRebindingProtection(HttpServletRequest request, HttpServletResponse response)
```
chemicL, line 387: Please avoid boolean methods with side effects like this. The way forward that I see:

* `boolean passesDnsRebindingProtectionCheck(HttpServletRequest request)`
* `void failDnsRebindingProtection(HttpServletResponse response)`

And called like this:

```
if (!passesDnsRebindingProtectionCheck(request)) {
    failDnsRebindingProtection(response);
    return;
}
```

mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcSseServerTransportProvider.java
```diff
@@ -416,4 +467,23 @@ public void close() {
 
 	}
 
+	/**
+	 * Validates DNS rebinding protection for the given request.
+	 * @param request The incoming server request
+	 * @return A ServerResponse with forbidden status if validation fails, or null if
+	 * validation passes
+	 */
+	private ServerResponse validateDnsRebindingProtection(ServerRequest request) {
```
chemicL, line 476: I believe a similar strategy as in the `HttpServletSseServerTransportProvider` can be applied here (isValid + fail).

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java
```diff
@@ -255,6 +286,12 @@ private Mono<ServerResponse> handleSseConnection(ServerRequest request) {
 			return ServerResponse.status(HttpStatus.SERVICE_UNAVAILABLE).bodyValue("Server is shutting down");
 		}
 
+		// Validate headers
+		Mono<ServerResponse> validationError = validateDnsRebindingProtection(request);
```
chemicL, line 290: It's not a valid style to return a `null` `Mono` reference. If there's a synchronous action that does not block, simply return a `boolean` to drive the behaviour.

* `boolean passesDnsRebindingProtectionCheck(ServerRequest request)`
* `Mono<ServerResponse failDnsRebindingProtection()`

```
if (!passesDnsRebindingProtectionCheck(request)) {
    return failDnsRebindingProtection();
}
```

I think the style you're using in this PR is inspired by errors from Go. The style present in the codebase of this project should be consistent.

Kehrlann:
Implemented in #771 and #775 .

PR #290, opened 2025-06-02
Adding StreamableHttpServerTransportProvider class and unit tests

ZachGerman:
Still missing:
- Origin header validation (ongoing discussions about spec for that in Discord)
- Backward-compatible endpoint combo

## Motivation and Context
Trying to reach spec parity with TS and Python for Java. Will continue working on other aspects of this.

## How Has This Been Tested?
Unit tests and integ tests using a partially-complete sHTTP client transport class, but I didn't include the integ test file as the client isn't complete.

## Breaking Changes
N/A

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
This is part of a larger commit seen here: https://github.com/modelcontextprotocol/java-sdk/pull/289

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
@@ -1,0 +1,765 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ */
+
+package io.modelcontextprotocol.server.transport;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.util.Enumeration;
+import java.util.function.Supplier;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.modelcontextprotocol.spec.McpError;
+import io.modelcontextprotocol.spec.McpSchema;
+import io.modelcontextprotocol.spec.McpSchema.JSONRPCMessage;
+import io.modelcontextprotocol.spec.McpServerSession;
+import io.modelcontextprotocol.spec.McpServerTransport;
+import io.modelcontextprotocol.spec.McpServerTransportProvider;
+import io.modelcontextprotocol.util.Assert;
+import jakarta.servlet.AsyncContext;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.annotation.WebServlet;
+import jakarta.servlet.http.HttpServlet;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.core.publisher.Sinks;
+
+/**
+ * Implementation of the MCP Streamable HTTP transport provider for servers. This
+ * implementation follows the Streamable HTTP transport specification from protocol
+ * version 2025-03-26.
+ *
+ * <p>
+ * The transport handles a single HTTP endpoint that supports POST, GET, & DELETE methods:
+ * <ul>
+ * <li>POST - For receiving client messages and optionally establishing SSE streams for
+ * responses</li>
+ * <li>GET - For establishing SSE streams for server-to-client communication</li>
+ * <li>DELETE - For terminating sessions</li>
+ * </ul>
+ *
+ * <p>
+ * Features:
+ * <ul>
+ * <li>Session management with secure session IDs</li>
+ * <li>Support for resumable SSE streams</li>
+ * <li>Support for multiple concurrent client connections</li>
+ * <li>Graceful shutdown support</li>
+ * </ul>
+ *
+ */
+@WebServlet(asyncSupported = true)
+public class StreamableHttpServerTransportProvider extends HttpServlet implements McpServerTransportProvider {
+
+	private static final Logger logger = LoggerFactory.getLogger(StreamableHttpServerTransportProvider.class);
+
+	public static final String UTF_8 = "UTF-8";
+
+	public static final String APPLICATION_JSON = "application/json";
+
+	public static final String TEXT_EVENT_STREAM = "text/event-stream";
+
+	public static final String SESSION_ID_HEADER = "Mcp-Session-Id";
+
+	public static final String LAST_EVENT_ID_HEADER = "Last-Event-Id";
+
+	public static final String MESSAGE_EVENT_TYPE = "message";
+
+	public static final String ACCEPT_HEADER = "Accept";
+
+	public static final String ORIGIN_HEADER = "Origin";
+
+	public static final String CACHE_CONTROL_HEADER = "Cache-Control";
+
+	public static final String CONNECTION_HEADER = "Connection";
+
+	public static final String CACHE_CONTROL_NO_CACHE = "no-cache";
+
+	public static final String CONNECTION_KEEP_ALIVE = "keep-alive";
+
+	/** JSON object mapper for serialization/deserialization */
+	private final ObjectMapper objectMapper;
+
+	/** The endpoint path for handling MCP requests */
+	private final String mcpEndpoint;
+
+	/** Supplier for generating unique session IDs */
+	private final Supplier<String> sessionIdProvider;
+
+	/** UUID.randomUUID().toString() */
+	private static final Supplier<String> DEFAULT_SESSION_ID_PROVIDER = () -> UUID.randomUUID().toString();
+
+	/** Map of active client sessions, keyed by session ID */
+	private final Map<String, McpServerSession> sessions = new ConcurrentHashMap<>();
+
+	/** Map of active SSE streams, keyed by session ID */
+	private final Map<String, StreamableHttpSseStream> sseStreams = new ConcurrentHashMap<>();
+
+	/** Flag indicating if the transport is in the process of shutting down */
+	private final AtomicBoolean isClosing = new AtomicBoolean(false);
+
+	/** Session factory for creating new sessions */
+	private McpServerSession.Factory sessionFactory;
+
+	/**
+	 * Creates a new StreamableHttpServerTransportProvider instance.
+	 * @param objectMapper The JSON object mapper to use for message
+	 * serialization/deserialization
+	 * @param mcpEndpoint The endpoint path for handling MCP requests
+	 * @param sessionIdProvider optional Supplier for providing unique session IDs
+	 */
+	public StreamableHttpServerTransportProvider(ObjectMapper objectMapper, String mcpEndpoint,
+			Supplier<String> sessionIdProvider) {
+		Assert.notNull(objectMapper, "ObjectMapper must not be null");
+		Assert.hasText(mcpEndpoint, "MCP endpoint must not be empty");
+
+		this.objectMapper = objectMapper;
+		this.mcpEndpoint = mcpEndpoint;
+		this.sessionIdProvider = Objects.requireNonNullElse(sessionIdProvider, DEFAULT_SESSION_ID_PROVIDER);
+	}
+
+	@Override
+	public void setSessionFactory(McpServerSession.Factory sessionFactory) {
+		this.sessionFactory = sessionFactory;
+	}
+
+	@Override
+	public Mono<Void> notifyClients(String method, Object params) {
+		if (sessions.isEmpty()) {
+			logger.debug("No active sessions to broadcast message to");
+			return Mono.empty();
+		}
+
+		logger.debug("Attempting to broadcast message to {} active sessions", sessions.size());
+
+		return Flux.fromIterable(sessions.values())
+			.flatMap(session -> session.sendNotification(method, params)
+				.doOnError(
+						e -> logger.error("Failed to send message to session {}: {}", session.getId(), e.getMessage()))
+				.onErrorComplete())
+			.then();
+	}
+
+	@Override
+	public Mono<Void> closeGracefully() {
+		isClosing.set(true);
+		logger.debug("Initiating graceful shutdown with {} active sessions", sessions.size());
+
+		return Flux.fromIterable(sessions.values()).flatMap(McpServerSession::closeGracefully).then();
+	}
+
+	/**
+	 * Handles HTTP GET requests to establish SSE connections.
+	 * @param request The HTTP servlet request
+	 * @param response The HTTP servlet response
+	 * @throws ServletException If a servlet-specific error occurs
+	 * @throws IOException If an I/O error occurs
+	 */
+	@Override
+	protected void doGet(HttpServletRequest request, HttpServletResponse response)
+			throws ServletException, IOException {
+
+		String requestURI = request.getRequestURI();
+		logger.debug("GET request received for URI: {}", requestURI);
+
+		// Log all headers for debugging
+		Enumeration<String> headerNames = request.getHeaderNames();
+		while (headerNames.hasMoreElements()) {
+			String headerName = headerNames.nextElement();
+			logger.debug("Header: {} = {}", headerName, request.getHeader(headerName));
+		}
+
+		if (!requestURI.endsWith(mcpEndpoint)) {
+			logger.debug("URI does not match mcpEndpoint: {}", mcpEndpoint);
+			response.sendError(HttpServletResponse.SC_NOT_FOUND);
+			return;
+		}
+
+		if (isClosing.get()) {
+			logger.debug("Server is shutting down, rejecting request");
+			response.sendError(HttpServletResponse.SC_SERVICE_UNAVAILABLE, "Server is shutting down");
+			return;
+		}
+
+		String acceptHeader = request.getHeader(ACCEPT_HEADER);
+		logger.debug("Accept header: {}", acceptHeader);
+		if (acceptHeader == null || !acceptHeader.contains(TEXT_EVENT_STREAM)) {
+			logger.debug("Accept header missing or does not include {}", TEXT_EVENT_STREAM);
+			response.setContentType(APPLICATION_JSON);
+			response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+			response.getWriter().write(createErrorJson("Accept header must include text/event-stream"));
+			return;
+		}
+
+		String sessionId = request.getHeader(SESSION_ID_HEADER);
+		if (sessionId == null) {
+			response.setContentType(APPLICATION_JSON);
+			response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+			response.getWriter().write(createErrorJson("Session ID missing in request header"));
+			return;
+		}
+
+		McpServerSession session = sessions.get(sessionId);
+		if (session == null) {
+			response.setContentType(APPLICATION_JSON);
+			response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+			response.getWriter().write(createErrorJson("Session not found: " + sessionId));
+			return;
+		}
+
+		// Set up SSE connection
+		response.setContentType(TEXT_EVENT_STREAM);
+		response.setCharacterEncoding(UTF_8);
+		response.setHeader(CACHE_CONTROL_HEADER, CACHE_CONTROL_NO_CACHE);
+		response.setHeader(CONNECTION_HEADER, CONNECTION_KEEP_ALIVE);
+		response.setHeader(SESSION_ID_HEADER, sessionId);
+
+		// Start async processing
+		AsyncContext asyncContext = request.startAsync();
+		asyncContext.setTimeout(0); // No timeout
+
+		// Check for Last-Event-ID header for resumable streams
+		String lastEventId = request.getHeader(LAST_EVENT_ID_HEADER);
+
+		// Create or get SSE stream for this session
+		StreamableHttpSseStream sseStream = getOrCreateSseStream(sessionId);
+		if (lastEventId != null) {
+			sseStream.replayEventsAfter(lastEventId);
+		}
+
+		PrintWriter writer = response.getWriter();
+
+		// Subscribe to the SSE stream and write events to the response
+		sseStream.getEventFlux().doOnNext(event -> {
+			try {
+				if (event.id() != null) {
+					writer.write("id: " + event.id() + "\n");
+				}
+				if (event.event() != null) {
+					writer.write("event: " + event.event() + "\n");
+				}
+				writer.write("data: " + event.data() + "\n\n");
+				writer.flush();
+
+				if (writer.checkError()) {
+					throw new IOException("Client disconnected");
+				}
+			}
+			catch (IOException e) {
+				logger.debug("Error writing to SSE stream: {}", e.getMessage());
+				asyncContext.complete();
+			}
+		}).doOnComplete(() -> {
+			try {
+				writer.close();
+			}
+			finally {
+				asyncContext.complete();
+			}
+		}).doOnError(e -> {
+			logger.error("Error in SSE stream: {}", e.getMessage());
+			asyncContext.complete();
+		}).subscribe();
+	}
+
+	/**
+	 * Handles HTTP POST requests for client messages.
+	 * @param request The HTTP servlet request
+	 * @param response The HTTP servlet response
+	 * @throws ServletException If a servlet-specific error occurs
+	 * @throws IOException If an I/O error occurs
+	 */
+	@Override
+	protected void doPost(HttpServletRequest request, HttpServletResponse response)
+			throws ServletException, IOException {
+
+		String requestURI = request.getRequestURI();
+		logger.debug("POST request received for URI: {}", requestURI);
+
+		// Log all headers for debugging
+		Enumeration<String> headerNames = request.getHeaderNames();
+		while (headerNames.hasMoreElements()) {
+			String headerName = headerNames.nextElement();
+			logger.debug("Header: {} = {}", headerName, request.getHeader(headerName));
+		}
+
+		if (!requestURI.endsWith(mcpEndpoint)) {
+			logger.debug("URI does not match mcpEndpoint: {}", mcpEndpoint);
+			response.sendError(HttpServletResponse.SC_NOT_FOUND);
+			return;
+		}
+
+		if (isClosing.get()) {
+			logger.debug("Server is shutting down, rejecting request");
+			response.sendError(HttpServletResponse.SC_SERVICE_UNAVAILABLE, "Server is shutting down");
+			return;
+		}
+
+		// According to spec, client MUST include an Accept header listing both
+		// application/json and text/event-stream
+		String acceptHeader = request.getHeader(ACCEPT_HEADER);
+		logger.debug("Accept header: {}", acceptHeader);
+		if (acceptHeader == null
+				|| (!acceptHeader.contains(APPLICATION_JSON) || !acceptHeader.contains(TEXT_EVENT_STREAM))) {
+			logger.debug("Accept header validation failed. Header: {}", acceptHeader);
+			response.setContentType(APPLICATION_JSON);
+			response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+			response.getWriter()
+				.write(createErrorJson("Accept header must include both application/json and text/event-stream"));
+			return;
+		}
+
+		// Client accepts SSE since we've validated the Accept header contains
+		// text/event-stream
+		boolean acceptsEventStream = true;
+
+		// Get session ID from header
+		String sessionId = request.getHeader(SESSION_ID_HEADER);
+		boolean isInitializeRequest = false;
+
+		try {
+			// Read request body
+			StringBuilder body = new StringBuilder();
+			try (BufferedReader reader = request.getReader()) {
+				String line;
+				while ((line = reader.readLine()) != null) {
+					body.append(line);
+				}
+			}
+
+			// Parse the JSON-RPC message
+			JSONRPCMessage message = McpSchema.deserializeJsonRpcMessage(objectMapper, body.toString());
+
+			// Check if this is an initialize request
+			if (message instanceof McpSchema.JSONRPCRequest req && McpSchema.METHOD_INITIALIZE.equals(req.method())) {
+				isInitializeRequest = true;
+				// For initialize requests, create a new session if one doesn't exist
+				if (sessionId == null) {
+					sessionId = sessionIdProvider.get();
+					logger.debug("Created new session ID for initialize request: {}", sessionId);
+				}
+			}
+
+			// Validate session ID for non-initialize requests
+			if (!isInitializeRequest && sessionId == null) {
+				response.setContentType(APPLICATION_JSON);
+				response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+				response.getWriter().write(createErrorJson("Session ID missing in request header"));
+				return;
+			}
+
+			// Get or create session
+			McpServerSession session = getOrCreateSession(sessionId, isInitializeRequest);
+			if (session == null && !isInitializeRequest) {
+				response.setContentType(APPLICATION_JSON);
+				response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+				response.getWriter().write(createErrorJson("Session not found: " + sessionId));
+				return;
+			}
+
+			// Handle the message
+			session.handle(message).block(); // Block for servlet compatibility
+
+			// Set session ID header in response
+			response.setHeader(SESSION_ID_HEADER, sessionId);
+
+			// For requests that expect responses, we need to set up an SSE stream
+			if (message instanceof McpSchema.JSONRPCRequest && acceptsEventStream) {
+				// Set up SSE connection
+				response.setContentType(TEXT_EVENT_STREAM);
+				response.setCharacterEncoding(UTF_8);
+				response.setHeader(CACHE_CONTROL_HEADER, CACHE_CONTROL_NO_CACHE);
+				response.setHeader(CONNECTION_HEADER, CONNECTION_KEEP_ALIVE);
+
+				// Start async processing
+				AsyncContext asyncContext = request.startAsync();
+				asyncContext.setTimeout(0); // No timeout
+
+				StreamableHttpSseStream sseStream = getOrCreateSseStream(sessionId);
+				PrintWriter writer = response.getWriter();
+
+				// For initialize requests, include the session ID in the response
+				if (isInitializeRequest) {
+					response.setHeader(SESSION_ID_HEADER, sessionId);
+				}
+
+				// Subscribe to the SSE stream and write events to the response
+				sseStream.getEventFlux().doOnNext(event -> {
+					try {
+						if (event.id() != null) {
+							writer.write("id: " + event.id() + "\n");
+						}
+						if (event.event() != null) {
+							writer.write("event: " + event.event() + "\n");
+						}
+						writer.write("data: " + event.data() + "\n\n");
+						writer.flush();
+
+						if (writer.checkError()) {
+							throw new IOException("Client disconnected");
+						}
+					}
+					catch (IOException e) {
+						logger.debug("Error writing to SSE stream: {}", e.getMessage());
+						asyncContext.complete();
+					}
+				}).doOnComplete(() -> {
+					try {
+						writer.close();
+					}
+					finally {
+						asyncContext.complete();
+					}
+				}).doOnError(e -> {
+					logger.error("Error in SSE stream: {}", e.getMessage());
+					asyncContext.complete();
+				}).subscribe();
+			}
+			else if (message instanceof McpSchema.JSONRPCRequest) {
+				// Client doesn't accept SSE, we'll return a regular JSON response
+				response.setContentType(APPLICATION_JSON);
+				response.setStatus(HttpServletResponse.SC_OK);
+				// The actual response would be sent later through another channel
+			}
+			else {
+				// For notifications and responses, return 202 Accepted
+				response.setStatus(HttpServletResponse.SC_ACCEPTED);
+			}
+		}
+		catch (Exception e) {
+			logger.error("Error processing message: {}", e.getMessage());
+			response.setContentType(APPLICATION_JSON);
+			response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+			response.getWriter().write(createErrorJson("Invalid JSON-RPC message: " + e.getMessage()));
+		}
+	}
+
+	/**
+	 * Handles HTTP DELETE requests to terminate sessions.
+	 * @param request The HTTP servlet request
+	 * @param response The HTTP servlet response
+	 * @throws ServletException If a servlet-specific error occurs
+	 * @throws IOException If an I/O error occurs
+	 */
+	@Override
+	protected void doDelete(HttpServletRequest request, HttpServletResponse response)
+			throws ServletException, IOException {
+
+		String requestURI = request.getRequestURI();
+		if (!requestURI.endsWith(mcpEndpoint)) {
+			response.sendError(HttpServletResponse.SC_NOT_FOUND);
+			return;
+		}
+
+		String sessionId = request.getHeader(SESSION_ID_HEADER);
+		if (sessionId == null) {
+			response.setContentType(APPLICATION_JSON);
+			response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+			response.getWriter().write(createErrorJson("Session ID missing in request header"));
+			return;
+		}
+
+		McpServerSession session = sessions.remove(sessionId);
+		if (session == null) {
+			response.setContentType(APPLICATION_JSON);
+			response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+			response.getWriter().write(createErrorJson("Session not found: " + sessionId));
+			return;
+		}
+
+		// Close the session and any associated SSE stream
+		StreamableHttpSseStream sseStream = sseStreams.remove(sessionId);
+		if (sseStream != null) {
+			sseStream.complete();
+		}
+
+		session.close();
+		logger.debug("Session terminated: {}", sessionId);
+
+		response.setStatus(HttpServletResponse.SC_OK);
+	}
+
+	/**
+	 * Gets or creates a session for the given session ID.
+	 * @param sessionId The session ID
+	 * @param createIfMissing Whether to create a new session if one doesn't exist
+	 * @return The session, or null if it doesn't exist and createIfMissing is false
+	 */
+	private McpServerSession getOrCreateSession(String sessionId, boolean createIfMissing) {
+		McpServerSession session = sessions.get(sessionId);
+		if (session == null && createIfMissing) {
+			StreamableHttpServerTransport transport = new StreamableHttpServerTransport(sessionId);
+			session = sessionFactory.create(transport);
+			sessions.put(sessionId, session);
+			logger.debug("Created new session: {}", sessionId);
+		}
+		return session;
+	}
+
+	/**
+	 * Gets or creates an SSE stream for the given session ID.
+	 * @param sessionId The session ID
+	 * @return The SSE stream
+	 */
+	private StreamableHttpSseStream getOrCreateSseStream(String sessionId) {
+		return sseStreams.computeIfAbsent(sessionId, id -> {
+			StreamableHttpSseStream stream = new StreamableHttpSseStream();
+			logger.debug("Created new SSE stream for session: {}", id);
+			return stream;
+		});
+	}
+
+	/**
+	 * Creates a JSON error response.
+	 * @param message The error message
+	 * @return The JSON error string
+	 */
+	private String createErrorJson(String message) {
+		try {
+			return objectMapper.writeValueAsString(new McpError(message));
+		}
+		catch (IOException e) {
+			logger.error("Failed to serialize error message", e);
+			return "{\"error\":\"" + message + "\"}";
+		}
+	}
+
+	/**
+	 * Implementation of McpServerTransport for Streamable HTTP sessions.
+	 */
+	private class StreamableHttpServerTransport implements McpServerTransport {
+
+		private final String sessionId;
+
+		/**
+		 * Creates a new session transport with the specified ID.
+		 * @param sessionId The unique identifier for this session
+		 */
+		StreamableHttpServerTransport(String sessionId) {
+			this.sessionId = sessionId;
+			logger.debug("Session transport {} initialized", sessionId);
+		}
+
+		@Override
+		public Mono<Void> sendMessage(JSONRPCMessage message) {
+			StreamableHttpSseStream sseStream = sseStreams.get(sessionId);
+			if (sseStream == null) {
+				logger.debug("No SSE stream available for session {}, message will be queued for next connection",
+						sessionId);
+				// Create a stream that will hold messages until a client connects
+				sseStream = getOrCreateSseStream(sessionId);
+			}
+
+			try {
+				String jsonText = objectMapper.writeValueAsString(message);
+				sseStream.sendEvent(MESSAGE_EVENT_TYPE, jsonText);
+				logger.debug("Message sent to session {}", sessionId);
+
+				// For responses to requests, we need to complete the stream to avoid
+				// hanging
+				if (message instanceof McpSchema.JSONRPCResponse) {
+					logger.debug("Completing SSE stream after sending response for session {}", sessionId);
+					sseStream.complete();
+				}
+
+				return Mono.empty();
+			}
+			catch (Exception e) {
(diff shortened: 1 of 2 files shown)


mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+		}
+
+	}
+
+	/**
+	 * Represents an SSE stream for a client connection.
+	 */
+	public class StreamableHttpSseStream {
+
+		private final Sinks.Many<SseEvent> eventSink = Sinks.many().multicast().onBackpressureBuffer();
```
ZachGerman: Meant to change this to unicast. Will address.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+	/**
+	 * Builder for creating instances of StreamableHttpServerTransportProvider.
+	 */
+	public static class Builder {
+
+		private ObjectMapper objectMapper = new ObjectMapper();
+
+		private String mcpEndpoint;
+
+		private Supplier<String> sessionIdProvider = () -> UUID.randomUUID().toString();
```
ZachGerman: Should change the value here to `DEFAULT_SESSION_ID_PROVIDER`. Will address.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+		if (sessions.isEmpty()) {
+			logger.debug("No active sessions to broadcast message to");
+			return Mono.empty();
+		}
+
+		logger.debug("Attempting to broadcast message to {} active sessions", sessions.size());
+
+		return Flux.fromIterable(sessions.values())
+			.flatMap(session -> session.sendNotification(method, params)
+				.doOnError(
```
chemicL: I think we should provide a handler so users can provide a callback here.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+						e -> logger.error("Failed to send message to session {}: {}", session.getId(), e.getMessage()))
+				.onErrorComplete())
+			.then();
+	}
+
+	@Override
+	public Mono<Void> closeGracefully() {
+		isClosing.set(true);
```
chemicL: This needs to be lazy, consider using `Mono.defer()`.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+		logger.debug("Initiating graceful shutdown with {} active sessions", sessions.size());
+
+		return Flux.fromIterable(sessions.values()).flatMap(McpServerSession::closeGracefully).then();
```
chemicL: Errors should be handled and onErrorComplete should be added for each inner chain

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+
+	/**
+	 * Handles HTTP GET requests to establish SSE connections.
+	 * @param request The HTTP servlet request
+	 * @param response The HTTP servlet response
+	 * @throws ServletException If a servlet-specific error occurs
+	 * @throws IOException If an I/O error occurs
+	 */
+	@Override
+	protected void doGet(HttpServletRequest request, HttpServletResponse response)
```
chemicL: There needs to be a configuration that allows session-less servers.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+			throws ServletException, IOException {
+
+		String requestURI = request.getRequestURI();
+		logger.debug("GET request received for URI: {}", requestURI);
+
+		// Log all headers for debugging
+		Enumeration<String> headerNames = request.getHeaderNames();
+		while (headerNames.hasMoreElements()) {
+			String headerName = headerNames.nextElement();
+			logger.debug("Header: {} = {}", headerName, request.getHeader(headerName));
```
chemicL: This log line could be aggregated into one with the debug log for receiving a request so only one line would be appended to the log.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+		}
+
+		if (isClosing.get()) {
+			logger.debug("Server is shutting down, rejecting request");
+			response.sendError(HttpServletResponse.SC_SERVICE_UNAVAILABLE, "Server is shutting down");
+			return;
+		}
+
+		String acceptHeader = request.getHeader(ACCEPT_HEADER);
+		logger.debug("Accept header: {}", acceptHeader);
```
chemicL: This debug log is unnecessary, the headers are logged above.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+			response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+			response.getWriter().write(createErrorJson("Accept header must include text/event-stream"));
+			return;
+		}
+
+		String sessionId = request.getHeader(SESSION_ID_HEADER);
+		if (sessionId == null) {
+			response.setContentType(APPLICATION_JSON);
+			response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+			response.getWriter().write(createErrorJson("Session ID missing in request header"));
```
chemicL: IMO this is a violation of the spec. The client should be free to issue the GET request as the first thing and the server should generate a session ID and proceed. The initialization phase should follow.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+		if (lastEventId != null) {
+			sseStream.replayEventsAfter(lastEventId);
+		}
+
+		PrintWriter writer = response.getWriter();
+
+		// Subscribe to the SSE stream and write events to the response
+		sseStream.getEventFlux().doOnNext(event -> {
+			try {
+				if (event.id() != null) {
```
chemicL: Are events without ID allowed?

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+			try {
+				writer.close();
+			}
+			finally {
+				asyncContext.complete();
+			}
+		}).doOnError(e -> {
+			logger.error("Error in SSE stream: {}", e.getMessage());
+			asyncContext.complete();
+		}).subscribe();
```
chemicL: It would be great to add a Reactor Context entry before subscribe containing the session ID. We can decide a common key, e.g. `MCP_SESSION_ID`.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+	@Override
+	protected void doPost(HttpServletRequest request, HttpServletResponse response)
+			throws ServletException, IOException {
+
+		String requestURI = request.getRequestURI();
+		logger.debug("POST request received for URI: {}", requestURI);
+
+		// Log all headers for debugging
+		Enumeration<String> headerNames = request.getHeaderNames();
+		while (headerNames.hasMoreElements()) {
```
chemicL: Similar comment on aggregation. Perhaps we can have a reusable method for this to avoid duplication.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+		if (isClosing.get()) {
+			logger.debug("Server is shutting down, rejecting request");
+			response.sendError(HttpServletResponse.SC_SERVICE_UNAVAILABLE, "Server is shutting down");
+			return;
+		}
+
+		// According to spec, client MUST include an Accept header listing both
+		// application/json and text/event-stream
+		String acceptHeader = request.getHeader(ACCEPT_HEADER);
+		logger.debug("Accept header: {}", acceptHeader);
```
chemicL: As in the case of GET, the headers are printed before.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+				|| (!acceptHeader.contains(APPLICATION_JSON) || !acceptHeader.contains(TEXT_EVENT_STREAM))) {
+			logger.debug("Accept header validation failed. Header: {}", acceptHeader);
+			response.setContentType(APPLICATION_JSON);
+			response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+			response.getWriter()
+				.write(createErrorJson("Accept header must include both application/json and text/event-stream"));
+			return;
+		}
+
+		// Client accepts SSE since we've validated the Accept header contains
```
chemicL: It should be up to the implementation of the particular server whether it's SSE or not. This decision can be based around the server configuration (session-less or supporting sessions) and potentially the type of the configured handler (Mono vs Flux).

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+		boolean acceptsEventStream = true;
+
+		// Get session ID from header
+		String sessionId = request.getHeader(SESSION_ID_HEADER);
+		boolean isInitializeRequest = false;
+
+		try {
+			// Read request body
+			StringBuilder body = new StringBuilder();
+			try (BufferedReader reader = request.getReader()) {
```
chemicL: Are we able to support non-blocking mode as well?

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+				isInitializeRequest = true;
+				// For initialize requests, create a new session if one doesn't exist
+				if (sessionId == null) {
+					sessionId = sessionIdProvider.get();
+					logger.debug("Created new session ID for initialize request: {}", sessionId);
+				}
+			}
+
+			// Validate session ID for non-initialize requests
+			if (!isInitializeRequest && sessionId == null) {
```
chemicL: We need to have a state machine on a per-session basis. [The spec says](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#listening-for-messages-from-the-server):
```
The client MAY issue an HTTP GET to the MCP endpoint.
This can be used to open an SSE stream, allowing the server to communicate to the client,
without the client first sending data via HTTP POST.
```

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+			McpServerSession session = getOrCreateSession(sessionId, isInitializeRequest);
+			if (session == null && !isInitializeRequest) {
+				response.setContentType(APPLICATION_JSON);
+				response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+				response.getWriter().write(createErrorJson("Session not found: " + sessionId));
+				return;
+			}
+
+			// Handle the message
+			session.handle(message).block(); // Block for servlet compatibility
```
chemicL: Can we handle async mode?

chemicL: Shouldn't the headers be written first? If we call into the handler first, it should attempt to respond, which might trigger flushing the response before headers are set.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+
+			// Set session ID header in response
+			response.setHeader(SESSION_ID_HEADER, sessionId);
+
+			// For requests that expect responses, we need to set up an SSE stream
+			if (message instanceof McpSchema.JSONRPCRequest && acceptsEventStream) {
+				// Set up SSE connection
```
chemicL: This logic should be encapsulated into how the session decides to go about it. If it decides it's a simple response, no SSE should be started.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+				response.setContentType(TEXT_EVENT_STREAM);
+				response.setCharacterEncoding(UTF_8);
+				response.setHeader(CACHE_CONTROL_HEADER, CACHE_CONTROL_NO_CACHE);
+				response.setHeader(CONNECTION_HEADER, CONNECTION_KEEP_ALIVE);
+
+				// Start async processing
+				AsyncContext asyncContext = request.startAsync();
+				asyncContext.setTimeout(0); // No timeout
+
+				StreamableHttpSseStream sseStream = getOrCreateSseStream(sessionId);
```
chemicL: Here [the spec](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server) is being slightly violated IMO. We should aim to respect the MAY keyword to the best of our ability. The responses should go on a dedicated SSE stream if a stream is to be used, not the one opened initially with GET. The free hanging GET stream is meant for notifications or requests from the server. The SSE events that deal with the originating request should be sent over the stream associated with this request.

Here's some explanation from the specification on POST:

```
If the server initiates an SSE stream:
...
The server MAY send JSON-RPC requests and notifications before sending a JSON-RPC response.
These messages SHOULD relate to the originating client request.
These requests and notifications MAY be [batched](https://www.jsonrpc.org/specification#batch).
The server SHOULD NOT close the SSE stream before sending a JSON-RPC response
per each received JSON-RPC request, unless the [session](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#session-management) expires
```

and for GET:

```
If the server initiates an SSE stream:
The server MAY send JSON-RPC requests and notifications on the stream.
These requests and notifications MAY be [batched](https://www.jsonrpc.org/specification#batch).
These messages SHOULD be unrelated to any concurrently-running JSON-RPC request from the client.
```

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+		AsyncContext asyncContext = request.startAsync();
+		asyncContext.setTimeout(0); // No timeout
+
+		// Check for Last-Event-ID header for resumable streams
+		String lastEventId = request.getHeader(LAST_EVENT_ID_HEADER);
+
+		// Create or get SSE stream for this session
+		StreamableHttpSseStream sseStream = getOrCreateSseStream(sessionId);
+		if (lastEventId != null) {
+			sseStream.replayEventsAfter(lastEventId);
```
chemicL: As I understand it, in case of resumption, once the final response is streamed, the SSE stream should be closed.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+					finally {
+						asyncContext.complete();
+					}
+				}).doOnError(e -> {
+					logger.error("Error in SSE stream: {}", e.getMessage());
+					asyncContext.complete();
+				}).subscribe();
+			}
+			else if (message instanceof McpSchema.JSONRPCRequest) {
+				// Client doesn't accept SSE, we'll return a regular JSON response
```
chemicL: This should not be possible. It's up to the server to decide, the clients are obliged to support SSE.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
chemicL: Looks like we can use Servlet 3.1's non-blocking support. I understand this implementation is using Jakarta Servlet 6.1, so we can use `request.getInputStream().setReadListener()`

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+
+		// Start async processing
+		AsyncContext asyncContext = request.startAsync();
+		asyncContext.setTimeout(0); // No timeout
+
+		// Check for Last-Event-ID header for resumable streams
+		String lastEventId = request.getHeader(LAST_EVENT_ID_HEADER);
+
+		// Create or get SSE stream for this session
+		StreamableHttpSseStream sseStream = getOrCreateSseStream(sessionId);
```
chemicL: Resumption needs to happen on a brand new SSE stream, otherwise the client is unable to distinguish between different streams.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,765 @@
...
+		catch (IOException e) {
+			logger.error("Failed to serialize error message", e);
+			return "{\"error\":\"" + message + "\"}";
+		}
+	}
+
+	/**
+	 * Implementation of McpServerTransport for Streamable HTTP sessions.
+	 */
+	private class StreamableHttpServerTransport implements McpServerTransport {
```
chemicL: I am guessing for Streamable HTTP we need a new abstraction, a `McpStreamableServerTransport` that has a different set of methods allowing opening a new stream and staying with the `McpServerTransport` in case of simple (sessionless) servers.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
chemicL: See my other comment for `StreamableHttpServerTransport`. We need to distinguish on a per message basis whether we are dealing with a new stream (upgrade to SSE) or a simple response and immediate flush. I imagine this could be handled with a new type of request handler (a new `McpServerSession` type or a new method) which returns a complex type that signifies whether the response is done (Mono) or will stream (Flux) and based on that we can decide what to do. The concept of a transport session from the client implementation should be introduced here as well.

The hierarchy I imagine is as follows:

```
StreamableHttpServerTransportProvider (1)
  StreamableHttpSession (1 per client initiated either with GET or POST)
    StreamableHttpSseStream (N - as many as the client initiates according to the server's upgrade policy and bound via session ID)
      StreamableHttpServerTransport (1 per stream) - this should route generic notifications and requests to one of the open SSE streams, preferably the one opened with GET); notifications, requests and responses tied to a POST should be handled differently (consumed from a Flux), via the containing StreamableHttpSseStream
```

The names and details are negotiable of course, I just wanted to bring some order to my thoughts.

ZachGerman:
Thank you very much for all of the input @chemicL! I will begin making changes accordingly this afternoon.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: I disagree on this for servers utilizing sessions, but it should not be required for session-less servers, for obvious reasons.
>The initialization phase MUST be the first interaction between client and server.
>...
>The client MUST initiate this phase by sending an initialize request [...]
>...
>The client SHOULD NOT send requests other than pings before the server has responded to the initialize request.

[[ref](https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle)]

>An MCP “session” consists of logically related interactions between a client and a server, beginning with the initialization phase

[[ref](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports)]

The following excerpt also caused a bit of confusion in a team discussion we had:
>The client MAY issue an HTTP GET to the MCP endpoint. This can be used to open an SSE stream, allowing the server to communicate to the client, without the client first sending data via HTTP POST.

...which I think should read "...without the client first sending data via HTTP POST (other than initialization)" as I don't believe the initialization request was considered "sending data" (even though it technically is), otherwise the spec is contradicting itself.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: Yes, `null` event IDs are used to control receiver behavior when reconnecting ("no [`Last-Event-ID`](https://html.spec.whatwg.org/multipage/server-sent-events.html#last-event-id) header [would] be sent in the event of a reconnection being attempted"[[ref](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation)]).

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: Forgot I added an `extractHeaders` method (returns header map) for debugging prior, so just using that and logging the request URI and header map in one line as suggested (for both POST and GET) 👍.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: TIL about Reactor Context. 😅 Thanks!

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: Implemented the class hierarchy, but current SSE upgrade logic results in non-streamed responses for everything except non-initialize request messages.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: We can probably just add a single stream session for GET requests into the StreamableHttpSession class. I can add that to my list today.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: Now writing headers as soon as they're known.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: Same comment:
We can probably just add a single stream session for GET requests into the StreamableHttpSession class. I can add that to my list today.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: Either client or server can close it at any time. Should we add a timer of sorts to hold the stream open and have it default to end right after stream completes?

ZachGerman:
Today I'm targeting origin header validation and moving the dedicated GET stream to the StreamableHttpSession class, then adding an integ test for GET on /mcp to start the listening stream.
After that, I believe we should have all core functionality except sessionless and proper SSE response upgrade logic.

sivankri:
Hi @ZachGerman I tried using your StreamableHttpServerTransportProvider.java file + java MCP SDK 0.10.0. My backend server is Jetty 12. The request from MCP Interceptor hangs in the below call.

 return streamSession.handle(message).then(Mono.just(responseType)).onErrorReturn(ResponseType.IMMEDIATE);

This is the actuall line which gets blocked
--McpServerSession.java --> return var10000.flatMap(var10001::sendMessage);

Can you please check this?

tzolov:
> Today I'm targeting origin header validation and moving the dedicated GET stream to the StreamableHttpSession class, then adding an integ test for GET on /mcp to start the listening stream. After that, I believe we should have all core functionality except sessionless and proper SSE response upgrade logic.

@ZachGerman  what do you mean by `origin header validation`?  I hope it is not overlapping with the https://github.com/modelcontextprotocol/java-sdk/pull/284

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
chemicL: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/282 please check this issue. I'm leaning towards separating the JSON-RPC lifecycle from the lower level transport lifecycle. In that world, GET and POST and session concepts are not at the JSON-RPC layer, so the lifecycle does not apply. In fact, the previous SSE transport did begin with a HTTP GET request before the initialization request could have been sent. For the stateless servers it should definitely be improved in the spec to mention that initialization is not required before other requests.

[EDIT]

For the record, the above was wishful thinking. I agree we are only able to generate the session ID upon POST of the initialize request. The spec is in fact melding the json-rpc layer with http, so the session ID gets generated only when processing a initialization request:

> A server using the Streamable HTTP transport MAY assign a session ID at initialization time, by including it in an `Mcp-Session-Id` header on the HTTP response containing the `InitializeResult`

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
chemicL: Right, but do we emit events with no ID ? Reading the code I don't see a situation like this. My understanding of the mentioned spec is that if the ID is missing then the client would not use this id for Last-Event-ID tracking, but we are on the server side and we always generate an ID.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
chemicL: My understanding is that we should close the stream on the server side once the `Flux` containing the json payloads completes. Not a timer though, just giving the user the ability to return a `Flux` that contains a set of notification/request/response payloads, where the response matching the initial request should be the last message from the Flux, followed by the `onComplete` signal. The other option would be for the user handling to return a `Mono` that signifies a non-SSE response directly to the POST message.

chemicL (commented):
Added a few more comments.

ZachGerman:
@sivankri: Changing the related logic after my meeting with Dariusz this morning, so (hopefully) your issue goes away with the new response type mapping.
@tzolov: It's related, but I don't think it's overlapping as I'm just adding the ability to set a list of allowed origins to the server transport provider and enforcing it if it's set.

viyaviya:
merge commit 6c4830bf5362ca71887736c357a3cd794fa86c1a  is bad

ZachGerman:
> merge commit [6c4830b](https://github.com/modelcontextprotocol/java-sdk/commit/6c4830bf5362ca71887736c357a3cd794fa86c1a) is bad

Oops! Thanks for the heads up! Fixed!

ZachGerman:
@sivankri the new logic added `streamTools` to McpServerFeatures, used in the Async constructor, which facilitates a list of tool specifications that return `Flux` instead of `Mono` via the new `AsyncStreamingToolSpecification` record type and uses the return type of the `call` to differentiate between direct-HTTP and SSE-stream responses.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: >Right, but do we emit events with no ID ?

No. Should I just skip the null check or log or throw an error if we come across one?

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: Should we wait until the issue in the spec is sorted out to add Simple/Stateless/Sessionless?

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: SSE upgrade logic is now based on the return type of the tool call.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
ZachGerman: Done. Dedicated GET/listening stream integrated.

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
```diff
@@ -215,8 +215,8 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response)
 		PrintWriter writer = response.getWriter();
 
 		// Create a new session transport
-		HttpServletMcpSessionTransport sessionTransport = new HttpServletMcpSessionTransport(sessionId, asyncContext,
-				writer);
+		HttpServletMcpSessionTransport sessionTransport = new HttpServletMcpSessionTransport(
+				McpServerSession.LISTENING_TRANSPORT, asyncContext, writer);
```
ZachGerman: This was unnecessary and being reverted.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncStreamableHttpServer.java
ZachGerman, line 1: Is this file worth keeping? (`McpAsyncServer.java` supports both with an `isStreamableHttp` bolean parameter)

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -170,13 +243,39 @@ public Mono<Void> handle(McpSchema.JSONRPCMessage message) {
...
+				else {
+					if (request.id() instanceof String) {
+						transportId = (String) request.id();
+					}
+					else if (request.id() instanceof Integer) {
+						transportId = request.id().toString();
+					}
+					else {
+						logger.error("Invalid request ID: {}", request.id());
+						return Mono.empty();
```
ZachGerman, line 260: Changed this to:
`return Mono.error(new RuntimeException("Invalid request ID: " + String.valueOf(request.id())));
`

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -170,13 +243,39 @@ public Mono<Void> handle(McpSchema.JSONRPCMessage message) {
...
+						// I think I'm missing some handling here. Please advise.
```
ZachGerman: Changed this return to:
`return Mono.error(new RuntimeException("Invalid request ID: " + String.valueOf(request.id())));`

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncStreamableHttpServer.java
```diff
@@ -0,0 +1,663 @@
...
+ * @author Christian Tzolov
+ * @author Dariusz Jędrzejczyk
+ * @author Jihoon Kim
+ * @author Zachary German
+ */
+public class McpAsyncStreamableHttpServer {
+
+	private static final Logger logger = LoggerFactory.getLogger(McpAsyncStreamableHttpServer.class);
+
+	private final StreamableHttpServerTransportProvider httpTransportProvider;
```
sivankri: We need reference to this transport provider. Can we have a getter method to get the reference?

        context.addServlet(new ServletHolder(asyncStreamableServer.getHttpTransportProvider()), "/*");

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -119,7 +123,7 @@ public class McpAsyncServer {
 	 */
 	McpAsyncServer(McpServerTransportProvider mcpTransportProvider, ObjectMapper objectMapper,
 			McpServerFeatures.Async features, Duration requestTimeout,
-			McpUriTemplateManagerFactory uriTemplateManagerFactory) {
+			McpUriTemplateManagerFactory uriTemplateManagerFactory, boolean isStreamableHttp) {
```
sivankri: This constructor is called only from overloaded methods present in this class. isStreamableHttp cannot be set to true currently. Please check.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncStreamableHttpServer.java
sivankri: We should try to leverage McpAsyncServer class with isStreamableHttp boolean property without creating a new class.

sivankri:
@ZachGerman I tried running my mcp remote server against the mcp sdk generated from your branch.  When i used the MCP inspector tool  by setting query parameter transport=streamable-http, i could not get the connection established. Can you please help me with this?

ZachGerman:
@sivankri I'm unfamiliar with the MCP inspector tool, but I imagine it's related to the fact that this JDK has `public static final String LATEST_PROTOCOL_VERSION = "2024-11-05";` in the McpSchema spec class & that is currently used for all server instantiation. You can update it locally and the server from my branch, as well as @chemicL's webClient, should work with the update, but it will break a lot of tests. I plan on tackling protocol version negotiation soon if nobody else does.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -203,8 +334,10 @@ private Mono<McpSchema.JSONRPCResponse> handleIncomingRequest(McpSchema.JSONRPCR
 			Mono<?> resultMono;
 			if (McpSchema.METHOD_INITIALIZE.equals(request.method())) {
 				// TODO handle situation where already initialized!
-				McpSchema.InitializeRequest initializeRequest = transport.unmarshalFrom(request.params(),
-						new TypeReference<McpSchema.InitializeRequest>() {
+				McpSchema.InitializeRequest initializeRequest = transports.isEmpty() ? listeningTransport
+					.unmarshalFrom(request.params(), new TypeReference<McpSchema.InitializeRequest>() {
+					}) : transports.get(request.id())
```
sivankri: Hi @ZachGerman I get NPE here. 
request.id() -> returns Integer
transports Key is String. 
So the conversion did not happen and it is throwing NPE.

I fixed it using 
transports.get(String.valueOf(request.id())).unmarshalFrom(request.params(), new TypeReference<McpSchema.InitializeRequest>() {});

Now it works nicely.

ZachGerman: Will push this change at end of day with the update to route client-request-driven server-notifications/requests to the request's response stream if available.

ZachGerman:
Once I realized that I needed to identify the transports in the exchange's session from the tool handlers, I realized I need to either pass the entire JSONRPCRequest to the handler (instead of the request.params()), or add a third request.id() parameter to tool handlers.

Either way, it brought more attention to the fact that we were using `Object` for JSONRPCRequest/Response `id`, so I added a little sub-class in the McpSchema to enforce the "(String | Number) && Non-null" requirements of that field.

Tomorrow I will be making the JSONRPCRequest `id` available to tool handlers and it can be passed to the `McpAsyncServerExchange` class methods in order to route notifications and requests to the related SSE stream.

-------------------------

Edit: Ended up going with a per-request exchange instance instead of passing transport ID to tool call handlers. This allows tool implementers to not worry about transport routing with their exchange method calls.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
ZachGerman: Didn't get the routing finished yesterday (wanted to enforce `id` rules) but I pushed this change with the `id` changes.
Thanks!

ZachGerman:
Removed streamTools in favor of upgrading transport instances to SSE based on McpAsyncServerExchange method usage. 
Also added a more extensive suite of integration tests.
Will speak to @chemicL about adding a config option at the server level to declare default transport type, as I believe he wants that to be SSE for session-bound requests.
Also planning to make event histories some kind of append-only ordered iterable struct, as using `Map.keySet().stream().sorted()` when replaying is a bit gross in hind-sight.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,893 @@
...
+	private final ObjectMapper objectMapper;
+
+	/** The endpoint path for handling MCP requests */
+	private final String mcpEndpoint;
+
+	/** Supplier for generating unique session IDs */
+	private final Supplier<String> sessionIdProvider;
+
+	/** Sessions map, keyed by Session ID */
+	private static final Map<String, McpServerSession> sessions = new ConcurrentHashMap<>();
```
etherandrius: This should not be static in order to allow having multiple independent MCP Server instances on the same Java Server. For example if we want to support having an MCP on paths `/mcp/v1/` and `/mcp/v2/`.

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
```diff
@@ -250,11 +247,13 @@ public Mono<Void> sendMessage(McpSchema.JSONRPCMessage message) {
...
 				.exchangeToFlux(response -> {
-					if (transportSession
-						.markInitialized(response.headers().asHttpHeaders().getFirst("mcp-session-id"))) {
-						// Once we have a session, we try to open an async stream for
-						// the server to send notifications and requests out-of-band.
-						reconnect(null).contextWrite(sink.contextView()).subscribe();
+					transportSession.markInitialized(response.headers().asHttpHeaders().getFirst("mcp-session-id"));
+					if (response.statusCode().is2xxSuccessful()
+							&& message instanceof JSONRPCNotification notification) {
+						if (notification.method().equals("notifications/initialized")) {
```
chemicL: It is not required to wait until the notification is sent out. If you see something in the spec, please let me know, but what I can see is this:

> A server using the Streamable HTTP transport MAY assign a session ID at initialization time, by including it in an `Mcp-Session-Id` header on the HTTP response containing the `InitializeResult`

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -141,6 +156,103 @@ public static final class ErrorCodes {
 
 	}
 
+	/**
+	 * MCP ID wrapper: MUST be non-null String or Number.
+	 */
+	@JsonSerialize(using = McpId.Serializer.class)
+	@JsonDeserialize(using = McpId.Deserializer.class)
+	public static final class McpId {
```
chemicL: As I understand this change is not strictly required for the Streamable HTTP Server implementation to work, so I'd prefer this to not be a part of this already big PR.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
```diff
@@ -75,6 +82,28 @@ public McpSchema.Implementation getClientInfo() {
...
+					new StreamableHttpServerTransportProvider.SseTransport(transport.getObjectMapper(),
+							transport.getResponse(), transport.getAsyncContext(), null, transportId, session.getId()));
+		}
+	}
+
+	/**
+	 * This is for tool writers to use if they want to send their tool response over an
+	 * SSE stream without using any other McpAsyncServerExchange methods
+	 */
+	public void upgradeTransport() {
```
chemicL: IMO this is not a correct way to go about it. I'd suggest that if in the future someone asks for dynamic upgrade we should do it differently by providing a hook to inspect the request first, instruct the consuming session and transport that it will not only be a response but a bit more, and only then feed an exchange that has the SSE stream producer to feed with user interactions, but not in a stateful way like this.

For now I suggest we keep it simple as other SDKs:
* notifications and responses -> handle as 201 ACCEPTED
* requests
  * sessionless -> application/json response expected
  * with sessions -> always upgrade to SSE

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
```diff
@@ -92,14 +121,17 @@ public McpSchema.Implementation getClientInfo() {
 	 * Specification</a>
 	 */
 	public Mono<McpSchema.CreateMessageResult> createMessage(McpSchema.CreateMessageRequest createMessageRequest) {
+
+		establishSseStream();
```
chemicL: Methods returning Flux/Mono need to be lazy, even if they have side-effects. Another issue with this is explained in the comment above.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -40,26 +70,34 @@ public class McpServerSession implements McpSession {
...
 
-	private static final int STATE_INITIALIZED = 2;
+	public static final int STATE_INITIALIZED = 2;
 
 	private final AtomicInteger state = new AtomicInteger(STATE_UNINITIALIZED);
 
+	private volatile McpSchema.LoggingLevel minLoggingLevel = McpSchema.LoggingLevel.INFO;
+
+	// TODO: Remove this if we split SHTTP Server Session into its own class.
+	private AtomicBoolean isStreamableHttp = new AtomicBoolean(false);
```
chemicL: I agree, we should not have a boolean flag to drive the logic internally.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -89,6 +152,90 @@ public String getId() {
...
+		transportEventHistories.put(transportId, eventHistory);
+	}
+
+	/**
+	 * Used for replayability support to retrieve the entire event history for a given
+	 * transport ID
+	 * @param transportId
+	 * @return Map of SseEvent objects, keyed by event ID
+	 */
+	public Map<String, SseEvent> getTransportEventHistory(String transportId) {
```
chemicL: The fact that this is a map makes it unclear whether it is ordered and that becomes and implementation detail. The ordering of history and the means to replay are the essence of the replay/resumability functionality. So we need a separate type for the event history in my view and offer an in-memory default implementation.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -334,6 +578,25 @@ public interface RequestHandler<T> {
 
 	}
 
+	/**
+	 * A handler for client-initiated requests return Flux.
+	 *
+	 * @param <T> the type of the response that is expected as a result of handling the
+	 * request.
+	 */
+	public interface StreamingRequestHandler<T> extends RequestHandler<T> {
```
chemicL: This is unused and should be removed I suppose?

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -350,4 +613,21 @@ public interface Factory {
 
 	}
 
+	/**
+	 * Factory for creating server sessions which delegate to a provided 1:1 transport
+	 * with a connected client.
+	 */
+	@FunctionalInterface
+	public interface StreamableHttpSessionFactory {
```
chemicL: Adding the boolean flag causes adding another factory and infects the transports with empty implementations. Unfortunately, we will need a parallel type hierarchy.

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
```diff
@@ -117,10 +118,6 @@ public static Builder builder(WebClient.Builder webClientBuilder) {
 	public Mono<Void> connect(Function<Mono<McpSchema.JSONRPCMessage>, Mono<McpSchema.JSONRPCMessage>> handler) {
 		return Mono.deferContextual(ctx -> {
 			this.handler.set(handler);
-			if (openConnectionOnStartup) {
```
chemicL: I agree this can be removed but we also have to remove it from the constructor.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,893 @@
...
+	 * @param mcpEndpoint String - Default: '/mcp'
+	 * @param sessionIdProvider Supplier(String) - Default: UUID.randomUUID().toString()
+	 */
+	public StreamableHttpServerTransportProvider() {
+		this(null, null, null);
+	}
+
+	@Override
+	public void setSessionFactory(McpServerSession.Factory sessionFactory) {
+		// Required but not used for this implementation
```
chemicL: We can't implement it this way. This transport needs to be of a different kind.

mcp/src/main/java/io/modelcontextprotocol/server/transport/StreamableHttpServerTransportProvider.java
```diff
@@ -0,0 +1,893 @@
...
+					else if (message instanceof JSONRPCResponse resp) {
+						transportId = resp.id().toString();
+					}
+					else {
+						transportId = null;
+					}
+
+					// Only set content type for requests
+					if (message instanceof McpSchema.JSONRPCRequest) {
+						response.setContentType(APPLICATION_JSON);
```
chemicL: So here you set it to application/json but later you let the user change this to text/event-stream. Unfortunately with other transports it might not be as straightforward and the builders are more type-safe and will forbid rolling back in such a way. That's why I am proposing to keep it simple for now as in another comment and always upgrade to SSE for requests.

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
ZachGerman: After we split the server session class, should we switch them all to mark the session as initialized once they send the `InitializeResult` response? (They currently only mark them as initialized upon receiving the `initialized` notification) Without that change or making the client wait to get the 200 from its `initialized` notification, we will have clients trying to establish streams prior to initialization being complete.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
ZachGerman: +1! From previous comment above:
>Also planning to make event histories some kind of append-only ordered iterable struct, as using Map.keySet().stream().sorted() when replaying is a bit gross in hind-sight.

Planning to go with `ConcurrentSkipListMap<String, SseEvent>` unless you have a better recommendation (maybe `NavigableMap<Long, SseEvent>`???).

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
ZachGerman: Yes. Forgot to remove this when I removed streamTools.

chemicL (approved):

qcloop:
Thanks for great work. Is there a coordination with Spring AI MCP? Would love to take it for a test drive

Issue #72, opened 2025-03-26
When is the "Streamable HTTP" transport scheduled to be supported?


taobaorun:
**Expected Behavior**

The ["Streamable HTTP" transport ](https://github.com/modelcontextprotocol/specification/pull/206)is a fantastic feature, particularly beneficial for stateless MCP servers and remote MCP servers in distributed scenarios. It enables MCP to be swiftly employed in production-grade agent applications

**Current Behavior**

The issue with the current session is that it needs to support distributed sessions.

hongling-dev:
When is  [2025-03-26 Implementation](https://github.com/orgs/modelcontextprotocol/projects/3) 's release date?  thx @chemicL

dyrnq:
stare

Pratiush:
I would love to contribute on this feature.

chemicL:
I started diving into the subject and began prototyping it. If there arise some areas where contributions will be useful, I will update this issue with links to individual tasks. Thanks for your patience.

chemicL:
Folks, feel free to test locally the work in progress in #292 and please report any issues. We will follow up with the JDK HttpClient implementation and then take care of the server side.

ZachGerman:
@chemicL Should I finish the server-side transport using [this](https://github.com/modelcontextprotocol/java-sdk/pull/290) or work on something else?

chemicL:
Hey, @ZachGerman. I must admit I haven't been actively reviewing nor following the issues/PRs recently. I see you did some work there which looks promising and thank you for the effort. If you'd like, please consider an integration test including the WIP client introduced here. In the future, please keep in mind our [contributing guidelines](https://github.com/modelcontextprotocol/java-sdk/blob/main/CONTRIBUTING.md) before implementing a major feature.

JavaProgrammerLB:
follow this

sunqb:
We really need this feature. When will it be available?  @chemicL

chemicL:
@taobaorun I edited the description a bit, hope you don't mind.

@sunqb please note the updated description, there are a few bits missing. The first client will become available once we merge the first PR and make a release. I'm not sure if we'd like to release the client first or together with the server, so no promises at this point, but we are focused on this effort right now.

ZachGerman:
[...]

Issue #275, opened 2025-05-28
Non initialized McpServerSession hangs and timesout instead of providing helpful error message


marianogonzalez:
This issue is closely related to #274 and could be considered as a follow up to that, but I decided to split it for convenience and readability.

Calling `McpServerSession#handle(JSonRPCMessage)` on an instance that hasn't yet received the `notifications/initialized` notification results in a Mono that never completes until a Timeout is raised.

This happens because the `io.modelcontextprotocol.spec.McpServerSession#exchangeSink` will never emit any value until such notification is received, and thus the `this.exchangeSink.asMono().flatMap(exchange -> handler.handle(exchange, request.params()));` in `McpServerSession#225` will never complete. 

Putting the considerations of #274 aside, it is fair for a non intiailized session to not handle messages, but it should do it by quickly returning a `Mono.error` instead of hanging and returning an unspecified timeout.

chemicL:
Thank you for the detailed analysis. I wonder if we can actually address this with the current spec requiring the initialized notification to arrive. The asynchronous nature of the exchanges forces us to actually wait until that notification happens for any other interaction by the client within a session. If we failed immediately, we'd potentially fail valid requests that happen from the client's perspective _after_ the initialization has been negotiated, but which are processed in different order by the server threads. Therefore, the timeout is the best we can do right now. Since the initialization phase is required for the server to properly identify the client capabilities, we can only rely on the timeout to fail a particular exchange. I wonder whether there are other ideas, but we could relax the requirement about waiting for the initialized notification provided that we have the session storage implemented.

arvindand:
@chemicL That makes sense for requests received while initialization is in progress but since the session already distinguishes UNINITIALIZED, INITIALIZING, and INITIALIZED, could we reject application requests received in UNINITIALIZED with INVALID_REQUEST or another appropriate protocol error while preserving the existing wait in INITIALIZING? This would retain the asynchronous ordering behavior while avoiding an indefinite wait when initialization has not started.

..for context, I ran into this while testing my stdio MCP server using a minimal Python client that sent tools/list before initialize. The server remained running but returned no response, whereas the correctly initialized client returned the tool list immediately. Although the client call sequence was incorrect, an INVALID_REQUEST response would have made it easier to diagnose.

chemicL:
[...]

Issue #274, opened 2025-05-28
McpServerSession lifecycle doesn't support distributed services


marianogonzalez:
This issue is closely related to #273 and could be considered as a follow up to that, but I decided to split it for convenience and readability.

Consider an MCP server written with this SDK and using the StreamableHttp transport. We deploy two replicas (`A` and `B`) of this server, with a load balancer in front.

1. A new client application initiates a session by sending an `initialize` message, followed by a `notifications/initialized` notifications. 
2. Let's say that the Load Balancer routed both messages to replica `A`
3. As a result, replica `A` generated a new session which Id is communicated back to the client
4.  Replica `A` now has an `McpServerSession` instance properly initialized with all the necessary client information
 
Problem:
 
1. Next the client sends a `tools/list` message with the obtain sessionId, but this time the load balancer routes the message to replica `B`
2. Even though I believe it should be the server's responsibility to share the session id across replicas, the `McpServerSession` instance that lives in replica `B` has never received (and will never receive) the initialize message and notifications
3. Although it's possible to distribute all that state and `simulate` that handshake in replica `B` (I actually did it), I think this SDK should have an OOTB support for that, or at the very least, SessionFactory methods that make it easier to generate "ready to use" Sessions based on previous initializations.

alexanderqiu:
add one

iaculch:
spring-session support would be nice here, we currently use the JDBC spring session store type and I would expect SSE sessions to ideally use the same / similar mechanism

gianielsevier:
The current McpAsyncServer constructor unconditionally sets a default sessionFactory on the provided McpTransportProvider, even if a custom factory has already been injected. This makes it impossible to plug in a distributed session store such as Hazelcast, Redis, etc... for cross-pod session sharing in a multi-instance deployment.

In my case, I registered a Hazelcast-backed sessionFactory on the McpTransportProvider prior to constructing the server, but it was silently overwritten. This behavior breaks the ability to externalize session state, which is critical for stateless MCP server deployments behind a load balancer.

Suggested improvement: Before setting the default factory, the constructor could check whether McpTransportProvider already has a sessionFactory defined. If so, it should respect the existing one or allow injection via the builder. This would enable advanced use cases like distributed caching, session affinity, and custom lifecycle management without requiring deep overrides or manual patching.

sargue:
[...]

Issue #273, opened 2025-05-28
Session+Transport tied lifecycle doesn't easily support stateless connections such as StreamableHttp


marianogonzalez:
Hello,

I'm trying to build an MCP server that uses the StremeableHttp transport (for which I created my own implementation). Although I finally got it to work, I found what I believe to be a design flaw in the way that McpAsyncServer -> McpServerSession -> Transport lifecycle are coupled and tied together. Let me explain:

Right now, the Session and Transport pretty much share the same lifecycle. They are created, used and stopped together. This works well for stdio and SSE transports, since those are stateful connections and the act of `reconnecting` results in brand new Session and Transport instances being created. This design works for SSE and will work for any other type of stateful connections (e.g: an eventual WebSockets transport).

This however doesn't work for stateless connection types such as streamable http, in which the session is created when the first initialize request is received but has a lifecycle that survives across many distinct http requests (even concurrent ones). 

IMO, the root cause of the problem is that the McpServerTransport is missing the concept of `TransportRequest`  or `TransportMessage`. Furthermore, the fact that the `McpServerTransportProvider` creates new `Transport` instances  per each new request or session, suggests to me that the Transport is kind of partially assuming responsibilities that would better fit a request object. 

Using StremeableHttp as an example, the advantages of introducing this `Request` object would be the following:

* `TransportProvider` dissapears and becomes simply `McpServerTransport`
* Whenever a JsonRPCMessage is received, the `McpServerTransport` instantiates a new `TransportRequest` object that will contain all the relevant information. In the case of StremeableHttp, it would contain the HttpRequest object, some kind of callback or Writer to generate the response, client connection information, etc. NOTE: This would also make it quite easier to get that information than it is today, making things like security and tracing easier to implement
* When needed, the `McpServerTransport` will use the `McpServerSessionFactory` to create new sessions. The session is then capable of handling further requests. 
* At this point, the lifecycle of all three components have been completely decoupled. Processing concurrent messages over stateless connections is now much much simpler than today

marianogonzalez:
This issue is slightly related to issue [274](https://github.com/modelcontextprotocol/java-sdk/issues/274) , but decided to split them for simplicity and readability

xyq043170:
could you please share the timeline to have streamable http for mcp java sdk ?

chemicL:
[...]

PR #292, opened 2025-06-04
WebClient Streamable HTTP support

chemicL:
Introducing WebClient support for Streamable HTTP transport.

## Motivation and Context
The 2025-03-26 specification of MCP introduces Streamable HTTP transport that replaces SSE transport.
This change introduces more resilient primitives that support the initial, WebClient-based client transport for Streamable HTTP.

Various failure modes at either the transport or logical layer have been addressed and are now fully supported by the new `WebClientStreamableHttpTransport` class. The surrounding `McpAsyncClient` and `McpClientSession` have been modified to handle possible issues according to the adequate layer's responsibility.

### Orientation

The existing APIs are organized quite logically with regards to various responsibilities using layers. Below is the status quo description and the changes required for Streamable HTTP to be implemented.

#### Rationale for the API

The client types are layered in the following way:

```
McpAsyncClient
  McpClientSession
    McpClientTransport
      McpTransportSession
      McpTransportStream
```

Here are the distinguishing features of these APIs:

`McpAsyncClient` (and the facade, `McpSyncClient`)
  - user facing class exposing capabilities like tools, etc.
  - maintains state for the initialization (and an inner class was added to encapsulate this state)
  - maintains state for capabilities handlers, these translate to lower-level handlers at JSON-RPC level in the session
  - (change) used to have a single session, now dynamically allocates sessions upon reconnects/invalidations
  - (change) used to delegate to session to manage the transport, now it owns the transport and its lifecycle

`Mcp(Client)Session`
  - internal implementation providing JSON-RPC mapping of requests with responses
  - maintains state for the requests pending responses
  - allows defining req/resp/notification handlers and keeps a state of them
  - dispatches messages to their handlers
  - (change) used to own the transport and manage its lifecycle, now it's independent

`Mcp(Client)Transport`
  - specific implementation for the transport layer of serialized JSON-RPC
  - STDIO, SSE, Streamable HTTP
  - maintains open streams in case of streamable connections (SSE, Streamable HTTP, STDIO)
  - maintains particular state of the transport type, e.g. message endpoint in case of SSE transport
  - (new) spawns transport-specific sessions (e.g. HTTP status-driven session management from streamable HTTP)
  - (new) spawns transport-specific streams that can come and go

`McpTransportSession`
 - new abstraction over transport-driven session concept
 - allows enriching the communication with transport-level session identifier to communicate with stateful servers

`McpTransportStream`
 - new abstraction for a transport-level stream (particularly SSE stream) with message id tracking for resumability purposes

## How Has This Been Tested?
* Regular integration tests using the `everything` MCP server.
* Resilience tests validating reconnections and session invalidation handling.

## Breaking Changes
The client initialization became lazy and more error-prone - connections happen upon first request and initialization is performed implicitly if it has not been issued before.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
The JDK HttpClient will be implemented next.

## Remaining tasks
- [x] Builder
- [x] Javadoc
- [x] Supporting DELETE call

Related #72


chemicL:
The CI fails for the Stdio tests. I can't reproduce it on my M1 Mac, but will try a linux box and investigate.

chemicL:
There is the missing bit for issuing a DELETE upon client-side session termination, I'll investigate how it could be achieved.

I will review the spec once more and see what else is missing.

bugzyz:
Hi @chemicL thanks for the excellent work! I can run the unit test on our mcp streamable http service.

But one issue is that our service is running on springframework 5.3.28. And I notice that your WebClientStreamableHttpTransport only work compatible with springframework 6.2.1.

So when I integrate with your code in our service. The below error popped up.
```
java.lang.NoClassDefFoundError: org/springframework/http/HttpStatusCode
```
this is from spring-web-6.2.1.jar

We integrated with HttpClientSseClientTransport before to take care of sse, and the springframework 5.3.28 works fine with the class. Now we would like to upgrade to WebClientStreamableHttpTransport and has this incompatible issue. Is it possible to make it compatible with the old spring version? Thanks!

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -355,28 +380,93 @@ public Mono<McpSchema.InitializeResult> initialize() {
...
+
+		static Initialization create() {
+			return new Initialization();
+		}
+
+		void setMcpClientSession(McpClientSession mcpClientSession) {
+			this.mcpClientSession.set(mcpClientSession);
+		}
+
+		McpClientSession session() {
```
tzolov: maybe rename to `mcpSession()` to avoid confusion b/w mcp and transport sessions.

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -326,26 +353,24 @@ public Mono<Void> closeGracefully() {
 	 * @see <a href=
 	 * "https://github.com/modelcontextprotocol/specification/blob/main/docs/specification/basic/lifecycle.md#initialization">MCP
 	 * Initialization Spec</a>
+	 * </p>
 	 */
 	public Mono<McpSchema.InitializeResult> initialize() {
+		return withSession("by explicit API call", init -> Mono.just(init.get()));
+	}
 
+	private Mono<McpSchema.InitializeResult> doInitialize(McpClientSession session) {
```
tzolov: perhaps rename `session` -> `mcpSession`

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -234,16 +232,31 @@ public class McpAsyncClient {
...
-		this.mcpSession = new McpClientSession(requestTimeout, transport, requestHandlers, notificationHandlers);
+		this.transport.setExceptionHandler(this::handleException);
+		this.sessionSupplier = () -> new McpClientSession(requestTimeout, transport, requestHandlers,
+				notificationHandlers);
 
 	}
 
+	private void handleException(Throwable t) {
+		logger.warn("Handling exception", t);
+		if (t instanceof McpSessionNotFoundException) {
```
tzolov: If the McpSessionNotFoundException is related to the Transport session a TransportSessionNotFoundException or McpTransportSessionNotFoundException would help to distinct b/w both session abstractions

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -234,16 +232,31 @@ public class McpAsyncClient {
...
+		}
+	}
+
 	/**
 	 * Get the server capabilities that define the supported features and functionality.
 	 * @return The server capabilities
 	 */
 	public McpSchema.ServerCapabilities getServerCapabilities() {
-		return this.serverCapabilities;
+		Initialization current = this.initialization.get();
```
tzolov: This is repeating and can be abstracted:
```
Initialization current = this.initialization.get();
McpSchema.InitializeResult initializeResult = current != null ? current.result.get() : null;
```

chemicL:
@bugzyz thank you! We are targetting Spring Framework gen 6, unfortunately. However, in your circumstances, once we introduce the JDK HttpClient support for Streamable HTTP, you should be able to use that in combination with older Spring Framework-based code. That work is coming once we finalize the architectural changes covered here. And then you wouldn't need the mcp-spring module dependency. Hope this helps.

chemicL:
I updated the description in the PR to reflect the architectural overview and decisions to improve the review. @tzolov I'll continue looking into the DELETE aspect and come back either here or via a follow-up PR.

noear:
Streamable is finally here:) Thanks!

noear:
@bugzyz java8, java11, spring5, etc., can use the solon-ai project

https://github.com/opensolon/solon-ai

---

We sync a java8 code and provide a simple experience.

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -234,16 +232,31 @@ public class McpAsyncClient {
...
 	}
 
+	private void handleException(Throwable t) {
+		logger.warn("Handling exception", t);
+		if (t instanceof McpSessionNotFoundException) {
+			Initialization previous = this.initialization.getAndSet(null);
+			if (previous != null) {
+				previous.close();
+			}
+			withSession("re-initializing", result -> Mono.empty()).subscribe();
```
tzolov, line 271: Not sure what the `result -> Mono.empty()` means?

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -80,24 +80,31 @@ public class McpAsyncClient {
...
+	};
+
+	public static final TypeReference<CreateMessageRequest> CREATE_MESSAGE_REQUEST_TYPE_REF = new TypeReference<>() {
+	};
+
+	public static final TypeReference<LoggingMessageNotification> LOGGING_MESSAGE_NOTIFICATION_TYPE_REF = new TypeReference<>() {
+	};
 
-	private AtomicBoolean initialized = new AtomicBoolean(false);
+	private final AtomicReference<Initialization> initialization = new AtomicReference<>();
```
tzolov: maybe we should be more explicit in naming like `initializationRef` ?

chemicL: agreed, thx

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
chemicL: Absolutely agreed, thx!

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
chemicL: I'll add a comment to clarify. In a nutshell, the argument is the operation to perform once initialization takes place - here we don't do anything, so we provide a `Mono.empty()`.

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
chemicL: Thanks for catching this.

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
chemicL: went with mcpClientSession to be more precise

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
chemicL: Yep, thx

bugzyz:
> @bugzyz thank you! We are targetting Spring Framework gen 6, unfortunately. However, in your circumstances, once we introduce the JDK HttpClient support for Streamable HTTP, you should be able to use that in combination with older Spring Framework-based code. That work is coming once we finalize the architectural changes covered here. And then you wouldn't need the mcp-spring module dependency. Hope this helps.

Thank you @chemicL . Looking forward to the release!

tzolov:
Thanks for the update. Looks great @chemicL !

chemicL:
Added the documentation, cleaned up. If there are no more remarks, I'll merge this.

The occasional CI failures with STDIO Transport are a result of running `npx -y @modelcontextprotocol/server-everything stdio` within the test which actually first makes a network call to download and install the node package first and than can take several seconds. Potentially we can fix that as a follow-up and run `npm install` for that package in the `@BeforeAll` junit utility and avoid timeouts.

marianogonzalez:
@chemicL  this  in response to [this comment](https://github.com/modelcontextprotocol/java-sdk/issues/273#issuecomment-2959952592) in issue #273 .

I think this is definitively a step in the right direction. I think what'd be good for completing this is also adding access to the transport's underlying properties. For instance, it'd be great to have a way of accessing the request's headers and query params. This is important for implementing authorization schemes, custom traceability and other concerns that go beyond what this SDK (or the Spring ecosystem) should implement OOTB (I also think that enabling non-spring apps is very important too)

Also, if you're looking for input on the server side, please see issues #274 and #275 as well.

Thank you!

chemicL:
Thanks, @marianogonzalez. We are currently also exploring the mechanisms to allow OAuth2 integration and this work will also dictate some means to inspect the token presence and drive, hopefully transparent, integration throughout the stack. Most preferably, for the other concerns, it would be best to open an issue with a particular need and the intended use cases and scenarios so that we can come up with a solution. Exposing the entire transport-layer headers set makes sense in the context of HTTP, but we can also have a client backed by STDIO-driven local subprocess so some abstractions need to adjust to both cases.

Issue #298, opened 2025-06-08
Support progress flow


134130:
**Expected Behavior**

- There is a [Progress Flow](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress#progress-flow) spec on MCP.
- When the client sends the `params._meta.progressToken`, the server may send progress notifications.

**Current Behavior**

- The SDK doesn't supports it currently

**Context**

- Some clients times out when there is no progress event from server.
- Of course the specification says "MAY", but it is also true that server doesn't supports the spec.
- Also, when the tool task is long task, the UX can lead to bad, that can't tell if it's stuck or not.

tzolov:
Thank you @134130 
I'm closing this as completed

PR #300, opened 2025-06-08
Support Progress Flow for McpServer

134130:
## Motivation and Context

- There is a [Progress Flow](https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/progress#progress-flow) spec on MCP, but the SDK doesn't implement it.
- Related: #298

## How Has This Been Tested?

- I've tested with [283f8f5](https://github.com/modelcontextprotocol/java-sdk/pull/300/commits/283f8f503e3a0de8537eb53a9afc21b3f263486d), but restored, to separate PR for server and client.

## Breaking Changes

- No

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

- Client PR: #389


brant-hwang:
This is really awesome pr! I really hope this one merged as soon as possible!

134130:
Force-pushed with resolving conflict.

134130:
Force-pushed with resolving conflict, again.

tzolov:
@134130 thank you for trying to fill this missing capabilities. 

But because the progress tracking introduces a breaking changes to the existing API and goes beyond the tool calling, or server side only, I've updated the #298 issue, listing the required task to implement this capability.  
I submitted #351 and #352 to address the pre-requisite tasks. Once those are merged, if you are still interested you can re-work your PR on to of them.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
```diff
@@ -139,6 +139,16 @@ public Mono<McpSchema.ListRootsResult> listRoots(String cursor) {
 				LIST_ROOTS_RESULT_TYPE_REF);
 	}
 
+	public Mono<Void> notification(String method, Object params) {
```
tzolov: This is too generic. We would need something like `progress(ProgressNotificaiton pn)` instead.

tzolov:
@134130 feel free to review #351 and #352. 
I've tried to keep backward compatibility as much as possible.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
134130: I thought restricting the method to only “progress” would block users until the SDK releases a new protocol adaptation, and it did in some cases: `loggingNotification(LoggingMessageNotification loggingMessageNotification)`.

I agree there is also `progress(ProgressNotification pn)` should be exists.

Also, there might have users that who want to send custom notifications to.

134130:
Thanks! I'll follow up after those PRs are merged.

134130:
Updated the Code and PR Description and separated Client implementation PR.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
134130: fixed with a specific method, but please consider keeping the general one too.

Issue #285, opened 2025-05-30
Add OutputSchema to Tool and CallToolResult.structuredContent


pantanurag555:
OutputSchema was added to Tool after updates to the specification: https://github.com/modelcontextprotocol/modelcontextprotocol/commit/c2e0fd686cd9051bb4c7d8800f964ce953acb84b

I am working on changes to add it to the java-sdk. Creating an issue to avoid duplicated work.

pantanurag555:
@tzolov @chemicL Hey there, I am picking up work on these changes and wanted to clarify the scope of the work based on the [contributing guidelines](https://github.com/modelcontextprotocol/java-sdk/blob/main/CONTRIBUTING.md).

Relevant PRs for python-sdk: [#685](https://github.com/modelcontextprotocol/python-sdk/pull/685/files#diff-9b043ee327a542607e73b38ebb19c77049650efbe3a1f9d5891b0f8af09599ae) [#684](https://github.com/modelcontextprotocol/python-sdk/pull/654/files#diff-b21bf5e21c058ccfddb404c209f2d4288e4f32b121111ffc9d466638631d13e4)

Scope:
- Schema changes to add Tool.outputSchema and CallToolResult.structuredContent
- No server side changes since the tool definition and result definition are dependent on the server owner.
- Client side changes to create a caching mechanism of sorts for the tool list and leveraging this mechanism to validate the Tool.outputSchema against the CallToolResult.structuredContent when required [According to the spec: `Clients SHOULD validate structured results against this schema.`]
- Relevant tests

My major question about the scope of work revolves around the client side changes. I saw that there was some discussion around a standalone schema validator in #271. Should I introduce the caching mechanism and validator changes in the first PR itself or should that be a PR of its own? I think they can be separated for the first iteration since this is a SHOULD statement. Would like some maintainer input on this to reduce churn.

loong-coder:
Hello @pantanurag555 , could you please provide an update on the [#285](https://github.com/modelcontextprotocol/java-sdk/issues/285) ? I'd appreciate knowing its current status.

pantanurag555:
Hey @loong-coder. I have the changes but based on the contributing guidelines and comments from a maintainer in #271, I need to clarify the scope of work after creating an issue to claim work on the feature. I have tried to that in the comment above yours. If I don't hear back from them in a couple of days, I will raise the PR as is.

pantanurag555:
[...]

PR #302, opened 2025-06-09
feat: Add Tool.outputSchema and CallToolResult.structuredContent

pantanurag555:
## Motivation and Context
This PR adds support for Tool.outputSchema and CallToolResult.structuredContent that was introduced in the specification in https://github.com/modelcontextprotocol/modelcontextprotocol/commit/03dc24b3fed8534bbec9b41ee5ca2c9371852d0e. The changes and scope of work were discussed in issue: #285

Changes in this PR:
- Implement support for outputSchema in Tool and structuredContent in CallToolResult.
- Add a client-side cache that maintains track of all server tools and their output schemas (if any).
- Implement a mechanism to refresh the client side cache when encountering a tool call request for an uncached tool.
- Implement json validation between tool's output schema and result's structured content, when an output schema is specified for the tool.

## How Has This Been Tested?
- Added unit test
- Added integration tests to test success and failure of client-side json validation

## Breaking Changes
A new field for structuredContent was added to CallToolResult as part of the changes. CallToolResult fields were modified from (List\<Content\> content, Boolean isError) -> (List\<Content\>  content, Boolean isError, Map<String, Object> structuredContent).

In order to maintain backward compatibility with earlier initializations of CallToolResult that relied on just specifying (List\<Content\> , Boolean), a new constructor was created that would accept (List\<Content\> , Boolean). This ensured that no further changes were required to the unit tests across java-sdk. There is already a preexisting constructor for CallToolResult that accepts (String, Boolean). 

The addition of the new constructor can lead to constructor ambiguity in case a CallToolResult was created with content set to null (since null can be either String or List<Content>). Example, CallToolResult(null, true). Such initializations would now require explicit typecasting to ensure that they are compiled successfully. This can be treated as a breaking change. I am open to suggestions on a better way to tackle this.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-test/src/main/java/io/modelcontextprotocol/server/.LCKAbstractMcpSyncServerTests.java~
+++ b/mcp-test/src/main/java/io/modelcontextprotocol/server/.LCKAbstractMcpSyncServerTests.java~
@@ -1,0 +1,1 @@
+/local/home/pantanur/workspace/java-sdk/mcp-test/src/main/java/io/modelcontextprotocol/server/AbstractMcpSyncServerTests.java
--- a/mcp/pom.xml
+++ b/mcp/pom.xml
@@ -202,6 +202,13 @@
 			<scope>test</scope>
 		</dependency>
 
+		<!-- Json validator dependency -->
+		<dependency>
+			<groupId>com.networknt</groupId>
+			<artifactId>json-schema-validator</artifactId>
+			<version>1.5.7</version>
+		</dependency>
+
 	</dependencies>
 
 
--- a/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
@@ -14,6 +14,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+
 import io.modelcontextprotocol.spec.McpClientSession;
 import io.modelcontextprotocol.spec.McpClientSession.NotificationHandler;
 import io.modelcontextprotocol.spec.McpClientSession.RequestHandler;
@@ -33,8 +34,11 @@ import io.modelcontextprotocol.spec.McpSchema.Root;
 import io.modelcontextprotocol.spec.McpTransport;
 import io.modelcontextprotocol.util.Assert;
 import io.modelcontextprotocol.util.Utils;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import io.modelcontextprotocol.server.McpServerFeatures;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.Sinks;
@@ -124,6 +128,11 @@ public class McpAsyncClient {
 	private McpSchema.Implementation serverInfo;
 
 	/**
+	 * Cached tool output schemas.
+	 */
+	private final HashMap<String, McpSchema.JsonSchema> toolsOutputSchemaCache;
+
+	/**
 	 * Roots define the boundaries of where servers can operate within the filesystem,
 	 * allowing them to understand which directories and files they have access to.
 	 * Servers can request the list of roots from supporting clients and receive
@@ -171,6 +180,7 @@ public class McpAsyncClient {
 		this.transport = transport;
 		this.roots = new ConcurrentHashMap<>(features.roots());
 		this.initializationTimeout = initializationTimeout;
+		this.toolsOutputSchemaCache = new HashMap<>();
 
 		// Request Handlers
 		Map<String, RequestHandler<?>> requestHandlers = new HashMap<>();
@@ -285,6 +295,14 @@ public class McpAsyncClient {
 	 */
 	public McpSchema.Implementation getClientInfo() {
 		return this.clientInfo;
+	}
+
+	/**
+	 * Get the cached tool output schemas.
+	 * @return The cached tool output schemas
+	 */
+	public HashMap<String, McpSchema.JsonSchema> getToolsOutputSchemaCache() {
+		return this.toolsOutputSchemaCache;
 	}
 
 	/**
@@ -525,7 +543,13 @@ public class McpAsyncClient {
 			if (this.serverCapabilities.tools() == null) {
 				return Mono.error(new McpError("Server does not provide tools capability"));
 			}
-			return this.mcpSession.sendRequest(McpSchema.METHOD_TOOLS_CALL, callToolRequest, CALL_TOOL_RESULT_TYPE_REF);
+			// Refresh tool output schema cache, if necessary, prior to making tool call
+			Mono<Void> refreshCacheMono = Mono.empty();
+			if (!this.toolsOutputSchemaCache.containsKey(callToolRequest.name())) {
+				refreshCacheMono = refreshToolOutputSchemaCache();
+			}
+			return refreshCacheMono.then(this.mcpSession.sendRequest(McpSchema.METHOD_TOOLS_CALL, callToolRequest,
+					CALL_TOOL_RESULT_TYPE_REF));
 		});
 	}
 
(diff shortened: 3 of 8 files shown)


mcp-test/src/main/java/io/modelcontextprotocol/server/.LCKAbstractMcpSyncServerTests.java~
```diff
@@ -0,0 +1 @@
+/local/home/pantanur/workspace/java-sdk/mcp-test/src/main/java/io/modelcontextprotocol/server/AbstractMcpSyncServerTests.java
```
LucaButBoring: Temp file, remove this

mcp/pom.xml
```diff
@@ -202,6 +202,13 @@
 			<scope>test</scope>
 		</dependency>
 
+		<!-- Json validator dependency -->
+		<dependency>
+			<groupId>com.networknt</groupId>
+			<artifactId>json-schema-validator</artifactId>
+			<version>1.5.7</version>
+		</dependency>
```
LucaButBoring, line 210: We should remove this until we revisit validation (for context for others, this was initially added in the context of including validation in this PR - which is no longer happening).

LucaButBoring: Worth getting extra clarification from maintainers on if we want schema validation as part of this or a separate effort, along the lines of the discussion in #271.

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -123,6 +127,11 @@ public class McpAsyncClient {
 	 */
 	private McpSchema.Implementation serverInfo;
 
+	/**
+	 * Cached tool output schemas.
+	 */
+	private final HashMap<String, McpSchema.JsonSchema> toolsOutputSchemaCache;
```
LucaButBoring: This is being updated from `Mono`s, so this should probably be a [`ConcurrentHashMap`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html).

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
```diff
@@ -55,6 +68,8 @@ public class McpSyncClient implements AutoCloseable {
 
 	private static final Logger logger = LoggerFactory.getLogger(McpSyncClient.class);
 
+	private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
LucaButBoring: IMO make this an instance field with a constructor overload. e.g.

```java
McpSyncClient(McpAsyncClient delegate) {
    this(delegate, new ObjectMapper());
}

McpSyncClient(McpAsyncClient delegate, ObjectMapper objectMapper)
```

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
```diff
@@ -215,7 +231,53 @@ public Object ping() {
...
+					// Handle validation errors
+					throw new McpError(
+							"CallToolResult validation failed: structuredContent does not match tool outputSchema.");
+				}
+			}
+			catch (JsonProcessingException e) {
+				// Log error if output schema can't be parsed to prevent erroring out for
+				// successful call tool request
+				logger.error("Encountered exception when parsing outputSchema: {}", e);
+			}
```
LucaButBoring: Worth getting other opinions on this, I think this should be rethrown (potentially wrapped in `McpError` to keep it unchecked).

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
```diff
@@ -353,4 +415,8 @@ public McpSchema.CompleteResult completeCompletion(McpSchema.CompleteRequest com
 		return this.delegate.completeCompletion(completeRequest).block();
 	}
 
+	private void isStrict(boolean b) {
```
LucaButBoring: Seems to be unused.

LucaButBoring:
Just gave this a preliminary review by request, review done for now.

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
pantanurag555: `Mono`s should be blocking and should not cause parallel updates afaik. However, to make this compliant with different client implementations that might cause parallel updates, changed from `HashMap` to `ConcurrentHashMap`.

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
pantanurag555: Makes sense. Made changes based on how the objectMapper is initialized in SSE client transport.

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
pantanurag555: Removed.

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
pantanurag555: I agree. For now, changing this to log a warning. This leaves the option open for the client side implementation to proceed as it deems fit. I would like feedback from others about what would be the ideal behavior in this scenario.

mcp/pom.xml
pantanurag555: Discussed some of the issues offline with @LucaButBoring. I wanted to get clarity around the scope of the work on the issue that I had created. There are a few questions that we might want to answer regarding the output validation:
1. Do we want validation to be part of this PR or do we want to separate it out and think more about how we can cache/validate the results?
2. Should cache be in the client class or should we construct a separate cache class?
3. Should there be refresh intervals for cache invalidation? How else can we deal with it and should it be configurable by the client?
4. Is a Map the best way to represent cache or should we opt for an in-memory cache implementation like [Guava Cache](https://www.baeldung.com/guava-cache)?

Would like inputs from the maintainers to decide on the best way to proceed.

pantanurag555:
Addressed all comments from @LucaButBoring barring a few that where I would like a discussion with maintainers/others from the community.

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
```diff
@@ -215,7 +244,54 @@ public Object ping() {
 	 * Boolean indicating if the execution failed (true) or succeeded (false/absent)
 	 */
 	public McpSchema.CallToolResult callTool(McpSchema.CallToolRequest callToolRequest) {
-		return this.delegate.callTool(callToolRequest).block();
+		McpSchema.CallToolResult result = this.delegate.callTool(callToolRequest).block();
```
pantanurag555: Note: Shift to sync instead.

pantanurag555:
Updated the PR to leverage the existing json validator and to perform only client-side validations. Added a new test file to mock transport for the sync client.

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
```diff
@@ -63,14 +69,23 @@ public class McpSyncClient implements AutoCloseable {
 
 	private final McpAsyncClient delegate;
 
+	private final JsonSchemaValidator jsonSchemaValidator;
+
+	/**
+	 * Cached tool output schemas.
+	 */
+	private final ConcurrentHashMap<String, Optional<Map<String, Object>>> toolsOutputSchemaCache;
```
LucaButBoring: Did you wind up looking at cache abstractions like we had discussed offline a while back? Not sure if that fell through or if we just don't want to do that anymore.

I just took a look at Guava's cache implementation and they actually [recommend](https://github.com/google/guava/blob/401c2896f5c34919d0449f268e77518c0c982383/guava/src/com/google/common/cache/CacheBuilder.java#L49) using [Caffeine](https://github.com/ben-manes/caffeine) now.

---

Separate comment: we should have a comment explaining the `Optional` if we're keeping that. It's not obvious why you would want an Optional value without reading the full implementation.

LucaButBoring:
Is there a reason this was rewritten to only apply to the sync client? This should probably be in the async client, which the sync client would then benefit from automatically.

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
pantanurag555: Didn't use any cache abstractions to keep the review simple for now and avoid additional dependencies. The cache implementation as part of JsonValidator was also a similar ConcurrentHashMap so wanted to stay close to that at least for the first review of the client side validation. Additionally, python SDK implementation also used a dict initially for the cache: https://github.com/modelcontextprotocol/python-sdk/pull/685

I will make a note to add additional comments about `Optional` to better explain it in the context of the code.

pantanurag555:
> Is there a reason this was rewritten to only apply to the sync client? This should probably be in the async client, which the sync client would then benefit from automatically.

This was applied to the sync client based on offline feedback from @tzolov. The previous review had most of the implementation as part of async, however in order to refresh the tool cache, the implementation needs to block on listTools() which leads to mixing of sync and async. The feedback was to consolidate the entire implementation in sync to avoid that.

tzolov:
> > Is there a reason this was rewritten to only apply to the sync client? This should probably be in the async client, which the sync client would then benefit from automatically.
> 
> This was applied to the sync client based on offline feedback from @tzolov. The previous review had most of the implementation as part of async, however in order to refresh the tool cache, the implementation needs to block on listTools() which leads to mixing of sync and async. The feedback was to consolidate the entire implementation in sync to avoid that.

@pantanurag555 , don't have recollection about such conversation. But we must have parity between the async an sync clients. And following the current architecture the sync (e.g. blocking) client is a wrapper that delegates to the async. 

Can you elaborate on the listTools() issue?

pantanurag555:
@tzolov Updated based on our discussions. Take a look and let me know your feedback.

tzolov:
Rebased, [extended](https://github.com/modelcontextprotocol/java-sdk/commit/8e11fa82ae0291d9b29566f5ab1299ddcb1ac639), squashed and merged at 8e11fa82ae0291d9b29566f5ab1299ddcb1ac639

extensions:
  - Add JSON schema validation for tool call results against output schemas
  - Implement automatic tool output schema caching during initialization
  - Add `enableCallToolSchemaCaching` configuration option to enable/disable
    schema caching
  - Add `JsonSchemaValidator` integration to McpClient builder APIs
  - Introduce post-initialization hook mechanism for performing operations
    after successful client initialization
  - Cache tool output schemas during `listTools` operations when caching
    is enabled
  - Validate tool results against cached schemas in `callTool` operations
  - Return error CallToolResult when validation fails
  - Add test coverage
  - Convert validateToolResult from Mono to synchronous method
  - Throw IllegalArgumentException on validation errors

PR #306, opened 2025-06-11
feat: Add client side handlers to handle pagination + fix: Support pagination in tools change notification handler

pantanurag555:
## Motivation and Context
Tools change notification handler is meant to send the updated list of tools to all consumers (clients) who have signed up for tools change notifications. In the current implementation in java-sdk, this would only return the first page of tools to the consumer. This PR:

1. Introduces a fix to ensure that the tools change notification handler deals with the pagination for list tools and returns a list of all tools across pages back to the consumer.
2. Introduces list handlers for the MCP client that can be used to obtain a list of all tools, resources, prompts and resource templates respectively without making multiple list calls to deal with pagination.

## How Has This Been Tested?
Wrote a unit test that broke down list tools result across pages. Tested with the code before and after the change introduced in the PR. The old code (without the PR change) fails to return the complete list of tools. The new code (with the change) successfully returns all the tools across all pages.

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
@@ -682,9 +682,17 @@ public class McpAsyncClient {
 	private NotificationHandler asyncToolsChangeNotificationHandler(
 			List<Function<List<McpSchema.Tool>, Mono<Void>>> toolsChangeConsumers) {
 		// TODO: params are not used yet
-		return params -> this.listTools()
-			.flatMap(listToolsResult -> Flux.fromIterable(toolsChangeConsumers)
-				.flatMap(consumer -> consumer.apply(listToolsResult.tools()))
+		return params -> this.listTools().expand(result -> {
+			if (result.nextCursor() != null) {
+				return this.listTools(result.nextCursor());
+			}
+			return Mono.empty();
+		}).reduce(new ArrayList<McpSchema.Tool>(), (allTools, result) -> {
+			allTools.addAll(result.tools());
+			return allTools;
+		})
+			.flatMap(allTools -> Flux.fromIterable(toolsChangeConsumers)
+				.flatMap(consumer -> consumer.apply(allTools))
 				.onErrorResume(error -> {
 					logger.error("Error handling tools list change notification", error);
 					return Mono.empty();
--- a/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
@@ -13,7 +13,11 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.fasterxml.jackson.core.type.TypeReference;
+
 import io.modelcontextprotocol.spec.McpClientSession;
 import io.modelcontextprotocol.spec.McpClientSession.NotificationHandler;
 import io.modelcontextprotocol.spec.McpClientSession.RequestHandler;
@@ -35,8 +39,6 @@ import io.modelcontextprotocol.spec.McpSchema.Root;
 import io.modelcontextprotocol.spec.McpTransportSessionNotFoundException;
 import io.modelcontextprotocol.util.Assert;
 import io.modelcontextprotocol.util.Utils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.Sinks;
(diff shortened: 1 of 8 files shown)


mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -682,9 +682,17 @@ public Mono<McpSchema.ListToolsResult> listTools(String cursor) {
...
-				.flatMap(consumer -> consumer.apply(listToolsResult.tools()))
+		return params -> this.listTools().expand(result -> {
+			if (result.nextCursor() != null) {
+				return this.listTools(result.nextCursor());
+			}
+			return Mono.empty();
+		}).reduce(new ArrayList<McpSchema.Tool>(), (allTools, result) -> {
+			allTools.addAll(result.tools());
+			return allTools;
+		})
```
LucaButBoring: Should we maybe pull this chunk out to a `public Mono<List<Tool>> listAllTools()`? And then have a `public List<Tool> listAllTools()` in the sync client. Would be helpful for e.g. #314, I think.

LucaButBoring (changes requested):

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
LucaButBoring: Discussed offline - for some reason I thought this was resources instead of tools like the other PR. Suggested expanding to other entity types here, as pagination is a concern of several things.

tzolov:
@pantanurag555, thanks for addressing this.

We already have proper pagination APIs for all server resources:
- [McpAsyncClient#listTools(cursor)](https://github.com/modelcontextprotocol/java-sdk/blob/0fe76ba993799e8820f13311347e27692b51ae3b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java#L671)
- [McpAsyncClient#listResources(cursor)](https://github.com/modelcontextprotocol/java-sdk/blob/0fe76ba993799e8820f13311347e27692b51ae3b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java#L729)
- [McpAsyncClient#listResourceTemplates(cursor)](https://github.com/modelcontextprotocol/java-sdk/blob/0fe76ba993799e8820f13311347e27692b51ae3b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java#L790)
- [McpAsyncClient#listPrompts(cursor)](https://github.com/modelcontextprotocol/java-sdk/blob/0fe76ba993799e8820f13311347e27692b51ae3b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java#L866)

One can use `listXyz(null)` for the first page, then pass the returned cursor for subsequent pages. For convenience, we could add `public final static String FIRST_PAGE = null;` to the McpSchema.

The non-paginated methods (`listTools()`, `listResources()`, etc.) should return all items but aren't implemented correctly. 
IMO, we should move the implementations: `listAllPrompts()` → `listPrompts()`, `listAllTools()` → `listTools()`, etc., and remove the `listAllXyz()` methods.

pantanurag555:
> @pantanurag555, thanks for addressing this.
> 
> We already have proper pagination APIs for all server resources:
> 
>     * [McpAsyncClient#listTools(cursor)](https://github.com/modelcontextprotocol/java-sdk/blob/0fe76ba993799e8820f13311347e27692b51ae3b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java#L671)
> 
>     * [McpAsyncClient#listResources(cursor)](https://github.com/modelcontextprotocol/java-sdk/blob/0fe76ba993799e8820f13311347e27692b51ae3b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java#L729)
> 
>     * [McpAsyncClient#listResourceTemplates(cursor)](https://github.com/modelcontextprotocol/java-sdk/blob/0fe76ba993799e8820f13311347e27692b51ae3b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java#L790)
> 
>     * [McpAsyncClient#listPrompts(cursor)](https://github.com/modelcontextprotocol/java-sdk/blob/0fe76ba993799e8820f13311347e27692b51ae3b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java#L866)
> 
> 
> One can use `listXyz(null)` for the first page, then pass the returned cursor for subsequent pages. For convenience, we could add `public final static String FIRST_PAGE = null;` to the McpSchema.
> 
> The non-paginated methods (`listTools()`, `listResources()`, etc.) should return all items but aren't implemented correctly. IMO, we should move the implementations: `listAllPrompts()` → `listPrompts()`, `listAllTools()` → `listTools()`, etc., and remove the `listAllXyz()` methods.

That makes sense. Let me move around the implementations and make the relevant changes.

tzolov:
Thanks @pantanurag555 !

Rebased, squashed, cleaned and merged at d4a1bfa58f237f85b9b2fb807ac810298206b42c

PR #313, opened 2025-06-13
Feat: Added the annotations attribute to McpSchema.Tool

### LINKED ISSUE #312: `McpSchema.Tool` requires the annotations attribute.
**Expected Behavior**

The `McpSchema.Tool` class should support the annotations attribute.

**Current Behavior**
The `McpSchema.Tool` class currently only has the `name`, `description`, and `inputSchema` properties.

**Context**

I believe we need to support the `annotations` property in the `McpSchema.Tool` class for the following reasons:

1. **It is part of the MCP specification**: As stated in the MCP specification (https://modelcontextprotocol.io/docs/concepts/tools#tool-definition-structure
), this property is very useful as it allows the MCP server to provide additional descriptions about a tool’s behavior to the client.

2. **Better integration with the Spring ecosystem (e.g., Spring AI)**: Supporting this property would enable better interoperability with frameworks like Spring AI. In Spring AI, there is a `returnDirect` attribute for tools(https://docs.spring.io/spring-ai/reference/1.1-SNAPSHOT/api/tools.html#_return_direct), which indicates that the result of the tool call should be returned directly to the user instead of being sent back to the model. However, since `McpSchema.Tool` currently lacks a way to carry this attribute, `returnDirect` becomes ineffective in MCP scenarios. I believe it's a good approach to pass this value through `annotations`.

3. **Support and extension in other language SDKs**: The `annotations` property is already supported in MCP SDKs for languages such as `Python` and `Swift`, and each has extended it in their own way. Aligning with these implementations will improve consistency across platforms.

sunyuhan1998:
Fixes #312 
As mentioned in the issue, it is highly necessary to support annotations. This PR includes the following changes:

1. Added the `annotations` property to `McpSchema.Tool`.
2. Added corresponding unit tests.

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -738,6 +738,17 @@ public final class McpSchema {
 		@JsonProperty("definitions") Map<String, Object> definitions) {
 	} // @formatter:on
 
+	@JsonInclude(JsonInclude.Include.NON_ABSENT)
+	@JsonIgnoreProperties(ignoreUnknown = true)
+	public record ToolAnnotations( // @formatter:off
+								   @JsonProperty("title")  String title,
+								   @JsonProperty("readOnlyHint")   Boolean readOnlyHint,
+								   @JsonProperty("destructiveHint") Boolean destructiveHint,
+								   @JsonProperty("idempotentHint") Boolean idempotentHint,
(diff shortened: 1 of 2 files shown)


sunyuhan1998:
@tzolov @chemicL What do you think of this?

tzolov:
Thanks @sunyuhan1998 !
LGTM

tzolov:
Rebased and merged at ba5bc9496381574aad26f88df297413f9aa694ca

PR #318, opened 2025-06-16
Fix Streamable HTTP WebClient GET SSE handling

chemicL:
The WebClient Streamable HTTP implementation was missing the incoming message handling on the SSE channel opened with `GET` HTTP method. This cause sampling to not work when used against the typescript-sdk based `everything` server 

The specification uses the word SHOULD when talking about requests from the server to the client that they should use the same stream which is used for the related client request (tool call in this case) but apparently the typescript-sdk uses a different SSE connection to send the createMessage request. We should handle that properly.


mcp-test/src/main/java/io/modelcontextprotocol/client/AbstractMcpAsyncClientTests.java
```diff
@@ -424,6 +444,20 @@ void testInitializeWithSamplingCapability() {
 				});
 	}
 
+	@Test
+	void testInitializeWithElicitationCapability() {
```
chemicL, line 448: Added this when synchronizing between mcp/src/test and mcp-test/src/main (the files should have the same contents) but the elicitation was not synchronized properly.

tzolov (approved):
LGTM.

PR #320, opened 2025-06-17
Standard header name used in mcp

CrazyHZM:
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -142,7 +142,17 @@ public final class McpSchema {
 	 */
 	public static final class Headers {
 
+		/**
+		 * The Model Context Protocol (MCP) session ID header name.
+		 */
 		public static final String MCP_SESSION_ID = "Mcp-Session-Id";
+
+		/**
+		 * The Last-Event-ID HTTP request header reports an EventSource object's last
+		 * event ID string to the server when the user agent is to reestablish the
+		 * connection.
+		 */
+		public static final String LAST_EVENT_ID = "Last-Event-ID";
 
 	}
 
--- a/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -134,6 +134,18 @@ public final class McpSchema {
 
 	}
 
+	// ---------------------------
+	// JSON-RPC Protocol Headers
+	// ---------------------------
+	/**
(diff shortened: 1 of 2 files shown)


mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -134,6 +134,18 @@ public static final class ErrorCodes {
...
 
+	// ---------------------------
+	// JSON-RPC Protocol Headers
+	// ---------------------------
+	/**
+	 * Standard header name used in MCP JSON-RPC request and responses.
+	 */
+	public static final class Headers {
+
+		public static final String MCP_SESSION_ID = "Mcp-Session-Id";
```
chemicL, line 155: Missing javadoc

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -134,6 +134,18 @@ public static final class ErrorCodes {
 
 	}
 
+	// ---------------------------
+	// JSON-RPC Protocol Headers
+	// ---------------------------
+	/**
+	 * Standard header name used in MCP JSON-RPC request and responses.
+	 */
+	public static final class Headers {
```
chemicL, line 150: I feel that this belongs around something like `McpHttpTransport` interface rather than in the `McpSpec`, which is more about the JSON-RPC layer.

chemicL (changes requested):
Thanks, please review the comments.

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
CrazyHZM: The primary purpose here is to provide users with the MCP standard header name information. The `MCP-Session-Id `does belong to the transport category. Of course, I have omitted the `last-event-id `here. At present, I think defining these header names as part of the MCPSchema is the most reasonable approach because I haven't thought of any other more reasonable alternatives. These header names are indeed mentioned in the Specification. Do you have any better suggestions? @chemicL

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
CrazyHZM: Done.

tzolov:
Thanks you @CrazyHZM , 
I believe https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp/src/main/java/io/modelcontextprotocol/spec/HttpHeaders.java has addressed this issue? 
Closing for now. Feel free to reopen (submit another) issue if we have missed something

PR #323, opened 2025-06-18
fix(client): sendMessage() should fail immediately when connect() fails

logan-querypie:
## Motivation and Context

sendMessage() waits on closeLatch until the message endpoint is ready.
However, if the HTTP connection fails during the connect phase, sendMessage() wastes 10 seconds waiting on closeLatch.
This commit replaces CountDownLatch with CompletableFuture and calls CompletableFuture.cancel() on connect failure.
As a result, sendMessage() now fails immediately instead of waiting for 10 seconds.

## How Has This Been Tested?

Our project’s server uses this SDK to relay MCP communication, and we’ve confirmed that it works well in that environment.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
@@ -93,8 +93,8 @@ public class HttpClientSseClientTransport implements McpClientTransport {
 	/** Flag indicating if the transport is in closing state */
 	private volatile boolean isClosing = false;
 
-	/** Latch for coordinating endpoint discovery */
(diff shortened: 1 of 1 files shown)


tzolov:
Hi @logan-querypie, the 0.11.0 had refactored the HttpClientSseClientTransport. Can you confirm that the issue is still present or not?

gyeo009:
@tzolov
Confirmed: the issue is still present on main. The 0.11.0 refactoring swapped CountDownLatch for Sinks.One, but messageEndpointSink never gets an error signal when the SSE stream fails. It's actually worse now since Sinks.One.asMono() has no timeout at all (the old CountDownLatch.await() had 10s).

Opened #787 with a fix against the current codebase.

Kehrlann:
Closing as stale, in favor of #787

PR #327, opened 2025-06-19
fix: support freeform stopReason in sampling

### LINKED ISSUE #328: Freeform stopReasons in sampling responses are unsupported
**Bug description**
The spec defines some well-known `stopReason` values, but also allows arbitrary values to be passed as well ([spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/f5ccad944fdf2b7d9cc70cf817f66ca5a8aa03a4/schema/2024-11-05/schema.ts#L807)). When parsing these, the Java SDK throws an exception:

```
com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `io.modelcontextprotocol.spec.McpSchema$CreateMessageResult$StopReason` from String "arbitrary value": not one of the values accepted for Enum class: [endTurn, stopSequence, maxTokens]
```

**Environment**
JDK21

**Steps to reproduce**
See PR #327

**Expected behavior**
SDK does not throw

See PR #327

LucaButBoring:
Supports freeform stop reasons in sampling responses by mapping them to a new `UNKNOWN` value.

## Motivation and Context
Previously, sampling response parsing would fail if a stopReason was provided besides endTurn/stopSequence/maxTokens. This fixes that to map the arbitrary values to UNKNOWN instead.

Spec: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/f5ccad944fdf2b7d9cc70cf817f66ca5a8aa03a4/schema/2024-11-05/schema.ts#L807

## How Has This Been Tested?
Unit tests

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
Closes #328

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -1077,7 +1077,7 @@ public final class McpSchema {
 			@JsonProperty("endTurn") END_TURN("endTurn"),
 			@JsonProperty("stopSequence") STOP_SEQUENCE("stopSequence"),
 			@JsonProperty("maxTokens") MAX_TOKENS("maxTokens"),
-			UNKNOWN("unknown");
+			@JsonProperty("unknown") UNKNOWN("unknown");
 
 			private final String value;
 


LucaButBoring:
Closes #328

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -1072,9 +1074,24 @@ public record CreateMessageResult(// @formatter:off
 		@JsonProperty("stopReason") StopReason stopReason) {
 
 		public enum StopReason {
-			@JsonProperty("endTurn") END_TURN,
-			@JsonProperty("stopSequence") STOP_SEQUENCE,
-			@JsonProperty("maxTokens") MAX_TOKENS
+			@JsonProperty("endTurn") END_TURN("endTurn"),
+			@JsonProperty("stopSequence") STOP_SEQUENCE("stopSequence"),
+			@JsonProperty("maxTokens") MAX_TOKENS("maxTokens"),
+			UNKNOWN("unknown");
```
tzolov: I guess we need to add @JsonProperty("unknown")

tzolov (changes requested):
@LucaButBoring, thanks for addressing this. 
The UNKNOWN approach masks the original stop reason. I'm not sure how important the exact stop-reason is for downstream applications? 
A better approach would be relaxing the stop reason type to String, but this would be a breaking change. What do you think?

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
LucaButBoring: I suppose, though it should be rare to need to serialize this on the server side (serializes to `UNKNOWN` right now implicitly). May as well add a `JsonProperty` for consistency, will change this.

LucaButBoring:
> The UNKNOWN approach masks the original stop reason. I'm not sure how important the exact stop-reason is for downstream applications?
> A better approach would be relaxing the stop reason type to String, but this would be a breaking change. What do you think?

Yes, I was thinking about this - the reason I added `UNKNOWN` was to avoid needing to do this. I think there are better ways to model `StopReason` in general, but didn't want to make a breaking change without a good reason to do so. Also, I did want to keep the well-modeled known stop reasons, though it's not strictly necessary to do so. Given that we can't really have a union of enums and a string like TS/Python, I'd probably want to lean towards a Kotlin-esque approach and do this if we accepted a breaking change:

```java
public sealed class StopReason permits EndTurn, StopSequence, MaxTokens, Unknown {
    private String value;

    public StopReason(String value) {
        this.value = value;
    }

    public class EndTurn : StopReason {
        public EndTurn() {
            super("endTurn");
        }
    }

    public class StopSequence : StopReason {
        public EndTurn() {
            super("stopSequence");
        }
    }

    public class MaxTokens : StopReason {
        public MaxTokens() {
            super("maxTokens");
        }
    }

    public class Unknown : StopReason {
        public Unknown(String value) {
            super(value);
        }

        public String getValue() {
            return value;
        }
    }
}
```

That way consumers could handle it like so:
```java
if (stopReason instanceof EndTurn) {
    // ...
} /* other cases */ else if (stopReason instanceof Unknown unknownStopReason) {
    log.info("Sampling stopped due to unhandled stop reason: {}", unknownStopReason)
}
```

I'll defer to you, though - if we're willing to make a breaking change for this and prefer that it just become a `String`, we can do that, instead.

tzolov:
I like the suggested approach, but lets postpone the breaking change until it is clearly needed.

tzolov:
Rebased, squashed and merged at aff24fdf2fd24a942337900c772c2f63f1a8bbfe

PR #334, opened 2025-06-21
refactor: consolidate MCP client initialization logic

tzolov:
- Move session setup, completion, and error callbacks into doInitialize method
- Rename variables in McpClientSession for better clarity (sink -> responseSink/resultSink)

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
@@ -387,7 +387,12 @@ public class McpAsyncClient {
 		return withSession("by explicit API call", init -> Mono.just(init.get()));
 	}
 
-	private Mono<McpSchema.InitializeResult> doInitialize(McpClientSession mcpClientSession) {
+	private Mono<McpSchema.InitializeResult> doInitialize(Initialization initializaiton) {
+
+		initializaiton.setMcpClientSession(this.sessionSupplier.get());
+
+		McpClientSession mcpClientSession = initializaiton.mcpSession();
+
 		String latestVersion = this.protocolVersions.get(this.protocolVersions.size() - 1);
 
 		McpSchema.InitializeRequest initializeRequest = new McpSchema.InitializeRequest(// @formatter:off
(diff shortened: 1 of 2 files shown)


mcp/src/main/java/io/modelcontextprotocol/spec/McpClientSession.java
```diff
@@ -229,27 +229,27 @@ private String generateRequestId() {
 	public <T> Mono<T> sendRequest(String method, Object requestParams, TypeReference<T> typeRef) {
 		String requestId = this.generateRequestId();
 
-		return Mono.deferContextual(ctx -> Mono.<McpSchema.JSONRPCResponse>create(sink -> {
+		return Mono.deferContextual(ctx -> Mono.<McpSchema.JSONRPCResponse>create(responseSink -> {
```
chemicL, line 232: `responseSink` vs `resultSink` still feel quite similar. I have a different proposal.
Sink from `Mono.create` shall be called `pendingResponseSink`.
Sink for `handle` operator shall be called `deliveredResponseSink`.
WDYT?

mcp/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -387,7 +387,12 @@ public Mono<McpSchema.InitializeResult> initialize() {
 		return withSession("by explicit API call", init -> Mono.just(init.get()));
 	}
 
-	private Mono<McpSchema.InitializeResult> doInitialize(McpClientSession mcpClientSession) {
+	private Mono<McpSchema.InitializeResult> doInitialize(Initialization initializaiton) {
```
chemicL, line 390: Typo in "initializaiton" -> should be `initialization`.

chemicL (changes requested):
Nice improvement, thanks. Left suggestions.

mcp/src/main/java/io/modelcontextprotocol/spec/McpClientSession.java
tzolov: I like it

tzolov:
Addressed review comments, rebased and merged at be719f53d80d5e813521432a34e5db0e974043fa

PR #337, opened 2025-06-24
feat: implement Streamable HTTP transport and refactor SSE transport to reactive streams

tzolov:
This change improves the transport layer with reactive patterns and adds support for the latest MCP specification while maintaining backward compatibility with existing SSE transport.

- Add HttpClientStreamableHttpTransport implementing 2025-03-26 MCP Streamable HTTP spec
- Add ResponseSubscribers utility for handling SSE and JSON HTTP responses
- Refactor HttpClientSseClientTransport to use reactive streams instead of CompletableFuture
  - Replace FlowSseClient with direct reactive stream handling
  - Use Disposable-based connection management instead of CountDownLatch
  - Replace message endpoint discovery with Sinks.One approach
- Add resiliency tests using Toxiproxy for network failure scenarios
- Minor type safety improvements in StdioClientTransport and DefaultMcpTransportStream

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
@@ -1,0 +1,668 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ */
+
+package io.modelcontextprotocol.client.transport;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.net.http.HttpResponse.BodyHandler;
+import java.time.Duration;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+import org.reactivestreams.Publisher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.modelcontextprotocol.client.transport.ResponseSubscribers.ResponseEvent;
+import io.modelcontextprotocol.spec.DefaultMcpTransportSession;
+import io.modelcontextprotocol.spec.DefaultMcpTransportStream;
+import io.modelcontextprotocol.spec.McpClientTransport;
+import io.modelcontextprotocol.spec.McpError;
+import io.modelcontextprotocol.spec.McpSchema;
+import io.modelcontextprotocol.spec.McpTransportSession;
+import io.modelcontextprotocol.spec.McpTransportSessionNotFoundException;
+import io.modelcontextprotocol.spec.McpTransportStream;
+import io.modelcontextprotocol.util.Assert;
+import io.modelcontextprotocol.util.Utils;
+import reactor.core.Disposable;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.FluxSink;
+import reactor.core.publisher.Mono;
+import reactor.util.function.Tuple2;
+import reactor.util.function.Tuples;
+
+/**
+ * An implementation of the Streamable HTTP protocol as defined by the
+ * <code>2025-03-26</code> version of the MCP specification.
+ *
+ * <p>
+ * The transport is capable of resumability and reconnects. It reacts to transport-level
+ * session invalidation and will propagate {@link McpTransportSessionNotFoundException
+ * appropriate exceptions} to the higher level abstraction layer when needed in order to
+ * allow proper state management. The implementation handles servers that are stateful and
+ * provide session meta information, but can also communicate with stateless servers that
+ * do not provide a session identifier and do not support SSE streams.
+ * </p>
+ * <p>
+ * This implementation does not handle backwards compatibility with the <a href=
+ * "https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse">"HTTP
+ * with SSE" transport</a>. In order to communicate over the phased-out
+ * <code>2024-11-05</code> protocol, use {@link HttpClientSseClientTransport} or
+ * {@link WebFluxSseClientTransport}.
+ * </p>
+ *
+ * @author Christian Tzolov
+ * @see <a href=
+ * "https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http">Streamable
+ * HTTP transport specification</a>
+ */
+public class HttpClientStreamableHttpTransport implements McpClientTransport {
+
+	private static final Logger logger = LoggerFactory.getLogger(HttpClientStreamableHttpTransport.class);
+
+	private static final String DEFAULT_ENDPOINT = "/mcp";
+
+	/**
+	 * HTTP client for sending messages to the server. Uses HTTP POST over the message
+	 * endpoint
+	 */
+	private final HttpClient httpClient;
+
+	/** HTTP request builder for building requests to send messages to the server */
+	private final HttpRequest.Builder requestBuilder;
+
+	/**
+	 * Event type for JSON-RPC messages received through the SSE connection. The server
+	 * sends messages with this event type to transmit JSON-RPC protocol data.
+	 */
+	private static final String MESSAGE_EVENT_TYPE = "message";
+
+	private static final String APPLICATION_JSON = "application/json";
+
+	private static final String TEXT_EVENT_STREAM = "text/event-stream";
+
+	public static int NOT_FOUND = 404;
+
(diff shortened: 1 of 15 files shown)


mcp/src/test/java/io/modelcontextprotocol/client/HttpSseMcpAsyncClientLostConnectionTests.java
```diff
@@ -0,0 +1,137 @@
...
+		long start = System.nanoTime();
+		try {
+			proxy.toxics().get("RESET_UPSTREAM").remove();
+			proxy.toxics().get("RESET_DOWNSTREAM").remove();
+			logger.info("Reconnect took {} ms", Duration.ofNanos(System.nanoTime() - start).toMillis());
+		}
+		catch (IOException e) {
+			throw new RuntimeException("Failed to reconnect", e);
+		}
+	}
```
chemicL, line 95: It looks like this is copied from the resiliency tests, can we unify it in some shared test util?

mcp/src/test/java/io/modelcontextprotocol/client/HttpSseMcpAsyncClientLostConnectionTests.java
```diff
@@ -0,0 +1,137 @@
...
+
+	@Test
+	void testPingWithEaxctExceptionType() {
+		withClient(HttpClientSseClientTransport.builder(host).build(), mcpAsyncClient -> {
+			StepVerifier.create(mcpAsyncClient.initialize()).expectNextCount(1).verifyComplete();
+
+			disconnect();
+
+			// Veryfiy that the exception type is IOException and not TimeoutException
+			StepVerifier.create(mcpAsyncClient.ping()).expectError(IOException.class).verify();
```
chemicL, line 129: 👍

mcp/src/test/java/io/modelcontextprotocol/client/HttpClientStreamableHttpAsyncClientResiliencyTests.java
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	@Test
+	void testPingWithEaxctExceptionType() {
+		withClient(createMcpTransport(), mcpAsyncClient -> {
+			StepVerifier.create(mcpAsyncClient.initialize()).expectNextCount(1).verifyComplete();
+
+			disconnect();
+
+			StepVerifier.create(mcpAsyncClient.ping()).expectError(CompletionException.class).verify();
```
chemicL: Should we look at the cause to find an IOException maybe to make sure it's not a CompletableFuture's timeout?

mcp/src/test/java/io/modelcontextprotocol/client/AbstractMcpAsyncClientResiliencyTests.java
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * Resiliency test suite for the {@link McpAsyncClient} that can be used with different
+ * {@link McpTransport} implementations that support Streamable HTTP.
+ *
+ * The purpose of these tests is to allow validating the transport layer resiliency
+ * instead of the functionality offered by the logical layer of MCP concepts such as
+ * tools, resources, prompts, etc.
+ *
+ * @author Dariusz Jędrzejczyk
+ */
```
chemicL, line 40: We need to add a note that it's a copy from mcp-test module as we do in other copied test classes to make sure we keep them in sync.

mcp/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
```diff
@@ -0,0 +1,354 @@
...
+		 * @param throwable the error that occurred
+		 */
+		@Override
+		protected void hookOnError(Throwable throwable) {
+			this.sink.error(throwable);
+		}
+
+	}
+
+	public static class BodylessResponseLineSubscriber extends BaseSubscriber<String> {
```
chemicL:
```suggestion
	public static class BodilessResponseLineSubscriber extends BaseSubscriber<String> {
```

mcp/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
```diff
@@ -0,0 +1,354 @@
...
+import reactor.core.publisher.BaseSubscriber;
+import reactor.core.publisher.FluxSink;
+
+public class ResponseSubscribers {
+
+	/**
+	 * Represents a Server-Sent Event with its standard fields.
+	 *
+	 * @param id the event ID, may be {@code null}
+	 * @param event the event type, may be {@code null} (defaults to "message")
```
chemicL: The record does not impose any default.

mcp/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
```diff
@@ -0,0 +1,354 @@
...
+	 * @param data the event payload data, never {@code null}
+	 */
+	public static record SseEvent(String id, String event, String data) {
+	}
+
+	public record ResponseEvent(ResponseInfo responseInfo, SseEvent sseEvent, JSONRPCMessage jsonRpcMessage) {
```
chemicL: Seems responseInfo should not be null, we should add a constructor with a check.

mcp/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
```diff
@@ -0,0 +1,354 @@
...
+					var matcher = EVENT_ID_PATTERN.matcher(line);
+					if (matcher.find()) {
+						this.currentEventId.set(matcher.group(1).trim());
+					}
+				}
+				else if (line.startsWith("event:")) {
+					var matcher = EVENT_TYPE_PATTERN.matcher(line);
+					if (matcher.find()) {
+						this.currentEventType.set(matcher.group(1).trim());
+					}
```
chemicL, line 165: Actually, matchers are quite slow. Perhaps we can replace the data/id/event with `line.substring(5).trim()` where `5` is the length of `"data:"`?

mcp/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
```diff
@@ -0,0 +1,354 @@
...
+			this.eventBuilder = new StringBuilder();
+			this.currentEventId = new AtomicReference<>();
+			this.currentEventType = new AtomicReference<>();
+			this.responseInfo = responseInfo;
+		}
+
+		/**
+		 * Initializes the subscription and sets up disposal callback.
+		 * @param subscription the {@link Subscription} to the upstream line source
+		 */
```
chemicL: We don't need these javadocs on protected methods.

mcp/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
```diff
@@ -0,0 +1,354 @@
...
+				}
+			});
+		}
+
+		/**
+		 * Processes each line from the SSE stream according to the SSE protocol. Empty
+		 * lines trigger event emission, other lines are parsed for data, id, or event
+		 * type.
+		 * @param line the line to process from the SSE stream
+		 */
```
chemicL: This javadoc should only be needed at the top of the class.

stantonk:
@tzolov is there a companion PR that handles Streamable HTTP on the server side?

stantonk:
@tzolov oh wait, this may be it? https://github.com/modelcontextprotocol/java-sdk/pull/290

chemicL (commented):
I had some comments from when I started reviewing, so posting them here. Most of the concerns I did address by pushing two commits. Perhaps we can merge this and follow up with what's remaining as lower priority.

mcp/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
chemicL: https://thecontentauthority.com/blog/bodyless-vs-bodiless

mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
```diff
@@ -373,7 +375,7 @@ public Mono<Void> sendMessage(McpSchema.JSONRPCMessage sendMessage) {
 						else {
 							logger.debug("SSE connection established successfully");
 						}
-					})).onErrorComplete().subscribe();
+					})).onErrorMap(CompletionException.class, t -> t.getCause()).onErrorComplete().subscribe();
```
chemicL, line 378: this is not necessary, since it's swallowed in the next operator. And that's ok since we emit the throwable to the sink. We can clean this up afterwards.

chemicL (approved):

tzolov:
Rebased, squashed and merged at c711f8315aa1908064022afe72aa3da045577cda

PR #341, opened 2025-06-25
feat: Add ResourceLink

bzsurbhi:
Implement `ResourceLink` as defined in the [spec](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts#L666)

## Motivation and Context
Adds a missing `ResourceLink` type from the most recent specification version

## How Has This Been Tested?
Added unit tests

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -479,6 +479,22 @@ public final class McpSchema {
 		@JsonProperty("priority") Double priority) {
 	} // @formatter:on
 
+	public interface ResourceType {
+
+		String uri();
+
+		String name();
+
+		String description();
+
+		String mimeType();
+
+		Long size();
+
+		Annotations annotations();
+
+	}
+
 	/**
 	 * A known resource that the server is capable of reading.
 	 *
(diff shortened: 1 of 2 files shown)


mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -1601,6 +1605,18 @@ public Double priority() {
 		}
 	}
 
+	@JsonInclude(JsonInclude.Include.NON_ABSENT)
+	@JsonIgnoreProperties(ignoreUnknown = true)
+	public record ResourceLink( // @formatter:off
```
tzolov, line 1626: At the moment, the `ResourceLink` content is equivalent to the existing `Resource` type.
Unfortunately, Java doesn't allow record inheritance. Perhaps we should consider a common interface (not sure how to call it? Perhaps ResourceType ) that both Resource and ResourceLink would implement. 

Also it will be useful to have Builder for the ResourceLink similar to the Resource's builder

bzsurbhi: That makes sense, I'll push an update

tzolov:
Thank you @bzsurbhi 
I've renamed ResourceType interface into ResourceContent and change the field order of the ResourceLink to match the Resource one.

tzolov:
rebased, squashed, minor adjustments and merged at 2f55dd00158a85ba9a23122bb7d298a56691871f

PR #347, opened 2025-06-28
feat: add ping functionality to MCP server exchanges

tzolov:
- Add ping() method to McpAsyncServerExchange that sends ping requests to clients
- Add ping() method to McpSyncServerExchange as synchronous wrapper
- Add OBJECT_TYPE_REF constant for ping response type handling
- Create McpSyncServerExchangeTests.java with full test coverage
- Add ping integration tests across WebFlux, WebMVC, and HttpServlet transports

Follow up of #203

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #353, opened 2025-06-29
 Sends a heartbeat (ping) to all connected clients to keep connection…

FrolikovEA:
…s alive. This method sends ping notifications to all active sessions without expecting a response, which helps prevent connection timeouts.

 @return A Mono that completes when heartbeat has been sent to all sessions

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
+++ b/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
@@ -30,6 +30,7 @@ import io.modelcontextprotocol.spec.McpTransportSession;
 import io.modelcontextprotocol.spec.McpTransportSessionNotFoundException;
 import io.modelcontextprotocol.spec.McpTransportStream;
 import io.modelcontextprotocol.util.Assert;
+import io.modelcontextprotocol.util.Utils;
 import reactor.core.Disposable;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
@@ -244,7 +245,7 @@ public class WebClientStreamableHttpTransport implements McpClientTransport {
 
 			Disposable connection = webClient.post()
 				.uri(this.endpoint)
-				.accept(MediaType.TEXT_EVENT_STREAM, MediaType.APPLICATION_JSON)
+				.accept(MediaType.APPLICATION_JSON, MediaType.TEXT_EVENT_STREAM)
 				.headers(httpHeaders -> {
 					transportSession.sessionId().ifPresent(id -> httpHeaders.add("mcp-session-id", id));
 				})
@@ -287,7 +288,7 @@ public class WebClientStreamableHttpTransport implements McpClientTransport {
 								logger.trace("Received response to POST for session {}", sessionRepresentation);
 								// communicate to caller the message was delivered
 								sink.success();
-								return responseFlux(response);
+								return directResponseFlux(message, response);
 							}
 							else {
 								logger.warn("Unknown media type {} returned for POST in session {}", contentType,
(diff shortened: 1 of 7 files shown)


FrolikovEA:
Sends a heartbeat (ping) to all connected clients to keep connections alive. This method sends ping notifications to all active sessions without expecting a response, which helps prevent connection timeouts.
 @return A Mono that completes when heartbeat has been sent to all sessions

tzolov:
@FrolikovEA can you elaborate on this? What connection timeouts do you see.
Also we just added the server ping() support, e.g. you can send pings but only from within the context of a running server operation (e.g. tool call, resource read, ....): https://github.com/modelcontextprotocol/java-sdk/pull/347

FrolikovEA:
Good afternoon @tzolov 
this is a standard problem when using nargs in cursor 
this method will help to periodically ping all sessions so that the connection does not drop.
for example, this is how I call it in Spring

    @Bean
    public Disposable heartbeatLoop(WebFluxSseServerTransportProvider provider) {
        return Flux.interval(Duration.ofSeconds(40))
                .flatMap(t -> provider.sendHeartbeat())
                .subscribe();
    }

chemicL:
I am wondering if it wasn't easier if we exposed `McpAsyncServer#ping()` and `McpSyncServer#ping()` which would go over all the currently connected clients and send ping to them. That would take care of the global SSE connection with the legacy SSE transport and in Streamable HTTP it would also address the global GET SSE stream.

Why in the Mcp*Server instead of transport? Because it's a domain level method meant for heartbeat and the domain level methods are there and also in `McpAsyncServerExchange` and `McpSyncServerExchange` - these would allow you to send periodic pings in case of a slower response with the Streamable HTTP transport.

tzolov:
`McpAsyncServer#ping()` and `McpSyncServer#ping()` methods for broadcasting pings to all connected clients would be useful, following patterns from other SDKs.

**TODO:** [PING](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/ping) is a `request`, not a `notification`. The current [McpServerTransportProvider#notifyClients](https://github.com/modelcontextprotocol/java-sdk/blob/1e93776354be133528ba6e2a8832cc1fa90b5a17/mcp/src/main/java/io/modelcontextprotocol/spec/McpServerTransportProvider.java#L49) only supports notification broadcasting. We'd need something like `McpServerTransportProvider#requestClients` or `McpServerTransportProvider#pingClients`.

**Temporary solution:** @FrolikovEA could you try using the existing [McpAsyncServer#loggingNotification](https://github.com/modelcontextprotocol/java-sdk/blob/1e93776354be133528ba6e2a8832cc1fa90b5a17/mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java#L729) and [McpSyncServer#loggingNotification](https://github.com/modelcontextprotocol/java-sdk/blob/1e93776354be133528ba6e2a8832cc1fa90b5a17/mcp/src/main/java/io/modelcontextprotocol/server/McpSyncServer.java#L192) methods that already broadcast to all connected clients?

Still PING's advantage might be its silent acknowledgment with empty responses, providing bidirectional communication validation.

**Alternative:** In system-design to setup a heart-beat, one would use [HTTP keep-alive](https://en.wikipedia.org/wiki/HTTP_persistent_connection) configuration with an HTTP proxy server for production MCP deployments - likely the standard cloud setup anyway.

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java
```diff
@@ -228,6 +228,28 @@ public Mono<Void> closeGracefully() {
 			.then();
 	}
 
+	/**
+	 * Sends a heartbeat (ping) to all connected clients to keep connections alive. This
+	 * method sends ping notifications to all active sessions without expecting a
+	 * response, which helps prevent connection timeouts.
+	 * @return A Mono that completes when heartbeat has been sent to all sessions
+	 */
+	public Mono<Void> sendHeartbeat() {
```
He-Pin, line 237: we need a configuration to do the scheduling of sending

tzolov:
Resolved by https://github.com/modelcontextprotocol/java-sdk/pull/430

PR #357, opened 2025-06-30
feat: add structured output support for MCP tools

tzolov:
- Add JsonSchemaValidator interface and DefaultJsonSchemaValidator implementation
- Extend Tool schema to support outputSchema field for defining expected output structure
- Add structuredContent field to CallToolResult for validated structured responses
- Implement automatic validation of tool outputs against their defined schemas
- Add comprehensive test coverage for structured output validation scenarios
- Add json-schema-validator and json-unit-assertj dependencies for validation and testing
- Update McpServer builders to accept custom JsonSchemaValidator instances
- Ensure backward compatibility with existing tools without output schemas

This PR implements **compulsory server-side validation** for MCP tools with output schemas, adding support for JSON schema validation of tool outputs according to the MCP specification. Tools can now define output schemas that are automatically validated on the server side, ensuring structured responses conform to expected formats before being sent to clients.
Spec Reference: https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content

## Motivation and Context

The MCP specification requires that **servers** with tools that have output schemas must provide structured results that conform to those schemas. This change implements that **mandatory server-side validation requirement** by:
- Enabling tools to define expected output structure via JSON schemas
- Automatically validating tool responses against their defined schemas **on the server side**
- Providing clear error messages when validation fails
- Ensuring backward compatibility with existing tools

## How Has This Been Tested?

Test coverage has been added across all server transport implementations:

- `WebFluxSseIntegrationTests.java` - Added 4 new test methods for structured output validation
- `WebMvcSseIntegrationTests.java` - Added 4 new test methods for structured output validation
- `HttpServletSseServerTransportProviderIntegrationTests.java` - Added 4 new test methods for structured output validation
- `McpSchemaTests.java` - Added 5 new test methods for Tool schema serialization/deserialization with output schemas

## Breaking Changes

**No breaking changes.** This is a backward-compatible addition:
- Existing tools without output schemas continue to work unchanged
- New `outputSchema` and `structuredContent` fields are optional
- Default `JsonSchemaValidator` is provided automatically for server-side validation
- All existing APIs maintain their current behavior

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

**Implementation Details:**
- Added `JsonSchemaValidator` interface with `DefaultJsonSchemaValidator` implementation using `json-schema-validator` library
- Extended `Tool` record with optional `outputSchema` field
- Extended `CallToolResult` record with optional `structuredContent` field
- Implemented `StructuredOutputCallToolHandler` wrapper that validates outputs automatically **on the server side**
- Added comprehensive test utilities using `json-unit-assertj` for JSON assertions

**Design Decisions:**
- **Server-side validation** is performed automatically when tools have output schemas defined
- Invalid outputs are converted to error responses with descriptive messages **before being sent to clients**
- Structured content is also serialized to text content for backward compatibility
- Custom `JsonSchemaValidator` implementations can be provided via builder methods
- Validation only occurs for tools that explicitly define output schemas
- This implements the **compulsory server-side validation** requirement from the MCP specification

**Dependencies Added:**
- `json-schema-validator` (1.5.7) for JSON schema validation on the server
- `json-unit-assertj` for enhanced JSON testing capabilities

**Scope:**
This PR focuses exclusively on **server-side validation** as required by the MCP specification. Client-side validation is not included in this implementation.

### DIFF
--- a/mcp-spring/mcp-spring-webmvc/src/test/java/io/modelcontextprotocol/server/WebMvcSseIntegrationTests.java
+++ b/mcp-spring/mcp-spring-webmvc/src/test/java/io/modelcontextprotocol/server/WebMvcSseIntegrationTests.java
@@ -45,6 +45,10 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.awaitility.Awaitility.await;
 import static org.mockito.Mockito.mock;
+import static net.javacrumbs.jsonunit.assertj.JsonAssertions.assertThatJson;
+import static net.javacrumbs.jsonunit.assertj.JsonAssertions.json;
+
+import net.javacrumbs.jsonunit.core.Option;
 
 class WebMvcSseIntegrationTests {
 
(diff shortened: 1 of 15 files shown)


mcp-spring/mcp-spring-webmvc/src/test/java/io/modelcontextprotocol/server/WebMvcSseIntegrationTests.java
```diff
@@ -862,4 +866,242 @@ void testInitialize() {
...
+			assertThat(response).isNotNull();
+			assertThat(response.isError()).isFalse();
+
+			// In WebMVC, structured content is returned properly
+			if (response.structuredContent() != null) {
+				assertThat(response.structuredContent()).containsEntry("result", 5.0)
+					.containsEntry("operation", "2 + 3")
+					.containsEntry("timestamp", "2024-01-01T10:00:00Z");
+			}
+			else {
```
pantanurag555, line 923: Question: Didn't quite understand why this is necessary since we are hardcoding the structured content in the above result.

mcp/pom.xml
```diff
@@ -83,6 +83,15 @@
 			<artifactId>reactor-core</artifactId>
 		</dependency>
 
+		<!-- Json validator dependency.
```
pantanurag555: Comment doesn't seem right since this will always be a dependency for the default implementation.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -303,6 +311,134 @@ public Mono<Void> addTool(McpServerFeatures.AsyncToolSpecification toolSpecifica
...
+					// {}", e.getMessage());
+					// return new CallToolResult(
+					// "Failed to parse text content as structured content: " +
+					// e.getMessage(), true);
+					// }
+
+					// }
+					// }
+					return new CallToolResult(
+							"Tool call with non-empty outputSchema must have a result with structured content", true);
```
pantanurag555: Suggestion: Error message could be reworded for the client. Something like "Response missing structured content which is expected when calling tool with non-empty outputSchema"

mcp/src/main/java/io/modelcontextprotocol/server/DefaultJsonSchemaValidator.java
```diff
@@ -0,0 +1,82 @@
...
+
+			// Convert structured content in reult to JsonNode
+			JsonNode jsonStructuredOutput = this.objectMapper.valueToTree(structuredContent);
+
+			// Validate outputSchema against structuredContent
+			Set<ValidationMessage> validationResult = jsonSchema.validate(jsonStructuredOutput);
+
+			// Check if validation passed
+			if (!validationResult.isEmpty()) {
+				logger.warn("Validation failed: structuredContent does not match tool outputSchema. "
```
pantanurag555: Suggestion: Since we already have validation failed in the error StructuredOutputCallToolHandler, maybe we can omit it in this message.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -303,6 +311,134 @@ public Mono<Void> addTool(McpServerFeatures.AsyncToolSpecification toolSpecifica
...
+					logger.warn("Tool call with non-empty outputSchema MUST have a result with structured content");
+					// if (!Utils.isEmpty(result.content())) {
+					// // TODO If the tesult.content() contains json text we can try to
+					// // convert it into a structured content (Experimental)
+					// var tc = result.content().stream().filter(c -> c instanceof
+					// McpSchema.TextContent).findFirst();
+					// if (tc.isPresent()) {
+					// try {
+					// Map<String, Object> structuredOutput = new
+					// ObjectMapper().readValue(
```
LucaButBoring: If we do implicitly convert here, we should reuse the `ObjectMapper` between operations, since `ObjectMapper` instances have internal caches and are fairly expensive memory-wise.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -303,6 +311,134 @@ public Mono<Void> addTool(McpServerFeatures.AsyncToolSpecification toolSpecifica
...
+				// If an output schema is provided, servers MUST provide structured
+				// results that conform to this schema.
+				// https://modelcontextprotocol.io/specification/2025-06-18/server/tools#output-schema
+				if (result.structuredContent() == null) {
+					logger.warn("Tool call with non-empty outputSchema MUST have a result with structured content");
+					// if (!Utils.isEmpty(result.content())) {
+					// // TODO If the tesult.content() contains json text we can try to
+					// // convert it into a structured content (Experimental)
+					// var tc = result.content().stream().filter(c -> c instanceof
+					// McpSchema.TextContent).findFirst();
```
LucaButBoring: Ambivalent on this, I think it's surprising behavior with nontrivial overhead - IMO we should just pass through the raw result and allow consumers to migrate on their own.

I assume this is a mirror of converting `structuredOutput` into a `TextContent` block, but I think it's too expensive to justify.

mcp/src/main/java/io/modelcontextprotocol/server/DefaultJsonSchemaValidator.java
```diff
@@ -0,0 +1,82 @@
...
+			// Create JsonSchema validator
+			ObjectNode schemaNode = (ObjectNode) this.objectMapper
+				.readTree(this.objectMapper.writeValueAsString(schema));
+
+			// Set additional properties to false if not specified in the schema
+			if (!schemaNode.has("additionalProperties")) {
+				schemaNode.put("additionalProperties", false);
+			}
+
+			JsonSchema jsonSchema = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V202012)
```
LucaButBoring: Worth peeking inside to see what this is doing, in case it's worth caching this as a field.

mcp/src/main/java/io/modelcontextprotocol/server/DefaultJsonSchemaValidator.java
```diff
@@ -0,0 +1,82 @@
...
+ * provides methods to validate structured content against a JSON schema. It uses the
+ * NetworkNT JSON Schema Validator library for validation.
+ *
+ * @author Christian Tzolov
+ */
+public class DefaultJsonSchemaValidator implements JsonSchemaValidator {
+
+	private static final Logger logger = LoggerFactory.getLogger(DefaultJsonSchemaValidator.class);
+
+	private ObjectMapper objectMapper = new ObjectMapper();
```
LucaButBoring: nit: assigned value here appears redundant with default constructor

mcp/pom.xml
tzolov: yes, this comment is incorrect. 
I'm conscious of adding new dependencies to the MCP core. Initially I thought we could make it optional, but that would require moving the DefaultJsonSchemaValidator to a separate module or using reflection magic. This feels like overkill for now - we can always refactor later if needed.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
tzolov: agree. Thats why I commented it out but forgot to remove it.

mcp/src/main/java/io/modelcontextprotocol/server/DefaultJsonSchemaValidator.java
tzolov: The topic of schema caching is important. 
I will add some basic support.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -303,6 +311,107 @@ public Mono<Void> addTool(McpServerFeatures.AsyncToolSpecification toolSpecifica
...
+				if (result.structuredContent() == null) {
+					logger.warn(
+							"Response missing structured content which is expected when calling tool with non-empty outputSchema");
+					return new CallToolResult(
+							"Response missing structured content which is expected when calling tool with non-empty outputSchema",
+							true);
+				}
+
+				// Validate the result against the output schema
+				var validation = this.jsonSchemaValidator.validate(outputSchema, result.structuredContent());
```
tzolov, line 364: I wonder if we need to make the sever json schema validation optional?
Would validation enabled by default have some regression impact on existing MCP server implementations?

mcp/src/main/java/io/modelcontextprotocol/server/DefaultJsonSchemaValidator.java
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	public DefaultJsonSchemaValidator(ObjectMapper objectMapper) {
+		this.objectMapper = objectMapper;
+		this.schemaFactory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V202012);
+		this.schemaCache = new ConcurrentHashMap<>();
+	}
+
+	@Override
+	public ValidationResponse validate(Map<String, Object> schema, Map<String, Object> structuredContent) {
```
tzolov: Maybe we should extend the signature to include an unique `schemaName`.  Then we can expose a method `purgeCache(schemaName)` or similar

mcp/src/main/java/io/modelcontextprotocol/server/DefaultJsonSchemaValidator.java
```diff
@@ -0,0 +1,169 @@
...
+public class DefaultJsonSchemaValidator implements JsonSchemaValidator {
+
+	private static final Logger logger = LoggerFactory.getLogger(DefaultJsonSchemaValidator.class);
+
+	private final ObjectMapper objectMapper;
+
+	private final JsonSchemaFactory schemaFactory;
+
+	// TODO: Implement a strategy to purge the cache (TTL, size limit, etc.)
+	private final ConcurrentHashMap<String, JsonSchema> schemaCache;
```
tzolov: Maybe some simple, interval based purge all strategy would be good enough.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
pantanurag555: There shouldn't be any regression since none of the existing servers will have tools with output schemas.

tzolov:
I moved the JsonSchemaValidator classes from server to spec package. So we can reuse them for other validation cases.

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
LucaButBoring: I can see a niche use case for making it optional, for example if you're writing an MCP server proxy and want to force the upstream server to handle validation instead.

Also think it's fine to enable by default.

mcp/src/main/java/io/modelcontextprotocol/server/McpServer.java
```diff
@@ -624,6 +629,20 @@ public AsyncSpecification objectMapper(ObjectMapper objectMapper) {
...
+		/**
+		 * Sets the JSON schema validator to use for validating tool and resource schemas.
+		 * This ensures that the server's tools and resources conform to the expected
+		 * schema definitions.
+		 * @param jsonSchemaValidator The validator to use. Must not be null.
+		 * @return This builder instance for method chaining
+		 * @throws IllegalArgumentException if jsonSchemaValidator is null
+		 */
+		public AsyncSpecification jsonSchemaValidator(JsonSchemaValidator jsonSchemaValidator) {
+			Assert.notNull(jsonSchemaValidator, "JsonSchemaValidator must not be null");
```
LucaButBoring, line 641: nit: might be nice to add `@Nullable` and then use the JetBrains `@Contract` annotation in the `Assert` class to have control flow-aware null warnings in the IDE. Not urgent or required though.

tzolov:
@LucaButBoring , @pantanurag555 , apart of the @Nullable annotations (which we can address in a separate PR) are there any outstanding remarks with this PR?

LucaButBoring:
Sorry for the delay, catching up on things (was out of town for the past week) - I gave everything another look and it looks good as-is.

Cache purging is something we should still keep track of, but I think we can push that to a follow-up PR tracked on the roadmap, since changing schema should be a very rare use case.

tzolov:
Thanks you @LucaButBoring 

Rebasing and resolving the origin/main conflicts was involving. For a references here is a branch with the conflict resolution changes:
https://github.com/tzolov/mcp-java-sdk/tree/add-tool-structured-ouput-with-server-validation-2

tzolov:
Rebased, squashed, resolved conflicts and merged at 8a2f97f1b1995880c492159dd1b4c50266d0b579

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -303,6 +312,107 @@ public Mono<Void> addTool(McpServerFeatures.AsyncToolSpecification toolSpecifica
...
+			return tools;
+		}
+
+		return tools.stream().map(tool -> withStructuredOutputHandling(jsonSchemaValidator, tool)).toList();
+	}
+
+	private static McpServerFeatures.AsyncToolSpecification withStructuredOutputHandling(
+			JsonSchemaValidator jsonSchemaValidator, McpServerFeatures.AsyncToolSpecification toolSpecification) {
+
+		if (toolSpecification.call() instanceof StructuredOutputCallToolHandler) {
```
YunKuiLu, line 401: This 'call' field is already deprecated. Should we change it to use the 'callHandler' field here?

PR #371, opened 2025-07-03
Add option for immediate execution in McpSyncServer

Kehrlann:
Add the ability to decide whether or not to use blocking code offloading in `McpSyncServer`. Turning it off allows for the propagation of `ThreadLocal`s.

## Motivation and Context

I use this with Spring Security. Authentication data about the JWT that was used in the request is stored in a ThreadLocal. With `.subscribeOn(Schedulers.boundedElastic()))`, thread hops are introduced in blocking scenarios, and thread locals are lost.

In select case, when the user is sure they are NOT in a reactive context, they might want to enable "immediate execution", on the thread that calls `.block()`, for example in the `McpSyncServerExchange`, and preserve thread locals.

## How Has This Been Tested?

Trying this change with a [Spring-based sample](https://github.com/Kehrlann/spring-ai-mcp-authorization-demo), using Spring Security to pass the `SecurityContext` and authentication information.

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/server/McpServer.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/server/McpServer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024-2024 the original author or authors.
+ * Copyright 2024-2025 the original author or authors.
  */
 
 package io.modelcontextprotocol.server;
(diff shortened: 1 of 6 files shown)


chemicL:
Thanks for this, @Kehrlann :) Let me think tomorrow whether we can add a test for it.
@tzolov this is a change me and @Kehrlann worked on together. Instead of providing hooks for propagating `ThreadLocal` state, we instead just don't lose the ThreadLocal state by not offloading to the `boundedElastic` `Scheduler`. When using a sync server this should be a completely desired situation.

chemicL:
LGTM :) In the future we could provide Schedulers.immediate() instead of the flag, but accepting a Scheduler should not be in the interfaces of an SPI module, but an implementation that uses Reactor. For the current state of things, it looks great, thanks!

PR #374, opened 2025-07-04
Fix non-OK HTTP POST responses handling

chemicL:
Currently, JDK HttpClient SSE implementation incorrectly handles a non-2xx HTTP response code for POST. Instead of immediately failing the caller, it ignores the issue and the user awaits until the timeout kicks in.

A related problem happened with the WebClient Streamable HTTP implementation, where the error was swallowed for non-400 responses.


chemicL:
@tzolov I noticed this change fails some cases of `HttpClientSseClientTransportTests`, which seem to be incorrect, as the server returns a 400 HTTP code due to incorrect input.

Also, I noticed the `HttpClientSseClientTransportTests.TestHttpClientSseClientTransport` has a Sink which is never consumed. I don't quite get how these tests are working.

I'll leave this in a draft state for you to investigate.

tzolov:
@chemicL ,  I fixed the tests. The 400 was due to the HTTP version. 
I also changed the implementation & tests to make use of the events sink, similar to the WebFluxSseClientTransportTests. Even added a test handler to track the passing messages.
But it seems that the simulateMessageEvent is useless for both the WebFluxSseClientTransportTests and the HttpClientSseClientTransportTests

tzolov:
Rebased, squashed and merged at 95df67ec0546c0721874ba9549ce5397889c3913

PR #383, opened 2025-07-10
feat:HttpClientSseClientTransport sendMessage throw McpError when status is not success

taobaorun:
## Motivation and Context

When sendMessage encounters abnormal HTTP status codes like 404 from the server, the MCP client can only wait until the client-side timeout occurs. Moreover, the client always receives a generic Timeout exception, which isn't helpful—especially when troubleshooting issues.

To improve this, we could return an MCPError instead, including the HTTP status code in the data field. This would give clients more flexibility to handle different status codes with appropriate fallback logic. For example, if the server intercepts invalid MCP requests and returns specific status codes, the client could use this information to take proper action instead of just seeing a timeout error, which often leaves users confused about how to proceed.

## How Has This Been Tested?

Tested by use case.

```
 listTools error io.modelcontextprotocol.spec.McpError: The server returned a 404 status code
	at io.modelcontextprotocol.client.transport.HttpClientSseClientTransport.lambda$sendMessage$0(HttpClientSseClientTransport.java:430)
	Suppressed: The stacktrace has been enhanced by Reactor, refer to additional information below: 
Assembly trace from producer [reactor.core.publisher.MonoCompletionStage] :
	reactor.core.publisher.Mono.fromFuture(Mono.java:629)
	io.modelcontextprotocol.client.transport.HttpClientSseClientTransport.sendMessage(HttpClientSseClientTransport.java:424)


```

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/pom.xml
+++ b/mcp-spring/mcp-spring-webflux/pom.xml
@@ -127,6 +127,13 @@
 			<scope>test</scope>
 		</dependency>
 
+		<dependency>
+			<groupId>net.javacrumbs.json-unit</groupId>
+			<artifactId>json-unit-assertj</artifactId>
+			<version>${json-unit-assertj.version}</version>
+			<scope>test</scope>
+		</dependency>
+
 	</dependencies>
 
 
--- a/mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
+++ b/mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/WebFluxSseIntegrationTests.java
@@ -3,6 +3,8 @@
  */
 package io.modelcontextprotocol;
 
+import static net.javacrumbs.jsonunit.assertj.JsonAssertions.assertThatJson;
+import static net.javacrumbs.jsonunit.assertj.JsonAssertions.json;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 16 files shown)


taobaorun:
@tzolov please review this pr to see if it's reasonable.thanks.


`sink.error(new RuntimeException(
								"Sending message failed with a non-OK HTTP code: " + response.statusCode()));`


-->						
						
`sink.error(new McpError(new JSONRPCError(ErrorCodes.INVALID_REQUEST,
								"Sending message failed with a non-OK HTTP code: " + response.statusCode(),
								response.statusCode())));`


I saw that the code submitted on June 13 has returned RuntimeException. Can it be returned as McpError, and the status code is returned as data so that the client can get its value? I want to change RuntimeException to McpError. Is it also suitable for the scenario?

tzolov:
@taobaorun please add a test that demonstrates how the client behaves before/after the suggested changes. 

Also how would you verify this claim: "Moreover, the client always receives a generic Timeout exception,"?  A running test would be appreciated.

taobaorun:
> @taobaorun please add a test that demonstrates how the client behaves before/after the suggested changes.
> 
> Also how would you verify this claim: "Moreover, the client always receives a generic Timeout exception,"? A running test would be appreciated.

@tzolov The latest version already returns a RuntimeException, which meets the requirements. Can we change it to return a McpError exception instead, while including the statusCode in the data field? This way, clients won't need to parse the errorMessage to get the statusCode.

PR #388, opened 2025-07-11
DRAFT - HttpRequest.Builder customizer

Kehrlann:
⚠️ This a draft PR, to start a discussion. It is NOT a full implementation, has no tests, lacks docs, etc.

Provide a way to customize HTTP requests before executing them, for both `HttpClientSseClientTransport` and `HttpClientStreamableHttpTransport`.

We introduce `AsyncHttpRequestCustomizer`, which is the core class that does the heavy lifting. For blocking contexts, we provide a simpler `SyncHttpRequestCustomizer` API, which ultimately gets wrapped. The details of the implementation in both transports is not relevant (yet) ; we are focused on both customizers and also on the transpots, specifically:
- `AsyncHttpRequestCustomizer`
- `SyncHttpRequestCustomizer`
- `HttpClientSseClientTransport.Builder#requestCustomizer`
- `HttpClientStreamableHttpTransport.Builder#requestCustomizer`

## Motivation and Context

These are transport-level hooks, to allow users to modify request, e.g. to add security in headers (OAuth2 tokens, API keys, etc).

## How Has This Been Tested?

Tested with a Servlet Spring app, using `SyncHttpRequestCustomizer`.

## Breaking Changes

No.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
@@ -102,6 +102,9 @@ public class HttpClientSseClientTransport implements McpClientTransport {
 	 */
 	protected final Sinks.One<String> messageEndpointSink = Sinks.one();
 
+	// TODO
+	private final AsyncHttpRequestCustomizer httpRequestCustomizer;
+
 	/**
 	 * Creates a new transport instance with default HTTP client and object mapper.
 	 * @param baseUri the base URI of the MCP server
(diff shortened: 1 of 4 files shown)


mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
```diff
@@ -310,96 +335,111 @@ public Builder objectMapper(ObjectMapper objectMapper) {
 			return this;
 		}
 
+		/**
+		 * In reactive, DONT USE THIS. Use AsyncHttpRequestCustomizer.
+		 */
+		public Builder httpRequestCustomizer(SyncHttpRequestCustomizer syncHttpRequestCustomizer) {
```
tzolov, line 353: This feels awkward. Perhaps we should leave only the async method and let the user convert the sync customizer outside

Kehrlann: I see your point.

I’m trying to ensure sync-based users don’t have to think about async versions of this API at all. My goal is for the user to make a “sync customizer” and only have to worry about that. Making the transformation outside is one more layer they’d have to think about (and that we would have to explain in the docs).
Also, we’ve had feedback from people absolutely NOT wanting to deal with async.

let me know what you think; happy to make the change to “only async” if you think it’s more confusing than helpful.

Kehrlann:
Nice! I’ll tidy up the PR then.

for WebClient-based transport, the WebClient has transformation primitives already (for ex ExchangeFilterFunction), so no need to add an MCP-specific customizer.

mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
chemicL: I completely understand the need for this. Would this work better:

```java
public Builder httpRequestCustomizer(SyncHttpRequestCustomizer syncHttpRequestCustomizer)
public Builder asyncHttpRequestCustomizer(AsyncHttpRequestCustomizer asyncHttpRequestCustomizer)
```

The user is now forced to understand there are two ways to go about it and if they are:
* on a Servlet/MVC Thread,
* not waiting for any async computation
* on a non-blocking Thread, but are not blocking
they can freely use the sync customizer. It would actually be the default choice.

For advanced cases, e.g. when the customizer needs to wait for something to happen - e.g. authorization token to be provided via a `Publisher` before it can provide the `HttpRequest.Builder`, it should go via the async route. We can document this.

tzolov:
Thanks @Kehrlann

YunKuiLu:
Would you consider renaming `AsyncHttpRequestCustomizer` to `McpAsyncHttpRequestCustomizer` to make it clearer that the class is specific to MCP?

Kehrlann:
@YunKuiLu I like the idea but unsure whether we want to introduce a breaking change.
@tzolov @chemicL thoughts?

chemicL:
A very recent class that's not intended for mainstream daily usage.. I think it should be acceptable. I'll leave it to @tzolov to make the final call.

PR #401, opened 2025-07-15
Enforce String or Integer non-null ID for MCP JSON-RPC Messages with a new McpSchema subclass

ZachGerman:
I considered `JSONRPCMessageID`, but didn't want the `JSONRPC` prefix as it might imply nullability or possibly a float, as is standard for a JSON-RPC `id`.
I also considered `id` but figured it was too ambiguous.
I can see an argument for `McpMessageId`, but I figured it being in the `McpSchema` namespace was good enough for someone to identify the `MessageId` type as specific to MCP.

## Motivation and Context
The specification conditions for JSON-RPC Message `id` values include the requirements of the value being a (non-null) String or Integer.

## How Has This Been Tested?
Multiple existing tests use this field, have been updated accordingly, and confirmed to work correctly after the changes.

## Breaking Changes
N/A

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/client/transport/WebFluxSseClientTransportTests.java
+++ b/mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/client/transport/WebFluxSseClientTransportTests.java
@@ -12,6 +12,7 @@ import java.util.function.Function;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.modelcontextprotocol.spec.McpSchema;
 import io.modelcontextprotocol.spec.McpSchema.JSONRPCRequest;
+import io.modelcontextprotocol.spec.McpSchema.MessageId;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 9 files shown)


tzolov:
@ZachGerman if you just want to validate the request ID this looks unnecessarily  complicated? 

Wouldn't something like this achieve the same:

```java
	@JsonInclude(JsonInclude.Include.NON_ABSENT)
	@JsonIgnoreProperties(ignoreUnknown = true)
	public record JSONRPCRequest( // @formatter:off
		@JsonProperty("jsonrpc") String jsonrpc,
		@JsonProperty("method") String method,
		@JsonProperty("id") Object id,
		@JsonProperty("params") Object params) implements JSONRPCMessage { // @formatter:on

		/**
		 * Constructor that validates MCP-specific ID requirements. Unlike base JSON-RPC,
		 * MCP requires that: (1) Requests MUST include a string or integer ID; (2) The ID
		 * MUST NOT be null
		 */
		public JSONRPCRequest {
			Assert.notNull(id, "MCP requests MUST include an ID - null IDs are not allowed");
			Assert.isTrue(id instanceof String || id instanceof Integer || id instanceof Long,
					"MCP requests MUST have an ID that is either a string or integer");
		}
	}
```

tzolov:
@ZachGerman here is how I would address this requirement. Let me know if i'm missing something:
https://github.com/modelcontextprotocol/java-sdk/pull/403

ZachGerman:
@tzolov nice catch on the Long case!

> @ZachGerman here is how I would address this requirement. Let me know if i'm missing something: #403

I think it's great. I just wanted to be sure the folks writing tools are informed of the rules conflicting with base JSON-RPC somewhere along the lines when confronting the types they use to write their tool. In hindsight, this was a bit like taking a flamethrower to an anthill 😅.

PR #413, opened 2025-07-18
fix: improve compatibility with non-compliant MCP servers (#406)

### LINKED ISSUE #406: Streamable HTTP transport hangs during initialization
The client hangs during the first `initialize` message to a (public) Shopify MCP `https://my-best-shoes.myshopify.com/api/mcp`

I can successfully connect `https://my-best-shoes.myshopify.com/api/mcp` from `npx @modelcontextprotocol/inspector`.

Note that #396 describes that the client transport hangs during sendNotification for `notifications/initialized`. In this case, the client hangs even earlier - during `initialize`.

Steps to reproduce:
```
String url = "https://my-best-shoes.myshopify.com/api";
McpClientTransport transport = HttpClientStreamableHttpTransport.builder(url).endpoint("/mcp").build();

McpSyncClient client = McpClient.sync(transport)
    .requestTimeout(Duration.ofSeconds(20))
    .initializationTimeout(Duration.ofSeconds(20))
    .build();

try {
    client.initialize();

    McpSchema.ListToolsResult tools = client.listTools();
    logger.log(Level.INFO, "Available tools: {0}", tools.tools());

} catch (Exception e) {
    logger.log(Level.SEVERE, "Failed to initialize MCP client from DebugServlet", e);
}
```

tzolov:
Addresses issues with servers like Shopify that violate MCP/HTTP specs:
- Prioritize application/json in Accept headers to fix content-type issues
- Handle empty notification bodies ({}) that should be bodiless per spec
- Add status code validation and null safety improvements

Resolves #406

Improves HTTP transport robustness to handle non-compliant MCP servers that violate MCP and HTTP specifications.

## Motivation and Context

This change addresses compatibility issues with real-world MCP servers that don't strictly follow MCP and HTTP specifications. Specifically, the Shopify MCP server exhibits two problematic behaviors:

1. **Incorrect content-type negotiation**: Returns `text/streamable` content-type when the Accept header prioritizes `text/stream` over `application/json`, but correctly returns `application/json` when the latter is first
2. **MCP spec violation**: Sends empty JSON bodies (`{}`) for notification messages that should be bodiless according to the MCP specification

These violations cause the Java SDK to fail when communicating with such servers, limiting its real-world applicability.

## How Has This Been Tested?

- Tested against the Shopify MCP server that exhibits the problematic behaviors
- Verified that existing compliant servers continue to work correctly
- Tested both WebFlux and standard HTTP client transport implementations
- Validated proper handling of empty responses and status code validation

## Breaking Changes

No breaking changes. This is a backward-compatible improvement that maintains existing API contracts while adding robustness.

## Types of changes

- [ ] Documentation update

## Checklist


mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
```diff
@@ -387,9 +388,14 @@ private static String sessionIdOrPlaceholder(McpTransportSession<?> transportSes
...
-				McpSchema.JSONRPCMessage jsonRpcResponse = McpSchema.deserializeJsonRpcMessage(objectMapper,
-						responseMessage);
-				s.next(List.of(jsonRpcResponse));
+				if (Utils.hasText(responseMessage) && !responseMessage.trim().equals("{}")) {
+					McpSchema.JSONRPCMessage jsonRpcResponse = McpSchema.deserializeJsonRpcMessage(objectMapper,
+							responseMessage);
+					s.next(List.of(jsonRpcResponse));
+				}
+				else {
+					logger.warn("Received empty response message: {}", responseMessage);
```
chemicL: I am a bit worried that if the request body was an MCP Request then we will never complete a pending operation that awaits for it. Please consider adding a check whether the original message was a request and if so send an error to the Flux (using `handle`'s sink).

mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
```diff
@@ -436,11 +436,19 @@ else if (contentType.contains(TEXT_EVENT_STREAM)) {
...
+							catch (IOException e) {
+								return Mono.error(e);
+							}
 						}
-						catch (IOException e) {
-							return Mono.error(e);
+						else {
+							// No content type means no response body
+							logger.debug("No content type returned for POST in session {}", sessionRepresentation);
+							return Mono.empty();
```
chemicL: Same comment as for the WebClient implementation.

mcp/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
```diff
@@ -135,36 +135,46 @@ protected void hookOnSubscribe(Subscription subscription) {
...
+							this.currentEventType.set(matcher.group(1).trim());
+						}
 					}
 				}
 			}
+			else {
+				// If the response is not successful, emit an error
+				System.out.println("Received non-successful response: " + this.responseInfo.statusCode());
+				SseEvent sseEvent = new SseEvent(null, null, null);
+				this.sink.next(new SseResponseEvent(responseInfo, sseEvent));
```
chemicL: This looks unfinished. Did you intend to use the logger and call `sink.error` ?

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
tzolov: Unfortunately if we receive an empty response message (e.g. `null`, `" "` or `{}`) there is no way to determine if it is a bogus response from a previous request,  compliment notification response or bogus request. 
Here we assume that if an empty even is received thread it as incorrect notification response.

mcp/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
tzolov: Yeh good catch. I've mistakenly commit some experiments. Above logic is incorrect.

mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
tzolov: it is the same problem as above

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
chemicL: I think we have to cancel the request's sink though. We end up in this method if the server responded with an application/json content type. That means semantically we are to consume a response here. If there is nothing, we have to clean up the request sink, otherwise we are creating memory leaks and also not really doing the users of the client a favour not letting them know that the request has failed to fulfil the promise of a response.

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
```diff
@@ -384,12 +385,18 @@ private static String sessionIdOrPlaceholder(McpTransportSession<?> transportSes
...
-	private Flux<McpSchema.JSONRPCMessage> responseFlux(ClientResponse response) {
+	private Flux<McpSchema.JSONRPCMessage> directResponseFlux(McpSchema.JSONRPCMessage sentMessage,
+			ClientResponse response) {
 		return response.bodyToMono(String.class).<Iterable<McpSchema.JSONRPCMessage>>handle((responseMessage, s) -> {
 			try {
-				McpSchema.JSONRPCMessage jsonRpcResponse = McpSchema.deserializeJsonRpcMessage(objectMapper,
-						responseMessage);
-				s.next(List.of(jsonRpcResponse));
+				if (sentMessage instanceof McpSchema.JSONRPCNotification && Utils.hasText(responseMessage)) {
+					logger.warn("Notificaiton: {} received non-compliant response: {}", sentMessage, responseMessage);
```
chemicL:
```suggestion
					logger.warn("Notification: {} received non-compliant response: {}", sentMessage, responseMessage);
```

mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
```diff
@@ -436,6 +436,11 @@ else if (contentType.contains(TEXT_EVENT_STREAM)) {
 					else if (contentType.contains(APPLICATION_JSON)) {
 						messageSink.success();
 						String data = ((ResponseSubscribers.AggregateResponseEvent) responseEvent).data();
+						if (sentMessage instanceof McpSchema.JSONRPCNotification && Utils.hasText(data)) {
+							logger.warn("Notificaiton: {} received non-compliant response: {}", sentMessage, data);
```
chemicL:
```suggestion
							logger.warn("Notification: {} received non-compliant response: {}", sentMessage, data);
```

mcp/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
```diff
@@ -164,6 +166,12 @@ else if (line.startsWith("event:")) {
 						this.currentEventType.set(matcher.group(1).trim());
 					}
 				}
+				else {
+					// If the response is not successful, emit an error
+					this.sink.error(new McpError(
```
chemicL, line 172: I believe this should be a `McpTransportError`, please keep a note of it :)

tzolov:
Rebased, squashed and merged at 4937fc1ef5317fee4b8febf2c623e1796956bfbf

PR #404, opened 2025-07-16
feat: implement MCP protocol version header support

tzolov:
Add MCP-Protocol-Version header to all HTTP requests as required by MCP specification 2025-06-18. This enables proper server-side version negotiation and backwards compatibility handling.

- Add protocol version constants (2025-06-18 for streamable HTTP, 2024-11-05 for SSE)
- Include MCP-Protocol-Version header in all GET/POST/DELETE requests
- Update WebClientStreamableHttpTransport, WebFluxSseClientTransport, HttpClientSseClientTransport, and HttpClientStreamableHttpTransport

Complies with MCP specification requirement that HTTP clients MUST include protocol version header on all requests to MCP servers.

Related to #398 , #363 , #250

**Implementation Details:**
- Added `MCP_PROTOCOL_VERSION` and `MCP_PROTOCOL_VERSION_HEADER_NAME` constants to all HTTP transport classes
- Uses protocol version `2025-06-18` for streamable HTTP transports (WebClientStreamableHttpTransport, HttpClientStreamableHttpTransport)
- Uses protocol version `2024-11-05` for SSE transports (WebFluxSseClientTransport, HttpClientSseClientTransport) for backwards compatibility
- Header is included in all HTTP requests (GET, POST, DELETE) across all transport implementations

**Specification Reference:**
- [MCP Specification - Protocol Version Header](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#protocol-version-header)

This change ensures full compliance with the MCP specification requirement that "the client MUST include the MCP-Protocol-Version HTTP header on all subsequent requests to the MCP server."

Add MCP protocol version header support to all HTTP transport implementations as required by MCP specification 2025-06-18.

## Motivation and Context

The MCP specification 2025-06-18 requires that HTTP clients **MUST** include the `MCP-Protocol-Version` header on all requests to MCP servers. This enables:
Without this header, servers cannot properly handle version-specific behavior and may not function correctly with clients using different protocol versions.

## How Has This Been Tested?

- All existing unit tests pass with the changes
- HTTP requests now include the MCP-Protocol-Version header in all transport implementations
- Tested with both streamable HTTP transport (2025-06-18) and SSE transport (2024-11-05) scenarios
- Verified header is included in GET, POST, and DELETE requests across all affected transport classes

## Breaking Changes

**No breaking changes.** This is a backwards-compatible addition:
- Existing client code continues to work without modification
- Only adds required headers to HTTP requests
- Servers that don't check for the header will continue to function normally
[...]

PR #420, opened 2025-07-22
Streamable HTTP Server abstractions and WebFlux transport provider

chemicL:
Introducing the Streamable HTTP abstractions and an initial WebFlux-based implementation for the transport.

It involves:
* A stateful, session-based stack for the full experience with bi-directional communication.
* A stateless variant which allows for effortless horizontal scalability and a REST-like experience for MCP servers that don't need to keep track of sessions and the bidirectionality aspect

Some limitations have been taken into account to make the initial implementation pragmatic:
* In the case of sessions, each MCP request is converted to an SSE stream for simplicity and allowing the user to implement more complex interactions involving requesting from/ and notifying the client without much hassle.
* In the case of stateless servers, each MCP request is handled via direct application/json response. This is caused by the potential risks of transporting a request to a client that has no means to direct the response to the appropriate instance in case of scaling. With a session ID it would be possible to devise a routing mechanism in the infrastructure to find a session leader to handle it. With a simple, stateless server there are no means to resolve this conflict.

A more comprehensive explanation of the architectural changes can be provided later upon request :-)

Based on the research started in cooperation with @ZachGerman in #290.

### DIFF
--- a/mcp/src/test/java/io/modelcontextprotocol/server/McpCompletionTests.java
+++ b/mcp/src/test/java/io/modelcontextprotocol/server/McpCompletionTests.java
@@ -41,6 +41,8 @@ class McpCompletionTests {
 
 	private HttpServletSseServerTransportProvider mcpServerTransportProvider;
 
+	private static final int PORT = TomcatTestUtil.findAvailablePort();
+
 	private static final String CUSTOM_MESSAGE_ENDPOINT = "/otherPath/mcp/message";
 
 	McpClient.SyncSpec clientBuilder;
@@ -55,7 +57,7 @@ class McpCompletionTests {
 			.messageEndpoint(CUSTOM_MESSAGE_ENDPOINT)
 			.build();
 
-		tomcat = TomcatTestUtil.createTomcatServer("", 3400, mcpServerTransportProvider);
+		tomcat = TomcatTestUtil.createTomcatServer("", PORT, mcpServerTransportProvider);
 		try {
 			tomcat.start();
 			assertThat(tomcat.getServer().getState()).isEqualTo(LifecycleState.STARTED);
@@ -64,7 +66,7 @@ class McpCompletionTests {
 			throw new RuntimeException("Failed to start Tomcat", e);
 		}
 
-		this.clientBuilder = McpClient.sync(HttpClientSseClientTransport.builder("http://localhost:" + 3400).build());
+		this.clientBuilder = McpClient.sync(HttpClientSseClientTransport.builder("http://localhost:" + PORT).build());
 	}
 
 	@AfterEach


He-Pin:
can we support sse and streamble at the same time ？

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -70,8 +73,9 @@ public class McpServerSession implements McpSession {
 	 * @param notificationHandlers map of notification handlers to use
 	 */
 	public McpServerSession(String id, Duration requestTimeout, McpServerTransport transport,
-			InitRequestHandler initHandler, InitNotificationHandler initNotificationHandler,
-			Map<String, RequestHandler<?>> requestHandlers, Map<String, NotificationHandler> notificationHandlers) {
+			McpInitRequestHandler initHandler, InitNotificationHandler initNotificationHandler,
```
tzolov, line 102: InitNotificationHandler -> McpInitNotificationHandler

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -9,6 +9,9 @@
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import io.modelcontextprotocol.server.McpAsyncServerExchange;
+import io.modelcontextprotocol.server.McpInitRequestHandler;
```
tzolov, line 12: I believe the McpInitRequestHandler, McpNotificationHandler, McpRequestHandler should be under the spec package?

mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
```diff
@@ -0,0 +1,325 @@
...
+		if (listeningStream != null) {
+			listeningStream.close();
+		}
+		// TODO: Also close all open streams
+	}
+
+	/**
+	 * Request handler for the initialization request.
+	 */
+	public interface InitRequestHandler {
```
tzolov, line 258: Don't think this is used and should be removed?

mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
```diff
@@ -0,0 +1,325 @@
...
+		 * @return a Mono that will emit the result of the initialization
+		 */
+		Mono<McpSchema.InitializeResult> handle(McpSchema.InitializeRequest initializeRequest);
+
+	}
+
+	/**
+	 * Notification handler for the initialization notification from the client.
+	 */
+	public interface InitNotificationHandler {
```
tzolov: Don't think this is used and should be removed?

ZachGerman:
> can we support sse and streamble at the same time ？

We could with spec support for something like [this](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/1055), amongst other routes.

gemo12123:
Will WebMVC based implementations be provided in the future? My project is based on WebMVC and has been running for some time now. I don't want to modify it to WebFlux anymore.

weicheng113:
I agree with @gemo12123 . Plain Java without reactive version should be supported, especially with Project Loom.

mcp/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
```diff
@@ -0,0 +1,571 @@
...
+	 * This method is package-private and used for test only. Should not be called by user
+	 * code.
+	 * @param protocolVersions the Client supported protocol versions.
+	 */
+	void setProtocolVersions(List<String> protocolVersions) {
+		this.protocolVersions = protocolVersions;
+	}
+
+	@FunctionalInterface
+	interface RequestHandler<T> extends BiFunction<McpTransportContext, Object, Mono<T>> {
```
tzolov: - This should be called `StatlessRequestHandler`. 
- Also all request and notification handlers should be grouped in common (parent) package inside the spec.

mcp/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
```diff
@@ -0,0 +1,571 @@
...
+		this.instructions = features.instructions();
+		this.tools.addAll(features.tools());
+		this.resources.putAll(features.resources());
+		this.resourceTemplates.addAll(features.resourceTemplates());
+		this.prompts.putAll(features.prompts());
+		this.completions.putAll(features.completions());
+		this.uriTemplateManagerFactory = uriTemplateManagerFactory;
+		this.jsonSchemaValidator = jsonSchemaValidator;
+
+		Map<String, RequestHandler<?>> requestHandlers = new HashMap<>();
```
tzolov: I believe we will need notification handling as well. I guess we will need stateless notification handler as well.

mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
```diff
@@ -0,0 +1,325 @@
...
+				.handle(new McpAsyncServerExchange(this.id, stream, clientCapabilities.get(), clientInfo.get(),
+						transportContext), jsonrpcRequest.params())
+				.map(result -> new McpSchema.JSONRPCResponse(McpSchema.JSONRPC_VERSION, jsonrpcRequest.id(), result,
+						null))
+				.flatMap(transport::sendMessage)
+				.then(transport.closeGracefully());
+		});
+	}
+
+	public Mono<Void> accept(McpSchema.JSONRPCNotification notification) {
```
tzolov, line 190: The `accept(McpSchema.JSONRPCNotification notification)` and `accept(McpSchema.JSONRPCResponse response)` methods appear to be a split version of `McpServerSession#handle(McpSchema.JSONRPCMessage message)`. For consistency, I’d prefer we stick with a single handler method.

Additionally, I believe a `handle(Message)` abstraction should be part of the `McpSession` interface. Since `McpSession` represents bidirectional JSON-RPC communication—already exposing `sendXyz(Message)` for outgoing messages—it makes sense to also include a `handle(Message)` method for incoming communication.

In fact all McpSession implement some form of handle(Msg). This applies for the clients as well

mcp/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
```diff
@@ -0,0 +1,571 @@
...
+			requestHandlers.put(McpSchema.METHOD_PROMPT_LIST, promptsListRequestHandler());
+			requestHandlers.put(McpSchema.METHOD_PROMPT_GET, promptsGetRequestHandler());
+		}
+
+		// Add completion API handlers if the completion capability is enabled
+		if (this.serverCapabilities.completions() != null) {
+			requestHandlers.put(McpSchema.METHOD_COMPLETION_COMPLETE, completionCompleteRequestHandler());
+		}
+
+		mcpTransport.setRequestHandler((context, request) -> requestHandlers.get(request.method())
```
tzolov: It is a good point that we don't use `McpSession` as the stateless server is unidirectional and should support only the handle(Message) for request and notifications. 

Related to my other comment about the McpSession not having handle(msg). Perhaps we can introduce `McpHandlerSession` that has `handle(JSONRPCMessage)` used by the stateless server). Then the McpSession can extend it.

mcp/src/main/java/io/modelcontextprotocol/spec/McpStatelessServerTransport.java
```diff
@@ -0,0 +1,27 @@
+package io.modelcontextprotocol.spec;
+
+import reactor.core.publisher.Mono;
+
+import java.util.function.BiFunction;
+
+public interface McpStatelessServerTransport {
+
+	void setRequestHandler(
+			BiFunction<McpTransportContext, McpSchema.JSONRPCRequest, Mono<McpSchema.JSONRPCResponse>> message);
```
tzolov: The signature of the setRequestHandler needs to change from JSONRPCRequest to JSONRPCMessage to allow handling both requests and notifications?

mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
```diff
@@ -0,0 +1,325 @@
...
+
+		McpStreamableServerSessionInit startSession(McpSchema.InitializeRequest initializeRequest);
+
+	}
+
+	public record McpStreamableServerSessionInit(McpStreamableServerSession session,
+			Mono<McpSchema.InitializeResult> initResult) {
+	}
+
+	public final class McpStreamableServerSessionStream implements McpSession {
```
tzolov: The `McpStreamableServerSessionStream` isn't bidirectional like `McpSession`. 
We can split this into two interfaces:
- `McpRpcSender` - handles outgoing messages (copies current McpSession functionality)
- `McpRpcReceiver` - handles incoming messages with `handle(JSONRPCMessage)`
(names  be refined)

McpSession would implement both interfaces, McpStreamableServerSessionStream would only implement McpRpcSender, and the Stateless Server would use McpRpcReceiver.

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -44,7 +44,7 @@ public final class McpSchema {
 	private McpSchema() {
 	}
 
-	public static final String LATEST_PROTOCOL_VERSION = "2024-11-05";
+	public static final String LATEST_PROTOCOL_VERSION = "2025-03-26";
```
tzolov, line 47: The LATEST_PROTOCOL_VERSION unfortunately is used for all client and server transports and raising it to `2025-03-26` would likely cause misbehaviors. 
One temporal solution is to let the transport implementations determine the protocol version.  Something like this suggestion: https://github.com/modelcontextprotocol/java-sdk/pull/404  
Then the init result should pick those version. Maybe the MpcTransport can have getSupporteProtocolVersion field or a like

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java
```diff
@@ -263,17 +261,28 @@ private Mono<ServerResponse> handlePost(ServerRequest request) {
 					McpStreamableServerSession.McpStreamableServerSessionInit init = this.sessionFactory
 						.startSession(initializeRequest);
 					sessions.put(init.session().getId(), init.session());
-					return init.initResult()
+					return init.initResult().map(initializeResult -> {
+						McpSchema.JSONRPCResponse jsonrpcResponse = new McpSchema.JSONRPCResponse(
+								McpSchema.JSONRPC_VERSION, jsonrpcRequest.id(), initializeResult, null);
+						try {
+							return this.objectMapper.writeValueAsString(jsonrpcResponse);
```
tzolov, line 212: This seems redundant as the WebFlux would do the conversion?

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
chemicL: Actually, this notification handler can go away. I intend to deprecate this constructor in favor of a one that doesn't have it. We actually provide an empty handler in `McpAsyncServer`, so this is useless.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
chemicL: They are server-specific and work either with the exchange or with the InitializeRequest, which itself is handled by the server side.

mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
chemicL: It is used. `McpAsyncServer` creates one and feeds it into `DefaultMcpStreamableServerSessionFactory` that gets passed onto the `McpStreamableServerTransportProvider`.

mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
chemicL: Correct, removed.

mcp/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
chemicL: I'll try to come up with a shared abstraction to contain these handlers under an composite class.

mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
chemicL: >The accept(McpSchema.JSONRPCNotification notification) and accept(McpSchema.JSONRPCResponse response) methods appear to be a split version of McpServerSession#handle(McpSchema.JSONRPCMessage message). For consistency, I’d prefer we stick with a single handler method.

Unfortunately, with `handle(McpSchema.JSONRPCMessage message)` you are communicating to the implementor that all of possible messages can be passed in. That is not my intention. I specifically want to convey that either a response or a notification is handled via this `accept` method. Handling requests requires providing a dedicated transport, hence it has a different signature.

Regarding the rest of the comment, my changes are intended to introduce minimal breakages. Without the per-request SSE streams the above would make sense potentially, but with the specific transport argument for request handling and lack of such for notifications and responses makes it a non-goal for this PR. I'm open to discussion, but I believe opening this pandora's box is exactly what I tried to avoid here. The reason is that with a generic handle you have to manage the specific transport routing inside the session, something that is not a concern of the session itself, but of the transport. This is something explored in #290 that I tried to make less error prone with concrete transport instances that are 1-1 mapped to a request.

mcp/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
chemicL: I intend to add a composite for the notification and request handlers. But I'd suggest avoiding the abstraction of `handle(JSONRPCMessage)` as that conveys the notion that we can handle responses here - we should not, that is an error in the stateless server (at least in my view). Also, I suggest we avoid using "Session" as part of this composite's name to avoid even further confusion for stateless server authors - the intention for them is to handle "sessionless" situations, so a session (despite being a JSON-RPC layer concept from our early implementation) would be rather confusing.

mcp/src/main/java/io/modelcontextprotocol/spec/McpStatelessServerTransport.java
chemicL: As explained in the comments above, a composite will cover requests and notifications.

mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
chemicL: I think this comment is outdated after our 1-1 sync? `McpStreamableServerSessionStream` is bidirectional.

mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
chemicL: This is not something I wanted to deal with in this PR but I guess we have to. I'm not sure how #404 addresses this? The protocol version handling in McpAsyncServer deals with the JSON-RPC MCP domain-level checks and that's what I think was failing when I stumbled upon the need to update this. I think we just have the "latest" concept so it is accurate. Please advise how we can go about this if you have more insight, I was unfortunately quite ignorant to this aspect :(

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java
chemicL: I suppose so! Although it's a temporary mitigation, I think we should deal with the init result conversion into JSON at an earlier layer. Need to come back to this.

chemicL:
> can we support sse and streamble at the same time ？

@He-Pin We decided not to provide a mix of both but as a user you are able to run the legacy SSE server. If you decide to use a Streamable HTTP transport provider it will not support the legacy mode. The spec is already quite involved and adding another dimension of complexity felt like not something we want to pursue at this moment.

> Will WebMVC based implementations be provided in the future? My project is based on WebMVC and has been running for some time now. I don't want to modify it to WebFlux anymore.

@gemo12123 It's delivered in #425.

> I agree with @gemo12123 . Plain Java without reactive version should be supported, especially with Project Loom.

@weicheng113 the reactive core is there because we support a Java 17 baseline that has no Virtual Threads. To achieve the scalability and benefits that the reactive core brings you'd need Virtual Threads with almost no pinning (Java 23) and Structured Concurrency which is still in preview. We are absolutely open to creating a parallel, synchronous-only API, but adding features to adhere to the specification is our primary concern right now as you can see in this PR and all the hard work that the contributors have put in over the last weeks.

stantonk:
@chemicL this is awesome! Thank you so, so much for taking on this work, this is a way cleaner implementation of what I prototyped internally (re:  https://github.com/modelcontextprotocol/java-sdk/pull/297#issuecomment-3005229541).

I am curious:

1. why `McpTransportContextExtractor` didn't make it into `HttpServletSseServerTransportProvider`, and
2. are y'all opposed to a PR adding it in?

We're in a bit of a state where we can't quite switch to `HttpServletStreamableServerTransportProvider` yet, but need the `McpTransportContextExtractor` capability for Authn/z.

Would appreciate your input here, I'm happy to submit a PR to add it. Thanks!

stantonk:
@chemicL Created an issue and working PR to add `McpTransportContextExtractor` to `HttpServletSseServerTransportProvider`, would appreciate a review!

https://github.com/modelcontextprotocol/java-sdk/pull/477

chemicL:
@stantonk thanks! I reviewed your PR, let's continue the discussion there.

stantonk:
@chemicL curious, is the `HttpServletStatelessServerTransport` something non-standard in MCP? Wondering how to support it in clients other than the client from the SDK.

chemicL:
@stantonk the stateless variants of server transports are what actually justified the creation of Streamable HTTP in the spec. The problem with the previous remote transport using SSE meant that you always have a stateful, bidirectional communication which is hard to scale. The stateless transports are based around not supporting the [session](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management). They allow for a request-response model without the means for the server to freely ask the client for things like sampling, elicitation or perform random logging or send tools/resources/prompts updates notifications.

Each MCP client MUST support all the possibilities of the MCP servers - with sessions or without them, with simple json responses or a stream of responses. Here are relevant snippets from the spec:

```
If the input is a JSON-RPC request, the server MUST either return Content-Type: text/event-stream, to initiate an SSE stream, or Content-Type: application/json, to return one JSON object. The client MUST support both these cases.
```

```
If an Mcp-Session-Id is returned by the server during initialization, clients using the Streamable HTTP transport MUST include it in the Mcp-Session-Id header on all of their subsequent HTTP requests.
```

Issue #414, opened 2025-07-18
SSE Server drop client connection at 5 min.


albayrakonur:
Please do a quick search on GitHub issues first, there might be already a duplicate issue for the one you are about to create.
If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

**Bug description**
I developed a MCP server with webmvc SSE endpoint but when a client connects via /sse endpoint, after appx 5 min later server drops connection. This is happen because Client-side needs from server-side to ping itself. Current release doesnt support this and 

**Environment**
JDK:17
Spring Boot: 3.5.3
Spring AI: 1.0.0

**Steps to reproduce**
You can reproduce this bug on any setup. I try with Spring example of webmvc [https://github.com/spring-projects/spring-ai-examples/tree/main/model-context-protocol/weather/starter-webmvc-server](url) You can start this app and just connect it with cursor or vscode and then 5min later client will got an error.

Client-side servername: LOCAL-SERVER,
    "LOCAL-SERVER": {
      "url": "http://localhost:8080/sse"
    }

`2025-07-18 21:18:05.422 [info] user-LOCAL-SERVER: Handling CreateClient action
2025-07-18 21:18:05.422 [info] user-LOCAL-SERVER: Creating streamableHttp transport
2025-07-18 21:18:05.422 [info] user-LOCAL-SERVER: Connecting to streamableHttp server
2025-07-18 21:18:05.432 [error] user-LOCAL-SERVER: Client error for command Error POSTing to endpoint (HTTP 404): {"timestamp":"2025-07-18T18:18:05.431+00:00","status":404,"error":"Not Found","path":"/sse"}
2025-07-18 21:18:05.433 [info] user-LOCAL-SERVER: Client closed for command
2025-07-18 21:18:05.433 [error] user-LOCAL-SERVER: Error connecting to streamableHttp server, falling back to SSE: Error POSTing to endpoint (HTTP 404): {"timestamp":"2025-07-18T18:18:05.431+00:00","status":404,"error":"Not Found","path":"/sse"}
2025-07-18 21:18:05.433 [error] user-LOCAL-SERVER: Error connecting to streamableHttp server, falling back to SSE: Error POSTing to endpoint (HTTP 404): {"timestamp":"2025-07-18T18:18:05.431+00:00","status":404,"error":"Not Found","path":"/sse"}
[...]

Issue #158, opened 2025-04-14
Question: Do sse clients need to stay connected? If so, would it be better to specify it in the specification?


noear:
Currently, based on keepAlive, many http clients will experience read timeouts if the server does not issue sse (e.g. okhttp).

Implementations can have a convention, if defined by a specification.

For example:

Keep the server alive with an sse heartbeat, or set the read timeout to 0 on the client, and so on

xionghuyi:
I also want to know the answer

hongling-dev:
Keeping the SSE heartbeat does not completely solve the problem, as the mcp client cannot actively reconnect after the connection is unexpectedly disconnected (such as when the mcp server restarts).
When implementing it within the enterprise, I implemented an automatic reconnection mechanism (a temporary solution) to address this issue.

Additionally, `Streamable HTTP` can solve this problem in some scenarios.

noear:
@lambochen That's not necessarily... This is how solon-ai-mcp implements disconnection reconnections (e.g., server restarts). It has been tested and used in production.

https://github.com/opensolon/solon-ai/blob/dev/solon-ai-mcp/src/main/java/org/noear/solon/ai/mcp/client/McpClientToolProvider.java

```java
import io.modelcontextprotocol.client.McpClient;
import io.modelcontextprotocol.client.McpSyncClient;
import io.modelcontextprotocol.client.transport.ServerParameters;
import io.modelcontextprotocol.client.transport.StdioClientTransport;
import io.modelcontextprotocol.client.transport.WebRxSseClientTransport;
import io.modelcontextprotocol.spec.McpClientTransport;
import io.modelcontextprotocol.spec.McpSchema;
import org.noear.snack.ONode;
import org.noear.solon.Utils;
import org.noear.solon.ai.chat.tool.FunctionTool;
import org.noear.solon.ai.chat.tool.RefererFunctionTool;
import org.noear.solon.ai.chat.tool.ToolProvider;
import org.noear.solon.ai.image.Image;
import org.noear.solon.ai.mcp.McpChannel;
import org.noear.solon.ai.mcp.exception.McpException;
import org.noear.solon.core.Props;
import org.noear.solon.core.util.Assert;
import org.noear.solon.core.util.ResourceUtil;
import org.noear.solon.core.util.RunUtil;
import org.noear.solon.net.http.HttpTimeout;
import org.noear.solon.net.http.HttpUtilsBuilder;

import java.io.Closeable;
[...]

PR #430, opened 2025-07-30
feat: implement MCP-compliant keep-alive functionality for server transports

tzolov:
- Add KeepAliveScheduler utility class for configurable periodic session pings
- Integrate keep-alive support in WebFlux, WebMVC, and HttpServlet SSE transport providers
- Add keepAliveInterval configuration option to all transport provider builders
- Deprecate existing constructors in favor of builder pattern with enhanced configuration
- Update graceful shutdown to properly clean up keep-alive schedulers
- Add unit tests for KeepAliveScheduler functionality
- Configurable ping frequency to suit different network environments
- Optional keep-alive (disabled by default) to avoid excessive network overhead

Resolves: #414, #158
Replaces #353

## Motivation and Context

This change implements the [Model Context Protocol specification recommendations for connection health detection](https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/ping#implementation-considerations). The MCP spec states that implementations SHOULD periodically issue pings to detect connection health, with configurable frequency and appropriate timeouts for different network environments.

Without keep-alive functionality, long-running MCP sessions over SSE connections can experience idle timeouts, leading to unexpected connection drops and poor user experience. This implementation provides:

- Configurable periodic ping messages to maintain connection health
- Prevention of idle connection timeouts in proxy servers and load balancers
- Improved reliability for long-running MCP sessions
- Compliance with MCP specification guidelines

## How Has This Been Tested?

- **Unit Tests**: Comprehensive test suite added for `KeepAliveScheduler` covering:
- **Integration Testing**: Tested with existing SSE transport provider test suites to ensure backward compatibility

## Breaking Changes

**No breaking changes.** This is a backward-compatible enhancement:

## Types of changes

- [ ] Documentation update

## Checklist

### Design Decisions

- Keep-alive is **disabled by default** to avoid network overhead for applications that don't need it
- Uses MCP's standard `ping` method for compatibility
- Start the KeepAliveScheduler in provider's constructor if the keep-alive interval is provided. 
- Shutdown the KeepAliveScheduler in the `closeGracefully()` methods
- **Error Resilience**: Failed pings are logged but don't stop the scheduler

### Migration Path

Users can gradually adopt the new builder pattern:

```java
// Old (still works, but deprecated)
new WebFluxSseServerTransportProvider(objectMapper, messageEndpoint);

// New (recommended)
WebFluxSseServerTransportProvider.builder()
    .objectMapper(objectMapper)
    .messageEndpoint(messageEndpoint)
    .keepAliveInterval(Duration.ofSeconds(30))
    .build();

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/util/KeepAliveScheduler.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/util/KeepAliveScheduler.java
@@ -1,0 +1,216 @@
+/**
+ * Copyright 2025 - 2025 the original author or authors.
+ */
+package io.modelcontextprotocol.util;
+
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+
+import io.modelcontextprotocol.spec.McpSchema;
+import io.modelcontextprotocol.spec.McpSession;
+import reactor.core.Disposable;
+import reactor.core.publisher.Flux;
+import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
+
+/**
+ * A utility class for scheduling regular keep-alive calls to maintain connections. It
+ * sends periodic keep-alive, ping, messages to connected mcp clients to prevent idle
+ * timeouts.
+ *
+ * The pings are sent to all active mcp sessions at regular intervals.
+ *
+ * @author Christian Tzolov
+ */
+public class KeepAliveScheduler {
+
+	private static final Logger logger = LoggerFactory.getLogger(KeepAliveScheduler.class);
+
+	private static final TypeReference<Object> OBJECT_TYPE_REF = new TypeReference<>() {
+	};
+
+	/** Initial delay before the first keepAlive call */
+	private final Duration initialDelay;
+
+	/** Interval between subsequent keepAlive calls */
+	private final Duration interval;
+
+	/** The scheduler used for executing keepAlive calls */
+	private final Scheduler scheduler;
+
+	/** The current state of the scheduler */
+	private final AtomicBoolean isRunning = new AtomicBoolean(false);
+
+	/** The current subscription for the keepAlive calls */
+	private Disposable currentSubscription;
+
+	// TODO Currently we do not support the streams (streamable http session created by
+	// http post/get)
+
+	/** Supplier for reactive McpSession instances */
+	private final Supplier<Flux<McpSession>> mcpSessions;
+
+	/**
+	 * Creates a KeepAliveScheduler with a custom scheduler, initial delay, interval and a
+	 * supplier for McpSession instances.
+	 * @param scheduler The scheduler to use for executing keepAlive calls
+	 * @param initialDelay Initial delay before the first keepAlive call
+	 * @param interval Interval between subsequent keepAlive calls
+	 * @param mcpSessions Supplier for McpSession instances
+	 */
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by He-Pin** [comment]
Nice to have KeepAliveScheduler, thanks, otherwise the connection will be broken by middlebox。

Should it be `PingScheduler` instead of `KeepAliveScheduler`

File: mcp/src/main/java/io/modelcontextprotocol/util/KeepAliveScheduler.java [hunk 1]
```diff
@@ -0,0 +1,212 @@
...
+/**
+ * A utility class for scheduling regular keep-alive calls to maintain connections. It
+ * sends periodic keep-alive, ping, messages to connected mcp clients to prevent idle
+ * timeouts.
+ *
+ * The pings are sent to all active mcp sessions at regular intervals.
+ *
+ * @author Christian Tzolov
+ */
+public class KeepAliveScheduler {
```
He-Pin, line 31: `PingScheduler`?

mcp/src/main/java/io/modelcontextprotocol/util/KeepAliveScheduler.java
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	/**
+	 * Starts regular keepAlive calls with sessions supplier.
+	 * @return Disposable to control the scheduled execution
+	 */
+	public Disposable start() {
+		if (this.isRunning.compareAndSet(false, true)) {
+
+			this.currentSubscription = Flux.interval(this.initialDelay, this.interval, this.scheduler)
```
He-Pin, line 90: We should use maxIdle instead of this. If there are interactions between the servers and clients, then we do not need to ping it, in pekko-stream, there is an `keepAlive` operator for this. not sure about flux.

mcp/src/main/java/io/modelcontextprotocol/util/KeepAliveScheduler.java
tzolov: KeepAliveScheduler better explains the intention of this class.  
Ping is just one way to implement it

mcp/src/main/java/io/modelcontextprotocol/util/KeepAliveScheduler.java
tzolov: @He-Pin can you please elaborate on this?

chemicL: I'm not an expert on this, but potentially a more robust approach could be devised by hooking into the transport `FluxSink` and applying a `switchOnNext` with the result as `Flux.just(result)`, followed a `concat` of the `interval` operator. Each next item would first cancel the ongoing pinging sequence and would trigger a new one.

That would address @He-Pin concerns I believe, but would be a more complicated solution.

He-Pin: @tzolov If there is already an interaction happend between client and server, we should cancel the scheduled ping task, and reschedule a new one.

tzolov: @chemicL, would current approach do the job?

chemicL: Yes, as far as I see it, it's just some additional payload that could be avoided. All in all, this change only applies to the generic SSE stream for Streamable HTTP transports and the one-and-only SSE stream for the legacy SSE. I do understand @He-Pin's concern and we should come up with a more robust solution, but this is already a step in the right direction to provide something resolving an issue that idle clients face where they disconnect and can't make further progress unless they restart.

PR #448, opened 2025-08-05
Remove duplicate header MCP_PROTOCOL_VERSION

quaff:


tzolov:
Thank you @quaff, nice catch

PR #449, opened 2025-08-05
Use `Last-Event-ID` instead of `last-event-id`

quaff:
see https://html.spec.whatwg.org/multipage/server-sent-events.html#the-last-event-id-header

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/spec/HttpHeaders.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/spec/HttpHeaders.java
@@ -19,7 +19,7 @@ public interface HttpHeaders {
 	/**
 	 * Identifies events within an SSE Stream.
 	 */
-	String LAST_EVENT_ID = "last-event-id";
+	String LAST_EVENT_ID = "Last-Event-ID";
 
 	/**
 	 * Identifies the MCP protocol version.


chemicL:
https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1

> Field names are case-insensitive and ought to be registered within the "Hypertext Transfer Protocol (HTTP) Field Name Registry"

If this is causing any issues when the server provides lowercase or the client is unable to work with various cases then please open an issue with a reproducer so we can investigate and address it properly.

tzolov:
@chemicL , a good point but better to play "safe".  Most SDK implementations likely aren't case insensitive - Seems like the Java SDK is not either. 
> Correction: The Java SDK handles the case insensitive as expected as it relies on the Java/Spring API for header management.

quaff:
> https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1
> 
> > Field names are case-insensitive and ought to be registered within the "Hypertext Transfer Protocol (HTTP) Field Name Registry"
> 
> If this is causing any issues when the server provides lowercase or the client is unable to work with various cases then please open an issue with a reproducer so we can investigate and address it properly.

I'm not saying it will cause any issue, I think it's better to align with SSE spec and `MCP-Protocol-Version` in the same source file.

tzolov:
back-ported to 0.11.x

chemicL:
> @chemicL , a good point but better to play "safe".

I consider this to be opposite to "safe". Not treating headers as case-insensitive is a very risky business and should be fixed in the particular implementation on a case-by-case basis. Your understanding of "safe" seems to be rather that we wash our hands in case someone else's implementation is broken and would break for someone using it, e.g. behind an Envoy proxy which lowercases all headers. These things are difficult to investigate in a distributed environment.

chemicL:
To add to that - I share the views from the `Criticism` section of https://en.wikipedia.org/wiki/Robustness_principle

E.g.
> A flaw can become entrenched as a de facto standard. Any implementation of the protocol is required to replicate the aberrant behavior, or it is not interoperable.

PR #457, opened 2025-08-06
refactor: downgrade unhandled notification logging from error to warn

tzolov:
- Change logger.error() to logger.warn() for unhandled notification methods
- Log full notification object instead of just method name for better context
- Affects McpClientSession, McpServerSession, and McpStreamableServerSession


tzolov:
back-ported to 0.11.x

PR #463, opened 2025-08-06
ServerCapabilities builder should not assign logging an initial value

Randgalt:
## Motivation and Context

ServerCapabilities builder should not assign logging an initial value

## How Has This Been Tested?

added `assertThat(result.capabilities().logging()).isNull();` to `testSuccessfulInitialization`

## Breaking Changes

`ServerCapabilities` builder no longer has logging enabled by default.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -533,7 +533,7 @@ public final class McpSchema {
 
 			private Map<String, Object> experimental;
 
-			private LoggingCapabilities logging = new LoggingCapabilities();
+			private LoggingCapabilities logging;
 
 			private PromptCapabilities prompts;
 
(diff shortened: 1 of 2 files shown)


tzolov:
Thanks @Randgalt. Could you elaborate on this? Does it break existing clients? I don't see how this could be harmful. Also, some existing (deprecated) code expects the logging to be present: https://github.com/modelcontextprotocol/java-sdk/blob/4532b614e4d9431b8290f87dd2f20f2e5dd0f962/mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java#L767

tzolov:
I see one issue with logging enabled in the stateless server context: it will falsely claim logging capability but won't actually be able to send logs.

Randgalt:
> I see one issue with logging enabled in the stateless server context: it will falsely claim logging capability but won't actually be able to send logs.

There are two issues:

- The stateless server currently has no way of sending log messages. This would require SSE (BTW, I have some ideas to support this for progress messages)
- The spec for Logging from what I can tell requires sessions. This is because clients can send the server the log level that they will accept. Frankly, it would be better if the spec didn't have this or makes it optional, but it is there. 

So, the stateless server advertises that it supports logging but it doesn't.

Randgalt:
OK - I understand. I'll close this.

tzolov:
I will reopen the PR and apply the missing change while merging it

tzolov:
Thanks @Randgalt 
Rebased, extended and merged at 713ee1add0e29d184224aabdf06d024ef30a2754

PR #465, opened 2025-08-07
If a handler throws McpError, use its values for the RPC error

Randgalt:
Handlers should be able to throw RPC errors and `McpError` is the right exception for that.

Improve `DefaultMcpStatelessServerHandler` error handler to check if the exception is `McpError` and, if so, use it to build the RPC error result instead of re-writing as `INTERNAL_ERROR`.

## Motivation and Context

Handlers will want to send the client errors such as not found, etc.  The MCP spec allows for RPC errors to be returned.

## How Has This Been Tested?

New test, `testThrownMcpError()` in `HttpServletStatelessIntegrationTests`.

## Breaking Changes

none

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/server/DefaultMcpStatelessServerHandler.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/server/DefaultMcpStatelessServerHandler.java
@@ -35,9 +35,17 @@ class DefaultMcpStatelessServerHandler implements McpStatelessServerHandler {
 		}
 		return requestHandler.handle(transportContext, request.params())
 			.map(result -> new McpSchema.JSONRPCResponse(McpSchema.JSONRPC_VERSION, request.id(), result, null))
-			.onErrorResume(t -> Mono.just(new McpSchema.JSONRPCResponse(McpSchema.JSONRPC_VERSION, request.id(), null,
-					new McpSchema.JSONRPCResponse.JSONRPCError(McpSchema.ErrorCodes.INTERNAL_ERROR, t.getMessage(),
-							null))));
+			.onErrorResume(t -> {
+				McpSchema.JSONRPCResponse.JSONRPCError error;
+				if (t instanceof McpError mcpError && mcpError.getJsonRpcError() != null) {
+					error = mcpError.getJsonRpcError();
+				}
+				else {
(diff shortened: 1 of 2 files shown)


tzolov:
Thanks @Randgalt . I like this. 
Also we are working toward removing the [McpError(Object)](https://github.com/modelcontextprotocol/java-sdk/blob/4532b614e4d9431b8290f87dd2f20f2e5dd0f962/mcp/src/main/java/io/modelcontextprotocol/spec/McpError.java#L18) constructor (having it was a mistake), leaving only the `McpError(JSONRPCError jsonRpcError)`

mcp/src/main/java/io/modelcontextprotocol/server/DefaultMcpStatelessServerHandler.java
```diff
@@ -35,9 +35,17 @@ public Mono<McpSchema.JSONRPCResponse> handleRequest(McpTransportContext transpo
 		}
 		return requestHandler.handle(transportContext, request.params())
 			.map(result -> new McpSchema.JSONRPCResponse(McpSchema.JSONRPC_VERSION, request.id(), result, null))
-			.onErrorResume(t -> Mono.just(new McpSchema.JSONRPCResponse(McpSchema.JSONRPC_VERSION, request.id(), null,
-					new McpSchema.JSONRPCResponse.JSONRPCError(McpSchema.ErrorCodes.INTERNAL_ERROR, t.getMessage(),
-							null))));
+			.onErrorResume(t -> {
+				McpSchema.JSONRPCResponse.JSONRPCError error;
+				if (t instanceof McpError mcpError) {
+					error = mcpError.getJsonRpcError();
```
tzolov, line 41: Until we remove the [McpError(Object)](https://github.com/modelcontextprotocol/java-sdk/blob/4532b614e4d9431b8290f87dd2f20f2e5dd0f962/mcp/src/main/java/io/modelcontextprotocol/spec/McpError.java#L18) constructor we need to check that the getJsonRpcError() is not null otherwise create new JSONRCPError.

Randgalt: done

tzolov:
back-ported to 0.11.x

Issue #476, opened 2025-08-10
Support McpTransportContext in HttpServletSseServerTransportProvider


stantonk:
**Expected Behavior**

https://github.com/modelcontextprotocol/java-sdk/pull/420 introduced `McpTransportContext`.

`McpTransportContext` should be supported in `HttpServletSseServerTransportProvider` to enable those who are still on that transport to leverage `McpTransportContext` to set attributes that Tools, etc. have access to. This is crucial for implementing Authentication & Authorization at the Tool level.

**Current Behavior**

`McpTransportContext` is absent from `HttpServletSseServerTransportProvider`

**Context**

How has this issue affected you?

We're not able to switch to `HttpServletStreamableServerTransportProvider` yet, but wish to use `McpTransportContext` instead of our custom built similar capability.

What are you trying to accomplish?

Authn/z at Tool level.

What other alternatives have you considered?

Continuing to use our internal fork similar implementation.

Are you aware of any workarounds?

No

chemicL:
@stantonk this has not attracted any attention from other users, unfortunately. I don't think we'd prioritize this work at this point. But please do let us know whether it is still an issue for you so that we have the context and make an informed decision. It feels like the community has moved on from the legacy SSE.

stantonk:
@chemicL i filed a PR and it was already merged in some time ago :)

https://github.com/modelcontextprotocol/java-sdk/pull/477#issuecomment-3224986760

we do still support it internally by way of the SDK

chemicL:
Oh! In that case that's great news :) Let me close this issue then!

PR #477, opened 2025-08-10
Feature: McpTransportContext for HttpServletSseServerTransportProvider

stantonk:
Implements https://github.com/modelcontextprotocol/java-sdk/issues/476.

## Motivation and Context

https://github.com/modelcontextprotocol/java-sdk/pull/420 introduced McpTransportContext.

`McpTransportContext` should be supported in `HttpServletSseServerTransportProvider` to enable those who are still on that transport to leverage `McpTransportContext` to set attributes that Tools, etc. have access to. This is crucial for implementing Authentication & Authorization at the Tool level.

Current Behavior

`McpTransportContext` is absent from `HttpServletSseServerTransportProvider`

Context

How has this issue affected you?

We're not able to switch to `HttpServletStreamableServerTransportProvider` yet, but wish to use `McpTransportContext` instead of our custom built similar capability.

What are you trying to accomplish?

Authn/z at Tool level.

## How Has This Been Tested?

* In a langchain4j agent application calling into a Dropwizard application serving MCP using `HttpServletSseServerTransportProvider`
* directly by testing a Dropwizard application serving MCP using `HttpServletSseServerTransportProvider`
* Integration tests included with PR

## Breaking Changes
No, use of the `McpTransportContext` is optional. 

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
None

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
@@ -149,7 +149,7 @@ public class HttpServletSseServerTransportProvider extends HttpServlet implement
 	@Deprecated
 	public HttpServletSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
 			String sseEndpoint) {
-		this(objectMapper, baseUrl, messageEndpoint, sseEndpoint, null, null);
+		this(objectMapper, baseUrl, messageEndpoint, sseEndpoint, null, (serverRequest, context) -> context);
 	}
 
 	/**
@@ -168,7 +168,8 @@ public class HttpServletSseServerTransportProvider extends HttpServlet implement
 	@Deprecated
 	public HttpServletSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
 			String sseEndpoint, Duration keepAliveInterval) {
-		this(objectMapper, baseUrl, messageEndpoint, sseEndpoint, keepAliveInterval, null);
+		this(objectMapper, baseUrl, messageEndpoint, sseEndpoint, keepAliveInterval,
+				(serverRequest, context) -> context);
 	}
 
 	/**
@@ -180,7 +181,7 @@ public class HttpServletSseServerTransportProvider extends HttpServlet implement
 	 * @param messageEndpoint The endpoint path where clients will send their messages
 	 * @param sseEndpoint The endpoint path where clients will establish SSE connections
 	 * @param keepAliveInterval The interval for keep-alive pings, or null to disable
-     * keep-alive functionality
+	 * keep-alive functionality
 	 * @param contextExtractor The extractor for transport context from the request.
 	 * @deprecated Use the builder {@link #builder()} instead for better configuration
 	 * options.
@@ -188,6 +189,11 @@ public class HttpServletSseServerTransportProvider extends HttpServlet implement
 	private HttpServletSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
 			String sseEndpoint, Duration keepAliveInterval,
 			McpTransportContextExtractor<HttpServletRequest> contextExtractor) {
+
+		Assert.notNull(objectMapper, "ObjectMapper must not be null");
+		Assert.notNull(messageEndpoint, "messageEndpoint must not be null");
+		Assert.notNull(sseEndpoint, "sseEndpoint must not be null");
+		Assert.notNull(contextExtractor, "Context extractor must not be null");
 
 		this.objectMapper = objectMapper;
 		this.baseUrl = baseUrl;
--- a/mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
@@ -185,8 +185,7 @@ public class HttpServletSseServerTransportProvider extends HttpServlet implement
 	 * @deprecated Use the builder {@link #builder()} instead for better configuration
 	 * options.
 	 */
-	@Deprecated
-	public HttpServletSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
+	private HttpServletSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
 			String sseEndpoint, Duration keepAliveInterval,
 			McpTransportContextExtractor<HttpServletRequest> contextExtractor) {
 
--- a/mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
@@ -371,7 +371,7 @@ public class HttpServletSseServerTransportProvider extends HttpServlet implement
 				body.append(line);
 			}
 
-			final McpTransportContext transportContext = contextExtractor.extract(request,
+			final McpTransportContext transportContext = this.contextExtractor.extract(request,
 					new DefaultMcpTransportContext());
 			McpSchema.JSONRPCMessage message = McpSchema.deserializeJsonRpcMessage(objectMapper, body.toString());
 
--- a/mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
@@ -16,6 +16,9 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.modelcontextprotocol.server.DefaultMcpTransportContext;
+import io.modelcontextprotocol.server.McpTransportContext;
+import io.modelcontextprotocol.server.McpTransportContextExtractor;
 import io.modelcontextprotocol.spec.McpError;
 import io.modelcontextprotocol.spec.McpSchema;
 import io.modelcontextprotocol.spec.McpServerSession;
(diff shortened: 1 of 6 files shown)


mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
```diff
@@ -163,11 +168,33 @@ public HttpServletSseServerTransportProvider(ObjectMapper objectMapper, String b
...
+	 * @param baseUrl The base URL for the server transport
+	 * @param messageEndpoint The endpoint path where clients will send their messages
+	 * @param sseEndpoint The endpoint path where clients will establish SSE connections
+	 * @param keepAliveInterval The interval for keep-alive pings, or null to disable
+     * keep-alive functionality
+	 * @param contextExtractor The extractor for transport context from the request.
+	 * @deprecated Use the builder {@link #builder()} instead for better configuration
+	 * options.
+	 */
+	@Deprecated
```
chemicL: Instead of adding a deprecated public constructor, let's just have a private one perhaps?

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
```diff
@@ -144,7 +149,7 @@ public HttpServletSseServerTransportProvider(ObjectMapper objectMapper, String m
 	@Deprecated
 	public HttpServletSseServerTransportProvider(ObjectMapper objectMapper, String baseUrl, String messageEndpoint,
 			String sseEndpoint) {
-		this(objectMapper, baseUrl, messageEndpoint, sseEndpoint, null);
+		this(objectMapper, baseUrl, messageEndpoint, sseEndpoint, null, null);
```
chemicL: Better to provide a no-op impl rather than null.

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
```diff
@@ -339,10 +366,13 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
 				body.append(line);
 			}
 
+			final McpTransportContext transportContext = contextExtractor.extract(request,
```
chemicL: In the current situation, `contextExtractor` can be `null` since the deprecated constructor sets the field to `null`. Let's avoid that. Also, it's best to refer to instance fields via `this.` prefix.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -266,7 +270,11 @@ private Mono<McpSchema.JSONRPCResponse> handleIncomingRequest(McpSchema.JSONRPCR
 									error.message(), error.data())));
 				}
 
-				resultMono = this.exchangeSink.asMono().flatMap(exchange -> handler.handle(exchange, request.params()));
+				resultMono = this.exchangeSink.asMono().flatMap(exchange -> {
+					McpAsyncServerExchange newExchange = new McpAsyncServerExchange(exchange.sessionId(), this,
```
chemicL, line 280: Interesting! I like the fact that it gets the job done, but I would kindly ask for providing a comment above (and in `handleIncomingRequest` too) that explains the reasoning, which is that this legacy implementation assumes an exchange is established upon the initialization phase completion and then creates the exchange object that is cached. Now we are creating a new one for each HTTP request and we fill it with the extracted transport context.

mcp/src/test/java/io/modelcontextprotocol/server/AbstractMcpClientServerIntegrationTests.java
```diff
@@ -825,6 +827,61 @@ void testThrowingToolCallIsCaughtBeforeTimeout(String clientType) {
...
+
+		var expectedCallResponse = new McpSchema.CallToolResult(
+				List.of(new McpSchema.TextContent("CALL RESPONSE; ctx=value")), null);
+		McpServerFeatures.SyncToolSpecification tool1 = McpServerFeatures.SyncToolSpecification.builder()
+			.tool(Tool.builder().name("tool1").description("tool1 description").inputSchema(emptyJsonSchema).build())
+			.callHandler((exchange, request) -> {
+
+				McpTransportContext transportContext = exchange.transportContext();
+				assertTrue(transportContext != null, "transportContext should not be null");
+				assertTrue(!transportContext.equals(McpTransportContext.EMPTY), "transportContext should not be empty");
```
chemicL: In reality these assertions might be executed on a Thread that is not a JUnit thread and would not cause the test to fail. It's best to set some Atomic* instances to the observed values and perform assertions after the execution of the intended flow.

mcp/src/test/java/io/modelcontextprotocol/server/AbstractMcpClientServerIntegrationTests.java
```diff
@@ -1531,4 +1588,9 @@ private double evaluateExpression(String expression) {
 		};
 	}
 
+	protected static McpTransportContextExtractor<HttpServletRequest> extractor = (r, tc) -> {
```
chemicL:
```suggestion
	protected static McpTransportContextExtractor<HttpServletRequest> TEST_CONTEXT_EXTRACTOR = (r, tc) -> {
```

stantonk:
@chemicL thanks for the quick review! All your suggestions make sense, I'll get those updates made.

RE: Spring support, I've never used Spring but I can take a pass at it, probably this weekend. If I hit some snags would you be open to merging the Servlet version in first? I just don't want to stay on our internal fork for too long as additional capabilities and bug fixes get merged in.

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
stantonk: good call!

mcp/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
stantonk: `null` should no longer be possible, all constructor pathways now have a no-op `McpTransportContextExtractor` of `(serverRequest, context) -> context`

mcp/src/test/java/io/modelcontextprotocol/server/AbstractMcpClientServerIntegrationTests.java
stantonk: ah interesting, did not know that! good catch. Updated my test, and also pulled the responseBody assertion into an `AtomicBoolean` as well in both the added test, and the original `testToolCallSuccess`.

stantonk:
@chemicL thanks again for the review!

## Updates

* I've made all your recommended updates

## Testing

* I re-ran `mvn clean verify` locally and all tests passed
* I ran `mvn clean install` locally, and configured our Dropwizard MCP Service to use the latest HEAD of this PR after all your requested changes. I confirmed our AI Agent service talking to our Dropwizard MCP Service is functioning properly with the latest changes, including our Authentication & Authorization into the individual MCP Tool call level via the `McpTransportContextExtractor` using `HttpServletSseServerTransportProvider`. Initialization, list tools, and tool calls are working properly.

### Spring WebMvc and WebFlux

I took a quick look through the code, it all looks pretty similar so I should be able to follow up in a subsequent PR adding `McpTransportContext` to those as well. I've got a busy week coming up, but I'll try for next weekend.

mcp/src/main/java/io/modelcontextprotocol/spec/McpServerSession.java
```diff
@@ -280,24 +294,36 @@ private Mono<McpSchema.JSONRPCResponse> handleIncomingRequest(McpSchema.JSONRPCR
...
-			return this.exchangeSink.asMono().flatMap(exchange -> handler.handle(exchange, notification.params()));
+			return this.exchangeSink.asMono().flatMap(exchange -> {
+				// This legacy implementation assumes an exchange is established upon
+				// the initialization phase see: exchangeSink.tryEmitValue(...),
+				// which creates a cached immutable exchange.
+				// Here, we create a new exchange and copy over everything from that
+				// cached exchange, and use it for a single HTTP request, with the
+				// transport context passed in.
+				McpAsyncServerExchange newExchange = new McpAsyncServerExchange(exchange.sessionId(), this,
+						exchange.getClientCapabilities(), exchange.getClientInfo(), transportContext);
```
chemicL, line 324: This can be extracted to a utility method to avoid the copy of the comment.

mcp/src/test/java/io/modelcontextprotocol/server/AbstractMcpClientServerIntegrationTests.java
```diff
@@ -825,6 +831,68 @@ void testThrowingToolCallIsCaughtBeforeTimeout(String clientType) {
...
+				McpTransportContext transportContext = exchange.transportContext();
+				transportContextIsNull.set(transportContext == null);
+				transportContextIsEmpty.set(transportContext.equals(McpTransportContext.EMPTY));
+				String ctxValue = (String) transportContext.get("important");
+
+				try {
+					HttpResponse<String> response = HttpClient.newHttpClient()
+						.send(HttpRequest.newBuilder()
+							.uri(URI.create(
+									"https://raw.githubusercontent.com/modelcontextprotocol/java-sdk/refs/heads/main/README.md"))
```
chemicL, line 859: I suppose there is no need to call an external API in this test. I'd say we don't need to call anything, just return the contents of the context from the tool.

mcp/src/test/java/io/modelcontextprotocol/server/AbstractMcpClientServerIntegrationTests.java
```diff
@@ -1531,4 +1599,9 @@ private double evaluateExpression(String expression) {
 		};
 	}
 
+	protected static McpTransportContextExtractor<HttpServletRequest> TEST_CONTEXT_EXTRACTOR = (r, tc) -> {
+		tc.put("important", "value");
+		return tc;
+	};
```
chemicL, line 1605: We could have a copy of this in both `HttpServletSseIntegrationTests` and `HttpServletStreamableIntegrationTests` to avoid a reference to a servlet type in the base class.

BTW, I thought this abstract test can be used outside of the servlet context, but it seems the variant in mcp-test module is not really a copy.. @tzolov do we have a strategy for aligning these? There's no comment like we have in other files of this sort (`// KEEP IN SYNC with the class in mcp-test module`)...

tzolov:
rebased, extended, squashed and merged at b4fef5241fa81dda87138c98566014a090250f06

Additionally:
- Port the feature to the WebMvcSseServerTransportProvider and WebFluxSseServerTransportProvider
- Align the AbstractStatelessIntegrationTests in the mcp and mcp-test modules.
- Fix an context propagation issue in the WebFluxStreamableServerTransportProvider
- minor code/test cleaning and improvements
- backported to 0.11.x

stantonk:
@tzolov @chemicL thanks for all your help on this! @tzolov appreciate you porting to `WebMvcSseServerTransportProvider` and `WebFluxSseServerTransportProvider`!

PR #489, opened 2025-08-14
fix: When call the MCP SSE endpoint, the Content-Type header not passed by default

zhangjingcn:
## Motivation and Context

When calling the MCP SSE endpoint in the current SDK, the Content-Type: application/json header is included by default. However, the SSE endpoint is accessed via a GET request, and according to HTTP specifications, the Content-Type header should not be sent with GET requests. This has caused some MCP servers to fail validation or handle the request incorrectly. Therefore, the HttpClientSseClientTransport class has been modified so that the default request builder no longer includes the Content-Type header. The Content-Type header is now only included by default for POST requests.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
@@ -247,8 +247,7 @@ public class HttpClientSseClientTransport implements McpClientTransport {
 
 		private ObjectMapper objectMapper = new ObjectMapper();
 
-		private HttpRequest.Builder requestBuilder = HttpRequest.newBuilder()
-			.header("Content-Type", "application/json");
+		private HttpRequest.Builder requestBuilder = HttpRequest.newBuilder();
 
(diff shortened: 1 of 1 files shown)


tzolov:
@zhangjingcn, thanks for raising this.
I agree that GET doesn’t require a Content-Type header, but the [HTTP spec doesn’t prohibit it](https://stackoverflow.com/questions/5661596/do-i-need-a-content-type-header-for-http-get-requests). 
Could you clarify what you mean by “some MCP servers fail validation or handle the request incorrectly”? Which servers misbehave?

zhangjingcn:
> @zhangjingcn, thanks for raising this. I agree that GET doesn’t require a Content-Type header, but the [HTTP spec doesn’t prohibit it](https://stackoverflow.com/questions/5661596/do-i-need-a-content-type-header-for-http-get-requests). Could you clarify what you mean by “some MCP servers fail validation or handle the request incorrectly”? Which servers misbehave?

@tzolov Thank you for your reply. Our MCP server is hosted on Higress (https://higress.ai/en/), which provides features such as service management and authentication. I consulted with the Higress development team, and they explained that when Higress, acting as a gateway, receives a request containing a Content-Type header, it assumes there is a request body and attempts to read it. However, since GET requests do not actually include a body, this causes Higress to hang during processing.

Issue #471, opened 2025-08-08
Streamable session clean up


He-Pin:
The current release doesn't contain any remove / clean up/ idle clean up logic, if the client never calls `DELETE` , then OOM.

taobaorun:
what about to fix this by server ping
https://github.com/modelcontextprotocol/java-sdk/pull/508

chemicL:
This is indeed a serious problem that requires addressing. Let's discuss ideas based on the comments from #508.

This can be some form of TTL and recency markers (update timestamp upon every interaction with a given session) and a scheduler task that checks that against a globally set TTL perhaps?

In the long run, this should be managed by a proper session storage, so I marked this issue with `feature/session-storage` label to keep in mind for the bigger theme.

bperhaps:
I can’t think of a better approach than using TTL. Wouldn’t it be quite reasonable to simply add a feature that keeps the session alive as long as the GET SSE connection is open? If we make TTL injectable, developers could configure it according to their business needs.

On a more personal note, once SSE gets involved in streamable-http (even if it’s not a GET SSE request, such as a sampling request), the session inevitably becomes more tightly coupled to a specific server. In that case, I feel like the need for a global session storage diminishes. Is it really necessary to consider introducing a new session store? Wouldn’t it be better to keep the current architecture and just focus on handling session cleanup properly? I’m curious about your thoughts.

jonnybbb:
We hit exactly this running a public Streamable HTTP MCP server (Spring AI 2.0.0 / `mcp-spring-webmvc` with SDK 2.0.0), and can add some production evidence plus a concrete proposal.

**Evidence**

- `WebMvcStreamableServerTransportProvider` puts a session into its map on every `initialize` and removes it only on an explicit client `DELETE` or `closeGracefully()`. In practice most MCP clients never send `DELETE`, so sessions accumulate until restart even with perfectly well-behaved traffic.
- Configuring `keep-alive-interval` does not help: `KeepAliveScheduler` logs a WARN on ping failure and continues (`doOnError` → log, `onErrorComplete`) — dead sessions are never removed (see #1022, and #1021 for the disconnect side).
- Because `initialize` is unauthenticated by protocol design, this is not just resource creep — an anonymous client can grow server memory without bound. On a public server this is a slow, unauthenticated memory-exhaustion vector, not only an OOM-after-weeks nuisance.

[...]

PR #508, opened 2025-08-21
try fix session potential leak by server sse ping

taobaorun:
## Motivation and Context

Streamable Potential session leaks when Client Fails to Invoke DELETE

Problem:

The current implementation can leak server-side sessions under the following conditions:

a. Session accumulation due to missing DELETE calls

When a client repeatedly establishes new connections but never issues the required DELETE request, the sessions map inside WebMvcStreamableServerTransportProvider grows unbounded.

i. Explicit DELETE is mandatory
Streamable clients must call the DELETE method when they are destroyed to ensure the server cleans up the corresponding session.
b. Client restarts prevent DELETE execution
If the client process (or container) is restarted before it has a chance to send the DELETE, the server retains the now-orphaned session indefinitely, leading to a rapid accumulation of stale entries.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcStreamableServerTransportProvider.java
+++ b/mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcStreamableServerTransportProvider.java
@@ -295,6 +295,13 @@ public class WebMvcStreamableServerTransportProvider implements McpStreamableSer
 					sseBuilder.onComplete(() -> {
 						logger.debug("SSE connection completed for session: {}", sessionId);
 						listeningStream.close();
+						// If the sse connection is closed,it means that the client
+						// connection has been disconnected
+						// due to network issues, etc.,
+						// and the session needs to be removed
+						if (sessionId != null) {
+							sessions.remove(sessionId);
+						}
 					});
 				}
 			}, Duration.ZERO);


taobaorun:
> According to [the specification](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management), sessions are rather permanent (outlive the client connection) and removing sessions should happen explicitly via a DELETE call by the client. The specification does not explain when sessions are removed so perhaps we could add a mechanism for the server to decide, e.g. introduce a TTL for inactive sessions. However, client disconnects should not trigger session removal unless explicitly configured that way (another strategy outside of TTL?). I'd suggest we discard this contribution and open a discussion in another issue, WDYT @tzolov ?

I think we need ttl

taobaorun:
https://github.com/modelcontextprotocol/java-sdk/blob/6ec443c89a7d3d122a4ff74ddbb67bb7b53c477e/mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcSseServerTransportProvider.java#L313

Here, with reference to the SSE implementation, cleanup is triggered when the SSE connection ends.

Damore:
Also, this problem affects the [WebFluxStreamableServerTransportProvider](https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java) as well. Here we can clearly see this leak happening on the logs, due to the [KeepAliverScheduler](https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp/src/main/java/io/modelcontextprotocol/util/KeepAliveScheduler.java#L95)

Screenshot 2025-09-10 at 16 58 51


The @taobaorun's suggestion of having a TTL sounds interesting, but what also seems a good choice, is to have a max number of keep-alive failure and, after that, we remove the session on the server.

Finally, one thing that calls my attention is that we have the method `handleGet` on the [sink.onDispose](https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java#L205). However, when the client disconects, it never invokes this flow. 

On the other hand, it correctly calls on the class [WebFluxSseServerTransportProvider](https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java) on the method `handleSseConnection`on the [sink.onCancel](https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java#L365-L368)

taobaorun:
> have a max number of keep-alive failure and, after that, we remove the session on the server.

I think it's a simple and effective solution. Just like this pr, it also attempts to solve the problem through ping.

Damore:
@chemicL Waiting for your feedback on my [comment](https://github.com/modelcontextprotocol/java-sdk/pull/508#issuecomment-3276320390) as well

ikauwsp:
facing this too. for long-lasting client which ping the server by initializing sessions every fixed interval, the sessions are leaking and grow unbounded which causes OOM (client wont explicitly sends DELETE to close sessions but server should be managing too)

ShemTovYosef:
Hi @tzolov , can you take a look on this PR, very critical fix for production system?

chemicL:
Hey, I agree this needs addressing but the solution proposed by this PR is in my view incorrect. The GET SSE stream is optional for Streamable HTTP while for legacy SSE it was the only means to communicate responses from the server to the client. That makes it all so hard with Streamable HTTP to properly address this. The GET SSE stream being closed does not signify the session can be removed. Let's get back to #471, discuss a potential solution (TTL + recency markers?) and once a solution is agreed upon, a contribution can be made.

PR #521, opened 2025-08-27
Stateless server transport shouldn't require TEXT_EVENT_STREAM

quaff:
## Motivation and Context
It's a trivial improvement.

## How Has This Been Tested?
Tested with curl.

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist


quaff:
@chemicL Could you review this? I confirm the failed test is not introduced by this PR since I can't reproduce it locally.

chemicL:
> @chemicL Could you review this? I confirm the failed test is not introduced by this PR since I can't reproduce it locally.

Hey, as I understand the spec, this is violating it:

> The client MUST include an Accept header, listing both application/json and text/event-stream as supported content types.

I do understand why you'd want to take away this strictness, but it's against another assumption behind the spec, which is that servers can be simple but clients have to support all kinds of servers, therefore this step is validating that a client is a proper one and if the server decides to return a stream it will be able to.

quaff:
Thanks for your explanation.

Issue #479, opened 2025-08-11
TransportContext for Client side


olamy:
Thanks for adding TransportContext on server-side, a similar feature is missing for client side as well.
My plan is to use a single McpClientTransport assuming it's thread safe and can be re use for diferent authz context.
The only way to customize http headers (needed for authz or whatever other http headers) is to add some value in `_meta` such:

```
  McpSchema.CallToolRequest request = McpSchema.CallToolRequest.builder()
          .name("test_add")
          .arguments(args)
          .meta(Map.of("Authorization", "really complicated password"))
          .build();

  McpSchema.CallToolResult result = mcpSyncClient.callTool(request);
```
Then having a `SyncHttpRequestCustomizer`

```
  public static SyncHttpRequestCustomizer getRequestCustomizer() {
      return (builder, method, endpoint, body) -> {
          try {
              if (body != null) {
                  McpSchema.JSONRPCMessage jsonrpcMessage = McpSchema.deserializeJsonRpcMessage(OBJECT_MAPPER, body);
                  if (jsonrpcMessage instanceof McpSchema.JSONRPCRequest jsonrpcRequest
                          && jsonrpcRequest.params() instanceof Map params) {
                      Optional<Map<String, Object>> meta = Optional.ofNullable((Map) params.get("_meta"));
                      Optional<String> authz = Optional.ofNullable(
                              (String) meta.orElse(Collections.emptyMap()).get(AUTHORIZATION_HEADER));
                      authz.ifPresent(s -> builder.header(AUTHORIZATION_HEADER, s));
                  }
              }
          } catch (Throwable e) {
              throw new RuntimeException(e);
          }
      };
  }
```

it works but looks brittle as the headers are part of the body and it's not possible to customize header with webflux such `WebClientStreamableHttpTransport`

I was thinking of something such 
```
McpSchema.CallToolResult result = mcpSyncClient.callTool(request, transportContext);

```
transportContext could be very similar to the one used for server part.

and changing the SyncHttpRequestCustomizer accordingly to accept this new arg? but what about `WebClientStreamableHttpTransport`. 
I'm happy to help here if you tell me if you have any preference for implementing this.

imod:
@olamy long time no see :) 

I'm interested in the exact same - the need to pass the enduser JWT token which is used by the MCP Server to call an other service in the name of the user.

unfortunate something like this does NOT work with SSE: 

```
            val timeSupplier: () -> String = { System.currentTimeMillis().toString() }

[...]

PR #522, opened 2025-08-27
Add McpTransportContext to McpSyncClient

Kehrlann:
This PR adds `McpTransportContext` to MCP Clients, and makes it available to `(Async|Sync)HttpRequestCustomizer`s. 

## Motivation and Context

The motivation is twofold:
- It creates a unified API for reading the "context" in which an MCP Client request is issued, regardless of sync/async. It can be consumed by HttpClient-based transports through `HttpRequestCustomizer`, or through the reactor context in WebClient-based transports.
- It also addresses limitations with thread local usage in request customizers, see #479

## How Has This Been Tested?

Test with Spring AI + Spring Security for passing JWT tokens to MCP Sync clients.

## Breaking Changes

Following the move of `(Async|Sync)HttpRequestCustomizer` to `Mcp(Async|Sync)HttpRequestCustomizer`, itself a breaking change, I also changed the signatures of those classes.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/client/McpClient.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/McpClient.java
@@ -11,10 +11,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
+import io.modelcontextprotocol.server.McpTransportContext;
 import io.modelcontextprotocol.spec.McpClientTransport;
 import io.modelcontextprotocol.spec.McpSchema;
-import io.modelcontextprotocol.spec.McpTransport;
 import io.modelcontextprotocol.spec.McpSchema.ClientCapabilities;
 import io.modelcontextprotocol.spec.McpSchema.CreateMessageRequest;
 import io.modelcontextprotocol.spec.McpSchema.CreateMessageResult;
@@ -22,6 +23,7 @@ import io.modelcontextprotocol.spec.McpSchema.ElicitRequest;
 import io.modelcontextprotocol.spec.McpSchema.ElicitResult;
 import io.modelcontextprotocol.spec.McpSchema.Implementation;
 import io.modelcontextprotocol.spec.McpSchema.Root;
+import io.modelcontextprotocol.spec.McpTransport;
 import io.modelcontextprotocol.util.Assert;
 import reactor.core.publisher.Mono;
 
@@ -182,6 +184,8 @@ public interface McpClient {
 		private Function<CreateMessageRequest, CreateMessageResult> samplingHandler;
 
 		private Function<ElicitRequest, ElicitResult> elicitationHandler;
+
+		private Supplier<McpTransportContext> contextProvider = () -> McpTransportContext.EMPTY;
 
 		private SyncSpec(McpClientTransport transport) {
 			Assert.notNull(transport, "Transport must not be null");
@@ -410,6 +414,22 @@ public interface McpClient {
 		}
 
 		/**
+		 * Add a provider of {@link McpTransportContext}, providing a context before
+		 * calling any client operation. This allows to extract thread-locals and hand
+		 * them over to the underlying transport.
+		 * <p>
+		 * There is no direct equivalent in {@link AsyncSpec}. To achieve the same result,
+		 * append {@code contextWrite(McpTransportContext.KEY, context)} to any
+		 * {@link McpAsyncClient} call.
+		 * @param contextProvider A supplier to create a context
+		 * @return This builder for method chaining
+		 */
+		public SyncSpec transportContextProvider(Supplier<McpTransportContext> contextProvider) {
+			this.contextProvider = contextProvider;
+			return this;
+		}
+
+		/**
 		 * Create an instance of {@link McpSyncClient} with the provided configurations or
 		 * sensible defaults.
 		 * @return a new instance of {@link McpSyncClient}.
@@ -423,7 +443,8 @@ public interface McpClient {
 			McpClientFeatures.Async asyncFeatures = McpClientFeatures.Async.fromSync(syncFeatures);
 
 			return new McpSyncClient(
-					new McpAsyncClient(transport, this.requestTimeout, this.initializationTimeout, asyncFeatures));
+					new McpAsyncClient(transport, this.requestTimeout, this.initializationTimeout, asyncFeatures),
+					this.contextProvider);
 		}
 
 	}
--- a/mcp/src/main/java/io/modelcontextprotocol/client/McpClient.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/McpClient.java
@@ -13,7 +13,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
-import io.modelcontextprotocol.server.McpTransportContext;
+import io.modelcontextprotocol.common.McpTransportContext;
 import io.modelcontextprotocol.spec.McpClientTransport;
 import io.modelcontextprotocol.spec.McpSchema;
 import io.modelcontextprotocol.spec.McpSchema.ClientCapabilities;
--- a/mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
@@ -5,16 +5,19 @@
 package io.modelcontextprotocol.client;
 
 import java.time.Duration;
+import java.util.function.Supplier;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import io.modelcontextprotocol.server.McpTransportContext;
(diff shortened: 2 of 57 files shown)


mcp/src/main/java/io/modelcontextprotocol/client/McpClient.java
```diff
@@ -11,17 +11,19 @@
 import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
+import io.modelcontextprotocol.server.McpTransportContext;
```
chemicL: Let's move `McpTransportContext` up a level into a new `common` package. It will be a breaking change, but it's a super recent type.

mcp/src/main/java/io/modelcontextprotocol/client/McpClient.java
```diff
@@ -183,6 +185,8 @@ class SyncSpec {
 
 		private Function<ElicitRequest, ElicitResult> elicitationHandler;
 
+		private Supplier<McpTransportContext> contextProvider = McpTransportContext.EMPTY::copy;
```
chemicL: We don't need to copy here. We should ensure that the copy is made whenever there is an intention for the user to modify the context, e.g. when creating an empty one that is being filled in and used as the provided result.

A couple of ideas:
* Hide `DefaultMcpTransportContext` (currently it's public, let's make it package private)
* Add factory methods to `McpTransportContext`: `empty()` which returns the immutable instance; and `create()` which creates an empty but mutable container.

chemicL: Another note on the `McpTransportContext`:

We use it already on the server side, where the extractor expects a mutable instance that it can fill in.
With the client we have the opposite where the provider creates an instance and expects to mutate it.

On the read side, however we have the user handlers in server specification that are only supposed to read contents and we should ensure that mutations are not allowed because there is no use of mutating the contents.

On the client end, the customizer is the reader also so it should be disallowed to write something to the context.

Therefore, perhaps let's break the API a little bit:

1. `McpTransportContext#put` is removed! The `McpTransportContext` is just a container of immutable context
2. `McpTransportExtractor` - deprecate `McpTransportContext extract(T request, McpTransportContext transportContext)` or even remove it (BREAKING) and just have `McpTransportContext extract(T request)` that allows to call `McpTransportContext.create(Map)` which would accept a `Map` that is already populated and now the `McpTransportContext` is only the view.
3. Client's transport context provider is also creating an instance like the extractor.
4. Customizer is just supposed to read values using `get(K)`.

mcp/src/main/java/io/modelcontextprotocol/client/transport/customizer/McpSyncHttpRequestCustomizer.java
```diff
@@ -6,16 +6,24 @@
...
 /**
  * Customize {@link HttpRequest.Builder} before executing the request, either in SSE or
- * Streamable HTTP transport.
+ * Streamable HTTP transport. Do not rely on thread-locals in this implementation, instead
+ * use {@link SyncSpec#transportContextProvider} to extract context, and then consume it
+ * through {@link McpTransportContext}.
  *
  * @author Daniel Garnier-Moiroux
  */
 public interface McpSyncHttpRequestCustomizer {
```
chemicL: Why not rename it to `McpSyncHttpClientRequestCustomizer` to reflect that it is tied to JDK's `HttpClient`.

mcp/src/main/java/io/modelcontextprotocol/client/McpClient.java
```diff
@@ -409,6 +413,18 @@ public SyncSpec progressConsumers(List<Consumer<McpSchema.ProgressNotification>>
...
 		}
 
+		/**
+		 * Add a provider of {@link McpTransportContext}, providing a context before
+		 * calling any client operation. This allows to extract thread-locals and hand
+		 * them over to the underlying transport.
+		 * @param contextProvider A supplier to create a context
+		 * @return This builder for method chaining
+		 */
+		public SyncSpec transportContextProvider(Supplier<McpTransportContext> contextProvider) {
```
chemicL, line 427: Let's perhaps add a comment so users are not too confused by the lack of this method on the `AsyncSpec` to clarify that the means to achieve the same is to simply append a `contextWrite(McpTransportContext.KEY, context)` to any `McpAsyncClient` API call.

mcp/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java
```diff
@@ -177,36 +185,43 @@ public boolean closeGracefully() {
 	public McpSchema.InitializeResult initialize() {
 		// TODO: block takes no argument here as we assume the async client is
 		// configured with a requestTimeout at all times
-		return this.delegate.initialize().block();
+		var context = this.contextProvider.get();
+		return this.delegate.initialize().contextWrite(ctx -> ctx.put(McpTransportContext.KEY, context)).block();
```
chemicL: How about adding a method like `Mono<?> withProvidedContext(...)` that wraps all the calls and attaches `.contextWrite(ctx -> ctx.put(McpTransportContext.KEY, this.contextProvider.get())`

mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
```diff
@@ -247,7 +250,9 @@ private Mono<Disposable> reconnect(McpTransportStream<Disposable> stream) {
 					.header("Cache-Control", "no-cache")
 					.header(HttpHeaders.PROTOCOL_VERSION, MCP_PROTOCOL_VERSION)
 					.GET();
-				return Mono.from(this.httpRequestCustomizer.customize(builder, "GET", uri, null));
+				var transportContext = ctx.getOrDefault(McpTransportContext.KEY, McpTransportContext.EMPTY);
```
chemicL: Let's perhaps use the innermost `Context` to be super certain that if at any point we alter the inner chain we do still grab the correct reference here. Replacing `Mono.defer` on line 236 with `Mono.deferContextual(connectionCtx -> {` should do the trick.

mcp/src/test/java/io/modelcontextprotocol/client/HttpClientStreamableHttpSyncClientTests.java
```diff
@@ -41,4 +56,20 @@ public void onClose() {
...
 
+	@Test
+	void customizesRequests() {
+		var mcpTransportContext = McpTransportContext.EMPTY.copy();
+		mcpTransportContext.put("some-key", "some-value");
+		withClient(createMcpTransport(), syncSpec -> syncSpec.transportContextProvider(() -> mcpTransportContext),
+				mcpSyncClient -> {
+					mcpSyncClient.initialize();
+
+					var transportCaptor = ArgumentCaptor.forClass(McpTransportContext.class);
```
chemicL: Hmm, perhaps we can add equality overrides to `DefaultMcpTransportContext`? :)

Kehrlann:
@chemicL @tzolov I have applied the requested changes.

tzolov:
Great work @Kehrlann !

imod:
Hey everyone - this is really great! thanks a lot for this. Will you also update the documentation - also together with the spring integration? In particular I think the origin use case to populate the authentication information is of interest for many, but its not really clearly documented.

Kehrlann:
We'll definitely update the docs for Spring AI.

We should probably write a blurb for the MCP reference as well. Would you like to work on a docs PR? For inspiration, there are some usage examples of this in [some integration tests](https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp/src/test/java/io/modelcontextprotocol/common/McpTransportContextIntegrationTests.java) ; we're gradually adding webclient as well.

imod:
@Kehrlann I'd be happy to contribute some documentation, but to be honest, I can't get my head around on how to use spring autoconfiguration with MCP and still be able to set `...transportContextProvider(clientContextProvider)`, e.g. like here: https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-spring/mcp-spring-webmvc/src/test/java/io/modelcontextprotocol/common/McpTransportContextIntegrationTests.java#L115-L120

or does it mean I have to opt out of autoconfiguration to be able to do so?

Kehrlann:
Hey @imod - we've updated the docs for MCP itself, here: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1454

For Spring AI auto-configuration, the transport context provider can be injected with a customizer, see [this example](https://github.com/spring-ai-community/mcp-security/blob/590a897838eedeb92d0e6d5498db070da1b32a40/samples/sample-mcp-client/src/main/java/org/springaicommunity/mcp/security/sample/client/McpConfiguration.java#L27-L30). I'm currently working on integration with Spring AI, here: https://github.com/spring-ai-community/mcp-security/ ; it's still a work in progress.

Let's move this discussion to the Spring AI repo itself if you want to pick it up (please tag me in the issue/PR)

PR #524, opened 2025-08-28
fix:avoid streamable listening sse duplicate creation

taobaorun:
## Motivation and Context

In the streamable context, repeated GET requests within the same session spawn multiple listening SSE connections, which 

can lead the server to create unnecessary SSE

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java
+++ b/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java
@@ -10,8 +10,10 @@ import io.modelcontextprotocol.server.DefaultMcpTransportContext;
 import io.modelcontextprotocol.server.McpTransportContextExtractor;
 import io.modelcontextprotocol.spec.HttpHeaders;
 import io.modelcontextprotocol.spec.McpError;
+import io.modelcontextprotocol.spec.McpLoggableSession;
 import io.modelcontextprotocol.spec.McpSchema;
 import io.modelcontextprotocol.spec.McpStreamableServerSession;
+import io.modelcontextprotocol.spec.McpStreamableServerSession.McpStreamableServerSessionStream;
 import io.modelcontextprotocol.spec.McpStreamableServerTransport;
 import io.modelcontextprotocol.spec.McpStreamableServerTransportProvider;
 import io.modelcontextprotocol.spec.ProtocolVersions;
@@ -187,11 +189,17 @@ public class WebFluxStreamableServerTransportProvider implements McpStreamableSe
 				return ServerResponse.notFound().build();
 			}
 
-			if (request.headers().asHttpHeaders().containsKey(HttpHeaders.LAST_EVENT_ID)) {
+			McpLoggableSession listenedStream = session.getListeningStream();
+			boolean replayRequest = request.headers().asHttpHeaders().containsKey(HttpHeaders.LAST_EVENT_ID);
+			if (replayRequest) {
 				String lastId = request.headers().asHttpHeaders().getFirst(HttpHeaders.LAST_EVENT_ID);
 				return ServerResponse.ok()
 					.contentType(MediaType.TEXT_EVENT_STREAM)
 					.body(session.replay(lastId), ServerSentEvent.class);
+			}
+			if (listenedStream instanceof McpStreamableServerSessionStream) {
+				logger.debug("Listening stream for session: {} exists.", sessionId);
+				return ServerResponse.ok().build();
 			}
 
 			return ServerResponse.ok()
--- a/mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
@@ -142,6 +142,10 @@ public class McpStreamableServerSession implements McpLoggableSession {
 		return listeningStream;
 	}
 
+	public McpLoggableSession getListeningStream() {
+		return this.listeningStreamRef.get();
+	}
+
 	// TODO: keep track of history by keeping a map from eventId to stream and then
 	// iterate over the events using the lastEventId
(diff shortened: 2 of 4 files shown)


He-Pin:
Are you working for TaoBao Group?

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java
```diff
@@ -187,12 +189,18 @@ private Mono<ServerResponse> handleGet(ServerRequest request) {
...
+			McpLoggableSession listenedStream = session.getListeningStream();
+			boolean replayRequest = request.headers().asHttpHeaders().containsKey(HttpHeaders.LAST_EVENT_ID);
+			if (replayRequest) {
 				String lastId = request.headers().asHttpHeaders().getFirst(HttpHeaders.LAST_EVENT_ID);
 				return ServerResponse.ok()
 					.contentType(MediaType.TEXT_EVENT_STREAM)
 					.body(session.replay(lastId), ServerSentEvent.class);
 			}
+			if (listenedStream instanceof McpStreamableServerSessionStream) {
+				logger.debug("Listening stream for session: {} exists.", sessionId);
```
He-Pin: if(logger.isDebugEnabled) ...

chemicL:
Hey, thanks for the proposal. I see this might be an issue, agreed. However, my worry is that someone might be making a new GET request because the previous one is broken. The way forward that I see is not to reject the new GET request but instead make sure that if one listening stream exists it is properly closed and replaced with this new one instead. WDYT?

taobaorun:
> Hey, thanks for the proposal. I see this might be an issue, agreed. However, my worry is that someone might be making a new GET request because the previous one is broken. The way forward that I see is not to reject the new GET request but instead make sure that if one listening stream exists it is properly closed and replaced with this new one instead. WDYT?

Thanks for the helpful suggestion—code updated, tested, and confirmed working.

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java
```diff
@@ -187,12 +189,19 @@ private Mono<ServerResponse> handleGet(ServerRequest request) {
...
 				String lastId = request.headers().asHttpHeaders().getFirst(HttpHeaders.LAST_EVENT_ID);
 				return ServerResponse.ok()
 					.contentType(MediaType.TEXT_EVENT_STREAM)
 					.body(session.replay(lastId), ServerSentEvent.class);
 			}
+			if (listenedStream instanceof McpStreamableServerSessionStream) {
+				logger.debug(
+						"Listening stream already exists for this session:{} and will be closed to make way for the new listening SSE stream",
+						sessionId);
+				listenedStream.close();
```
chemicL: To ensure proper timing you should actually combine `closeGracefully` with the rest of the code execution by wrapping the latter in a `Mono` lambda (`Mono.defer` for example). The method `close` is can be async and calling it like that could lead to race conditions if their implementation is changed in the future.

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java
```diff
@@ -187,12 +189,19 @@ private Mono<ServerResponse> handleGet(ServerRequest request) {
...
+			}
```
chemicL: Please notice that this logic can be part of `session.listeningStream(sessionTransport)` from line 211 in which case we'd avoid repeating this code in all the implementations and we'd avoid adding a new method for the getter since the reference is available internally (`this.listeningStreamRef`). In the end it should be an atomic replace operation.

taobaorun: Thanks for the suggestion—could you please take another look at the latest commit to see if it meets the requirements? Much appreciated.

mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
```diff
@@ -138,7 +136,13 @@ public Mono<Void> delete() {
 	 */
 	public McpStreamableServerSessionStream listeningStream(McpStreamableServerTransport transport) {
 		McpStreamableServerSessionStream listeningStream = new McpStreamableServerSessionStream(transport);
-		this.listeningStreamRef.set(listeningStream);
+		McpLoggableSession listenedStream = this.listeningStreamRef.getAndSet(listeningStream);
+		if (listenedStream != null) {
+			logger.debug(
+					"Listening stream already exists for this session:{} and will be closed to make way for the new listening SSE stream",
+					this.id);
+			listenedStream.closeGracefully().block();
```
chemicL: We can't just call block as this might be executed as part of a non-blocking execution model

chemicL (changes requested):
I left a comment that needs addressing if you'd like to take it forward. Also, I think we need tests to validate that this change is doing what is expected. Let us know if you want to continue making progress here. Thanks!

taobaorun:
> I left a comment that needs addressing if you'd like to take it forward. Also, I think we need tests to validate that this change is doing what is expected. Let us know if you want to continue making progress here. Thanks!

I will continue to follow up and solve this problem.🤝

mcp/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
taobaorun: fixed close block

taobaorun:
> I left a comment that needs addressing if you'd like to take it forward. Also, I think we need tests to validate that this change is doing what is expected. Let us know if you want to continue making progress here. Thanks!

@chemicL Please help review the changes，Thanks

PR #529, opened 2025-09-02
test: Add additional MCP transport context integration tests

tzolov:
- Add integration tests for transport context propagation between MCP clients and servers
- Test both sync and async server implementations across all transport types (stateless, streamable, SSE)
- Cover Spring WebFlux and WebMVC environments with dedicated test suites
- Validate context flow through HTTP headers for authentication, correlation IDs, and metadata
- Rename existing McpTransportContextIntegrationTests to SyncServerMcpTransportContextIntegrationTests for clarity

This change adds missing test coverage for the MCP transport context feature, which allows passing contextual information (like authentication tokens, correlation IDs, or custom metadata) between MCP clients and servers through HTTP headers. The tests ensure that context propagation works correctly across:
- Different server types (sync/async)
- Different transport mechanisms (stateless, streamable, SSE)
- Different Spring environments (WebFlux, WebMVC)
- Different client types (sync/async)

## Breaking Changes

None. This PR only adds new test files and renames one existing test file for clarity. No production code or APIs are changed.

## Types of changes

- [x] Documentation update (in this case, test coverage which serves as usage documentation)

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/common/AsyncServerMcpTransportContextIntegrationTests.java
+++ b/mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/common/AsyncServerMcpTransportContextIntegrationTests.java
@@ -6,12 +6,10 @@ package io.modelcontextprotocol.common;
 
 import java.util.Map;
 import java.util.function.BiFunction;
-import java.util.function.Supplier;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.modelcontextprotocol.client.McpAsyncClient;
 import io.modelcontextprotocol.client.McpClient;
-import io.modelcontextprotocol.client.McpSyncClient;
 import io.modelcontextprotocol.client.transport.WebClientStreamableHttpTransport;
 import io.modelcontextprotocol.client.transport.WebFluxSseClientTransport;
 import io.modelcontextprotocol.server.McpAsyncServerExchange;
@@ -49,14 +47,13 @@ import static org.assertj.core.api.Assertions.assertThat;
  *
  * <p>
  * This test class validates the end-to-end flow of transport context propagation in MCP
- * communication for asynchronous server implementations. It tests various combinations of
- * client types (sync/async) and server transport mechanisms (stateless, streamable, SSE)
- * to ensure proper context handling across different configurations.
+ * communication for asynchronous client and server implementations. It tests various
+ * combinations of client types and server transport mechanisms (stateless, streamable,
+ * SSE) to ensure proper context handling across different configurations.
  *
  * <h2>Context Propagation Flow</h2>
  * <ol>
- * <li>Client sets a value in its transport context (either via thread-local for sync or
- * Reactor context for async)</li>
+ * <li>Client sets a value in its transport context via thread-local Reactor context</li>
  * <li>Client-side context provider extracts the value and adds it as an HTTP header to
  * the request</li>
  * <li>Server-side context extractor reads the header from the incoming request</li>
@@ -67,33 +64,19 @@ import static org.assertj.core.api.Assertions.assertThat;
  *
  * @author Daniel Garnier-Moiroux
  * @author Christian Tzolov
- * @see McpTransportContext
- * @see McpTransportContextExtractor
- * @see WebFluxStatelessServerTransport
- * @see WebFluxStreamableServerTransportProvider
- * @see WebFluxSseServerTransportProvider
  */
 @Timeout(15)
 public class AsyncServerMcpTransportContextIntegrationTests {
 
 	private static final int PORT = TestUtil.findAvailablePort();
 
-	private static final ThreadLocal<String> SYNC_CLIENT_SIDE_HEADER_VALUE_HOLDER = new ThreadLocal<>();
-
 	private static final String HEADER_NAME = "x-test";
-
-	// Sync client context provider
-	private final Supplier<McpTransportContext> syncClientContextProvider = () -> {
-		var headerValue = SYNC_CLIENT_SIDE_HEADER_VALUE_HOLDER.get();
-		return headerValue != null ? McpTransportContext.create(Map.of("client-side-header-value", headerValue))
-				: McpTransportContext.EMPTY;
-	};
 
 	// Async client context provider
 	ExchangeFilterFunction asyncClientContextProvider = (request, next) -> Mono.deferContextual(ctx -> {
-		var context = ctx.getOrDefault(McpTransportContext.KEY, McpTransportContext.EMPTY);
+		var transportContext = ctx.getOrDefault(McpTransportContext.KEY, McpTransportContext.EMPTY);
 		// // do stuff with the context
-		var headerValue = context.get("client-side-header-value");
+		var headerValue = transportContext.get("client-side-header-value");
 		if (headerValue == null) {
 			return next.exchange(request);
 		}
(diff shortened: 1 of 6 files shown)


mcp/src/test/java/io/modelcontextprotocol/common/AsyncServerMcpTransportContextIntegrationTests.java
```diff
@@ -0,0 +1,393 @@
...
+
+	private static final ThreadLocal<String> CLIENT_SIDE_HEADER_VALUE_HOLDER = new ThreadLocal<>();
+
+	private static final String HEADER_NAME = "x-test";
+
+	private final Supplier<McpTransportContext> clientContextProvider = () -> {
+		var headerValue = CLIENT_SIDE_HEADER_VALUE_HOLDER.get();
+		return headerValue != null ? McpTransportContext.create(Map.of("client-side-header-value", headerValue))
+				: McpTransportContext.EMPTY;
+	};
```
Kehrlann: **[suggestion]** I don't think we need to test sync clients and async clients. Sync clients are already tested with Sync servers ; hopefully all servers and all clients are spec-compliant, and breaking something in a client should result in an issue, now matter what type of server is used (and vice-versa). Otherwise the matrix starts to get quite big.

So I suggest:
- SyncServer tests -> use sync clients
- AsyncServer test -> use async clients

This way they are "naturally" paired - even though sync server + async client and async server + sync client would also work.

mcp/src/test/java/io/modelcontextprotocol/common/AsyncServerMcpTransportContextIntegrationTests.java
```diff
@@ -0,0 +1,393 @@
...
+
+	private final McpSyncHttpClientRequestCustomizer clientRequestCustomizer = (builder, method, endpoint, body,
+			context) -> {
+		var headerValue = context.get("client-side-header-value");
+		if (headerValue != null) {
+			builder.header(HEADER_NAME, headerValue.toString());
+		}
+	};
```
Kehrlann: **[suggestion]** Let's use an async request customizer. It should change anything but serves as an example.

mcp-spring/mcp-spring-webmvc/src/test/java/io/modelcontextprotocol/common/McpTransportContextIntegrationTests.java
```diff
@@ -0,0 +1,320 @@
...
+		@Bean
+		public McpTransportContextExtractor<ServerRequest> serverContextExtractor() {
+			return (ServerRequest r) -> {
+				String headerValue = r.servletRequest().getHeader(HEADER_NAME);
+				return headerValue != null ? McpTransportContext.create(Map.of("server-side-header-value", headerValue))
+						: McpTransportContext.EMPTY;
+			};
+		};
+
+	}
```
Kehrlann: **[question]** does this need to be a bean wrapped in a config class? It holds a static function that could be used "as-is".

mcp-spring/mcp-spring-webmvc/src/test/java/io/modelcontextprotocol/common/McpTransportContextIntegrationTests.java
```diff
@@ -0,0 +1,320 @@
...
+		}
+
+	}
+
+	private final McpSyncClient streamableClient = McpClient
+		.sync(HttpClientStreamableHttpTransport.builder("http://localhost:" + PORT)
+			.httpRequestCustomizer(clientRequestCustomizer)
+			.build())
+		.transportContextProvider(clientContextProvider)
+		.build();
```
Kehrlann, line 113: **[suggestion]** swap the order, the beans are between fields of this class.

mcp-spring/mcp-spring-webmvc/src/test/java/io/modelcontextprotocol/common/McpTransportContextIntegrationTests.java
```diff
@@ -0,0 +1,320 @@
...
+	@Test
+	void statelessServer() {
+		startTomcat(TestStatelessConfig.class);
+
+		var statelessServerTransport = tomcatServer.appContext().getBean(WebMvcStatelessServerTransport.class);
+
+		var mcpServer = McpServer.sync(statelessServerTransport)
+			.capabilities(McpSchema.ServerCapabilities.builder().tools(true).build())
+			.tools(new McpStatelessServerFeatures.SyncToolSpecification(tool, statelessHandler))
+			.build();
```
Kehrlann: **[suggestion]** Make the server a bean. This way, all the servery bits are in a config class.

mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/common/AsyncServerMcpTransportContextIntegrationTests.java
```diff
@@ -0,0 +1,381 @@
...
+	// Sync client context provider
+	private final Supplier<McpTransportContext> syncClientContextProvider = () -> {
+		var headerValue = SYNC_CLIENT_SIDE_HEADER_VALUE_HOLDER.get();
+		return headerValue != null ? McpTransportContext.create(Map.of("client-side-header-value", headerValue))
+				: McpTransportContext.EMPTY;
+	};
+
+	// Async client context provider
+	ExchangeFilterFunction asyncClientContextProvider = (request, next) -> Mono.deferContextual(ctx -> {
+		var context = ctx.getOrDefault(McpTransportContext.KEY, McpTransportContext.EMPTY);
```
Kehrlann: **[nit]** clarify distinction between reactor context (ctx) and McpTransportContext
```suggestion
		var transportContext = ctx.getOrDefault(McpTransportContext.KEY, McpTransportContext.EMPTY);
```

mcp-spring/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/common/AsyncServerMcpTransportContextIntegrationTests.java
```diff
@@ -0,0 +1,381 @@
...
+	private static final ThreadLocal<String> SYNC_CLIENT_SIDE_HEADER_VALUE_HOLDER = new ThreadLocal<>();
+
+	private static final String HEADER_NAME = "x-test";
+
+	// Sync client context provider
+	private final Supplier<McpTransportContext> syncClientContextProvider = () -> {
+		var headerValue = SYNC_CLIENT_SIDE_HEADER_VALUE_HOLDER.get();
+		return headerValue != null ? McpTransportContext.create(Map.of("client-side-header-value", headerValue))
+				: McpTransportContext.EMPTY;
+	};
```
Kehrlann: Same comment as in `AsyncServerMcpTransportContextIntegrationTests` : consider using only async clients here.

Kehrlann (commented):
Nice and thorough work! We have a much better confidence in transports with this.  Thanks.

I wonder if we don't have a structural issue with these tests, even though that existed before.

We always build all server transports as the field of classes. And then each test only calls `.close()` on the server that it is using, which in turns closes the underlying server transport - but other transports in the class never have their `.close()` called. Is that a problem, or will everything be magically garbage-collected away? Maybe @chemicL has insights for us.

(Also, if a test fails, `mcpServer.close()` is never called).

tzolov:
@Kehrlann , thanks for the review!  
Latest commit should address most of the recommendation.

Kehrlann:
Looks good. Thanks for the updates!

PR #422, opened 2025-07-24
Rework MCP error handling

tzolov:
Refactor exception handling to introduce a proper exception hierarchy, replacing generic `McpError` usage with specific exception types for better error categorization and handling.

- Add McpTransportException for transport layer errors
- Add McpClientInternalException for internal client failures
- Use standard Java exceptions for validation errors
- Implement structured error builder with JSON-RPC error codes
- Change tool not found errors to be returned in CallToolResult
- Update logging levels for unregistered notification handlers
    
## Motivation and Context
The existing codebase was using `McpError` as a catch-all exception type for various error scenarios, making it difficult to distinguish between different types of failures. This change addresses several issues:

- **Poor error categorization**: Transport errors, validation errors, and protocol errors were all using the same exception type
- **Missing error context**: Generic exceptions didn't provide enough context for proper error handling
- **Tool error handling**: Tool errors were causing protocol-level failures instead of being visible to LLMs
- **Inconsistent error structure**: MCP protocol errors weren't following proper JSON-RPC error format

## How Has This Been Tested?
- All existing unit tests have been updated and pass with the new exception types
- Integration tests across all transport implementations (SSE, HTTP, Stdio) continue to pass
- Test coverage includes all error scenarios for each exception type
- Verified that tool errors are now properly returned as results rather than thrown exceptions

## Breaking Changes
**Yes, this is a breaking change** for applications that catch `McpError` exceptions. Users will need to update their exception handling code to catch the new specific exception types:

- `McpTransportException` for transport-related errors
- `McpClientInternalException` for internal client errors  
- `IllegalArgumentException` for validation errors
- `IllegalStateException` for state-related errors
- `McpError` now only for actual MCP protocol JSON-RPC errors

## Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Bug fix (non-breaking change which fixes an issue) - Tool error handling improvement
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

### Key Changes:
1. **New Exception Classes**:
   - `McpTransportException`: For transport layer errors (network, serialization, SSE parsing)
   - `McpClientInternalException`: For internal client errors (initialization failures, endpoint handling)

2. **McpError Refinement**: 
[...]

PR #540, opened 2025-09-07
fix: align tool error handling with MCP specification

### LINKED ISSUE #538: StructuredOutputCallToolHandler doesn't handle error based CallToolResult
https://github.com/modelcontextprotocol/java-sdk/blob/681fe7f90b12191ba18dacaffa5635d68c65a9c0/mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java#L377-L379

The method is checking for various combinations of content compared to the declared output schema, but if the result is an error then this should be just returned irrespective of the schema and structuredContent values i.e.

```
if (Boolean.TRUE.equals(result.isError())) {
    return result;
}
```

tzolov:
Implement proper error handling for tool calls according to MCP Tools' Error Handling specification (https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling)

- Tool errors are now returned within CallToolResult with isError=true instead of throwing MCP protocol-level errors
- This allows LLMs to see and handle tool errors as part of normal flow
- Return proper JSON-RPC error with code -32602 in case of "Unknown tools"
- Added Utils.findRootCause() helper to extract root cause from exception chains
- Skip output schema validation when tool result already indicates an error
- Updated all integration tests to verify new error handling behavior

BREAKING CHANGE: Tool call errors are no longer thrown as McpError exceptions. They are now returned as CallToolResult objects with isError=true and error details in the content field, following MCP specification guidelines.

Resolves #538
Related to #422

Changes tool error handling to return errors within CallToolResult objects instead of throwing MCP protocol-level exceptions. This allows LLMs to see and handle tool errors gracefully.

## Motivation and Context

Previously, when a tool call failed (e.g., due to timeout, missing capabilities, or runtime errors), the SDK would throw an McpError exception at the protocol level. This prevented LLMs from seeing and potentially handling these errors, as they would appear as protocol failures rather than tool execution failures.

This change aligns with MCP best practices where tool errors should be reported within the result object, allowing the LLM to:
- See the error and potentially retry or use alternative approaches
- Provide better error messages to users
- Distinguish between protocol-level failures and tool execution failures

## How Has This Been Tested?

- Updated all existing integration tests to verify the new error handling behavior
- Tested scenarios include:
  - Tool calls without required sampling capabilities
  - Tool calls with timeouts
  - Tool calls with runtime exceptions
  - Unknown tool requests
- All tests pass with the new error handling approach

## Breaking Changes

**Yes, this is a breaking change.** 

Users who have code that catches `McpError` exceptions from `callTool()` methods will need to update their code to check the `isError()` field of the returned `CallToolResult` instead:

**Before:**
```java
try {
    CallToolResult result = client.callTool(request);
} catch (McpError e) {
    // Handle error
}
```

**After:**
```java
CallToolResult result = client.callTool(request);
if (result.isError()) {
    // Handle error - error message is in result.content()
}
```

## Types of changes

- [ ] Documentation update

## Checklist

### Implementation Details:
- Added `findRootCause()` utility method to extract the root cause from exception chains
- Modified `McpAsyncServer` and `McpStatelessAsyncServer` to catch tool handler exceptions and wrap them in error results
- Error results include `isError=true` and a `TextContent` message with the error details
- Validation is skipped for error results (no schema validation when `isError=true`)
- Unknown tool errors now return INVALID_PARAMS error code for better protocol compliance

### Migration Guide:
Applications using the SDK should update their error handling logic to check `CallToolResult.isError()` instead of catching exceptions. The error message can be extracted from `result.content()` which will contain a `TextContent` object with the error details.

### DIFF
--- a/mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
+++ b/mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
@@ -29,10 +29,12 @@ import io.modelcontextprotocol.spec.McpClientSession;
 import io.modelcontextprotocol.spec.McpError;
 import io.modelcontextprotocol.spec.McpSchema;
 import io.modelcontextprotocol.spec.McpSchema.CallToolResult;
+import io.modelcontextprotocol.spec.McpSchema.JSONRPCResponse;
 import io.modelcontextprotocol.spec.McpSchema.LoggingLevel;
 import io.modelcontextprotocol.spec.McpSchema.LoggingMessageNotification;
 import io.modelcontextprotocol.spec.McpSchema.ResourceTemplate;
 import io.modelcontextprotocol.spec.McpSchema.SetLevelRequest;
+import io.modelcontextprotocol.spec.McpSchema.TextContent;
 import io.modelcontextprotocol.spec.McpSchema.Tool;
 import io.modelcontextprotocol.spec.McpServerSession;
 import io.modelcontextprotocol.spec.McpServerTransportProvider;
@@ -376,6 +378,11 @@ public class McpAsyncServer {
 
 			return this.delegateCallToolResult.apply(exchange, request).map(result -> {
 
+				if (result.isError() != null && result.isError()) {
+					// If the tool call resulted in an error, skip further validation
+					return result;
+				}
+
 				if (outputSchema == null) {
 					if (result.structuredContent() != null) {
 						logger.warn(
@@ -507,11 +514,23 @@ public class McpAsyncServer {
 				.findAny();
 
 			if (toolSpecification.isEmpty()) {
-				return Mono.error(new McpError("Tool not found: " + callToolRequest.name()));
+				return Mono.error(new McpError(new JSONRPCResponse.JSONRPCError(McpSchema.ErrorCodes.INVALID_PARAMS,
+						"Unknown tool: invalid_tool_name", "Tool not found: " + callToolRequest.name())));
 			}
+			else {
+				return toolSpecification.get().callHandler().apply(exchange, callToolRequest).onErrorResume(error -> {
+					logger.error("Error calling tool: {}", callToolRequest.name(), error);
 
-			return toolSpecification.map(tool -> Mono.defer(() -> tool.callHandler().apply(exchange, callToolRequest)))
-				.orElse(Mono.error(new McpError("Tool not found: " + callToolRequest.name())));
+					// Tool errors should be reported within the result object, not as MCP
+					// protocol-level errors. This allows the LLM to see and potentially
+					// handle the error.
+					return Mono.just(CallToolResult.builder()
+						.isError(true)
+						.content(List
+							.of(new TextContent("Error calling tool: " + Utils.findRootCause(error).getMessage())))
+						.build());
+				});
+			}
 		};
 	}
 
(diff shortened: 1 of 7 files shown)


mcp/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
```diff
@@ -375,11 +382,31 @@ private McpStatelessRequestHandler<CallToolResult> toolsCallRequestHandler() {
...
+						"Unknown tool: invalid_tool_name", "Tool not found: " + callToolRequest.name())));
 			}
+			else {
+				return toolSpecification.get().callHandler().apply(ctx, callToolRequest).onErrorResume(error -> {
+					logger.error("Error calling tool: {}", callToolRequest.name(), error);
+
+					// TODO: Should we handle the McpError+JsonRcpError specaially (e.g.
+					// propagate)
+					// or always return a CallToolResult with isError=true?
+					if (error instanceof McpError mcpError && mcpError.getJsonRpcError() != null) {
```
tzolov: I'm not sure this this assumption is correct? should we thread McpError + JsonRpcError content differently? 
If so we have to implement this logic for the McpAsyncServer's tool handler. Otherwise remove this condition here.
@chemicL any thoughts?

chemicL: I'm guessing we shouldn't allow users to throw a generic McpError? This way we don't have to perform these checks but provide the users a framework of errors to indicate the result. So e.g. in case of logical errors they either return the CallToolResult with isError themselves or fail with a known exception type. From the outside we can't tell what is the reason so without such family of errors we'd have to assume it's a server error and we treat it as JSON-RPC error response.

tzolov:
Thanks @chemicL . I've updated the PR.

mcp/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
```diff
@@ -249,6 +251,11 @@ public Mono<CallToolResult> apply(McpTransportContext transportContext, McpSchem
 
 			return this.delegateHandler.apply(transportContext, request).map(result -> {
 
+				if (result.isError() != null && result.isError()) {
```
chemicL, line 254: nit: can simplify to `Boolean.TRUE.equals(result.isError())`

mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -376,6 +378,11 @@ public Mono<CallToolResult> apply(McpAsyncServerExchange exchange, McpSchema.Cal
 
 			return this.delegateCallToolResult.apply(exchange, request).map(result -> {
 
+				if (result.isError() != null && result.isError()) {
```
chemicL, line 381: nit: can simplify to `Boolean.TRUE.equals(result.isError())`

mcp/src/main/java/io/modelcontextprotocol/util/Utils.java
```diff
@@ -107,4 +108,19 @@ private static boolean isUnderBaseUri(URI baseUri, URI endpointUri) {
 		return endpointPath.startsWith(basePath);
 	}
 
+	/**
+	 * Finds the root cause of the given throwable by traversing the cause chain.
+	 * @param throwable The throwable to analyze
+	 * @return The root cause throwable
+	 * @throws NullPointerException if the provided throwable is null
+	 */
+	public static Throwable findRootCause(Throwable throwable) {
```
chemicL, line 117: We don't need this any more

chemicL (approved):
Three minor comments left, please apply when merging. Otherwise all good!

tzolov:
Thanks @chemicL I've addressed those before the merge!

tzolov:
Rebased, squashed, cleaned and merged at improve-tool-call-error-handling c7736b6ec6b1eef9da48826294bfa9ba3e7f77d1
backported to 0.12.x

PR #543, opened 2025-09-10
Introduce McpJsonMapper interface to decouple from Jackson

graemerocher:
## Motivation and Context
The current MCP SDK is coupled to Jackson which prevents the use of other JSON serialization libraries and techniques.

## How Has This Been Tested?
Yes tests included.

## Breaking Changes
There are deprecations but the intent is to keep breaking changes to a minimum in this PR. Ideally one would remove the Jackson APIs completely from the public interfaces, but that would constitute a breaking change.

## Types of changes

- [ ] Documentation update

## Checklist

- [x ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x ] My code follows the repository's style guidelines
- [x ] New and existing tests pass locally
- [x ] I have added appropriate error handling

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
+++ b/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
@@ -22,8 +22,9 @@ import org.springframework.web.reactive.function.client.ClientResponse;
 import org.springframework.web.reactive.function.client.WebClient;
 import org.springframework.web.reactive.function.client.WebClientResponseException;
 
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.ObjectMapper;
+import io.modelcontextprotocol.spec.json.TypeRef;
+import io.modelcontextprotocol.spec.json.McpJsonMapper;
+import io.modelcontextprotocol.spec.json.jackson.JacksonMcpJsonMapper;
 
 import io.modelcontextprotocol.spec.DefaultMcpTransportSession;
 import io.modelcontextprotocol.spec.DefaultMcpTransportStream;
@@ -88,7 +89,7 @@ public class WebClientStreamableHttpTransport implements McpClientTransport {
 	private static final ParameterizedTypeReference<ServerSentEvent<String>> PARAMETERIZED_TYPE_REF = new ParameterizedTypeReference<>() {
 	};
 
-	private final ObjectMapper objectMapper;
+	private final McpJsonMapper jsonMapper;
 
 	private final WebClient webClient;
 
@@ -104,9 +105,9 @@ public class WebClientStreamableHttpTransport implements McpClientTransport {
 
 	private final AtomicReference<Consumer<Throwable>> exceptionHandler = new AtomicReference<>();
 
-	private WebClientStreamableHttpTransport(ObjectMapper objectMapper, WebClient.Builder webClientBuilder,
+	private WebClientStreamableHttpTransport(McpJsonMapper jsonMapper, WebClient.Builder webClientBuilder,
 			String endpoint, boolean resumableStreams, boolean openConnectionOnStartup) {
-		this.objectMapper = objectMapper;
+		this.jsonMapper = jsonMapper;
 		this.webClient = webClientBuilder.build();
 		this.endpoint = endpoint;
 		this.resumableStreams = resumableStreams;
@@ -366,8 +367,7 @@ public class WebClientStreamableHttpTransport implements McpClientTransport {
 			McpSchema.JSONRPCResponse.JSONRPCError jsonRpcError = null;
 			Exception toPropagate;
 			try {
-				McpSchema.JSONRPCResponse jsonRpcResponse = objectMapper.readValue(body,
-						McpSchema.JSONRPCResponse.class);
+				McpSchema.JSONRPCResponse jsonRpcResponse = jsonMapper.readValue(body, McpSchema.JSONRPCResponse.class);
 				jsonRpcError = jsonRpcResponse.error();
 				toPropagate = jsonRpcError != null ? new McpError(jsonRpcError)
 						: new McpTransportException("Can't parse the jsonResponse " + jsonRpcResponse);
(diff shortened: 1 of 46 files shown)


mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
```diff
@@ -494,9 +494,20 @@ private Builder(WebClient.Builder webClientBuilder) {
 		 * @param objectMapper instance to use
 		 * @return the builder instance
 		 */
-		public Builder objectMapper(ObjectMapper objectMapper) {
+		public Builder objectMapper(com.fasterxml.jackson.databind.ObjectMapper objectMapper) {
 			Assert.notNull(objectMapper, "ObjectMapper must not be null");
-			this.objectMapper = objectMapper;
+			this.jsonMapper = new JacksonMcpJsonMapper(objectMapper);
+			return this;
```
sdelamo:
```suggestion
            return jsonMapper(new JacksonMcpJsonMapper(objectMapper));
```

He-Pin:
Do you want to use Vert.x's json?

sdelamo:
> Do you want to use Vert.x's json?

We want to use [Micronaut Serialization](https://micronaut-projects.github.io/micronaut-serialization/latest/guide). But basically, the idea is to allow users to use whatever JSON Serialization they want.

sdeleuze:
I did not review everything in detail, but from a high level perspective, this PR looks great, thanks for contributing it @graemerocher @sdelamo.

The main open question for me is related to Jackson 3 which is about to be released. If we want to support Jackson 2 and Jackson 3 (which continues to use the same `jackson-annotation:2.x` dependency) side by side in the MCP SDK, we may want to rename the module `mcp-json-jackson` to `mcp-json-jackson2` and use a `jackson2` subpackage instead of `jackson` in this module.

If we just plan to upgrade from Jackson 2 to Jackson 3, current naming is fine.

graemerocher:
probably including the major version makes sense for Jackson.

mcp/pom.xml
```diff
@@ -65,6 +65,18 @@
 	</build>
 
 	<dependencies>
+        <dependency>
+            <groupId>io.modelcontextprotocol.sdk</groupId>
+            <artifactId>mcp-json</artifactId>
+            <version>0.13.0-SNAPSHOT</version>
+        </dependency>
+
+        <dependency>
```
tzolov: move to the test section of the pom

mcp-json/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>io.modelcontextprotocol.sdk</groupId>
+		<artifactId>mcp-parent</artifactId>
+		<version>0.13.0-SNAPSHOT</version>
+	</parent>
+	<artifactId>mcp-json</artifactId>
```
tzolov, line 11: add to the mcp-bom

mcp-json-jackson/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>io.modelcontextprotocol.sdk</groupId>
+		<artifactId>mcp-parent</artifactId>
+		<version>0.13.0-SNAPSHOT</version>
+	</parent>
+	<artifactId>mcp-json-jackson</artifactId>
```
tzolov: add to the mcp-bom

mcp-json/pom.xml
sdelamo: see: https://github.com/graemerocher/java-sdk/pull/2/commits/1b4271be4ece47b1e56e2b106bafe4d8b49e6284

mcp-json-jackson/pom.xml
sdelamo: see: https://github.com/graemerocher/java-sdk/pull/2/commits/1b4271be4ece47b1e56e2b106bafe4d8b49e6284

sdelamo:
@tzolov 

 >There's another conceptual change that I'm puzzled with. Currently the mcp is self-contained. But with this PR it becomes a "core" component requiring an additional mcp-json-xxx (like mcp-json-jackson) dependency for the same bootstrap experience without breaking changes.
 
>  wonder if we should rename the existing mcp module to mcp-core and create a small mcp module that includes both mcp-core and mcp-json-jackson by default?
This will preserver the existing single-click experience for the end users

You are right. To avoid breaking changes, I left in this PR mcp-json-jackson as a compile dependency. Once this PR gets merged, I will create a PR that moves the mcp classes to mcp-core.

sdelamo:
> I did not review everything in detail, but from a high level perspective, this PR looks great, thanks for contributing it @graemerocher @sdelamo.
> 
> The main open question for me is related to Jackson 3 which is about to be released. If we want to support Jackson 2 and Jackson 3 (which continues to use the same `jackson-annotation:2.x` dependency) side by side in the MCP SDK, we may want to rename the module `mcp-json-jackson` to `mcp-json-jackson2` and use a `jackson2` subpackage instead of `jackson` in this module.
> 
> If we just plan to upgrade from Jackson 2 to Jackson 3, current naming is fine.

[I renamed to module to `mcp-json-jackson2`](https://github.com/graemerocher/java-sdk/pull/2/commits/33a61196ed1db2148b7bc683f2d64e627c8b6e0a)

tzolov:
Thank you @graemerocher @sdelamo 

Rebased, conflicts resolved, squashed and merged at 80d0ad82a6b88a8ce8756dad3d4c90c4ae62ca69

PR #554, opened 2025-09-16
[fix] when mcp server not give mcp session id ,mcp session store null…

caojiaqi123:
… as session id

Before do reconnect, we should check the session id is null or not

## Motivation and Context
Mcp session id is optional, if mcp server did not return mcp session id, we will get a null value. it will throw a MCP Session id missing exception.

## How Has This Been Tested?
Build a mcp server, which not return mcp session id in response header.

## Breaking Changes
No
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
+++ b/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
@@ -106,7 +106,7 @@ public class WebClientStreamableHttpTransport implements McpClientTransport {
 	private final AtomicReference<Consumer<Throwable>> exceptionHandler = new AtomicReference<>();
 
 	private WebClientStreamableHttpTransport(McpJsonMapper jsonMapper, WebClient.Builder webClientBuilder,
-											 String endpoint, boolean resumableStreams, boolean openConnectionOnStartup) {
+			String endpoint, boolean resumableStreams, boolean openConnectionOnStartup) {
 		this.jsonMapper = jsonMapper;
 		this.webClient = webClientBuilder.build();
 		this.endpoint = endpoint;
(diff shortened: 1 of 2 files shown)


quaff:
>> A server using the Streamable HTTP transport MAY assign a session ID at initialization time, by including it in an Mcp-Session-Id header on the HTTP response containing the InitializeResult.

Yes, the `Mcp-Session-Id` is optional.

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
```diff
@@ -4,46 +4,35 @@
...
-import io.modelcontextprotocol.spec.ProtocolVersions;
-import io.modelcontextprotocol.util.Assert;
-import io.modelcontextprotocol.util.Utils;
 import reactor.core.Disposable;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.util.function.Tuple2;
 import reactor.util.function.Tuples;
 
+import java.io.IOException;
```
quaff: You should revert unrelated parts, keep null check changes only.

caojiaqi123: ok, reverted unrelated parts. Submit MR first time, thx reminder. @quaff

mcp/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
```diff
@@ -442,8 +443,8 @@ public Mono<Void> sendMessage(McpSchema.JSONRPCMessage sentMessage) {
 					})).onErrorMap(CompletionException.class, t -> t.getCause()).onErrorComplete().subscribe();
 
 			})).flatMap(responseEvent -> {
-				if (transportSession.markInitialized(
-						responseEvent.responseInfo().headers().firstValue("mcp-session-id").orElseGet(() -> null))) {
+				String mcpSessionId = responseEvent.responseInfo().headers().firstValue("mcp-session-id").orElseGet(() -> null);
+				if (Objects.nonNull(mcpSessionId) && transportSession.markInitialized(mcpSessionId)) {
```
quaff, line 447: >> API Note:
This method exists to be used as a java.util.function.Predicate, filter(Objects::nonNull)

I think `mcpSessionId != null` should be used here.

caojiaqi123: Oh, I see. It is used to listen for notifications. thx!

Issue #547, opened 2025-09-11
DefaultMcpTransportSession#closeGracefully leaks resources on incomplete DELETE response


1256064203:
**Bug description**
When the MCP server is abnormal, the statement this.onClose.apply(this.sessionId.get()) in the closeGracefully method of DefaultMcpTransportSession.java throws an exception, so this.openConnections::dispose is never executed and may leak resources.

**Environment**
Java version: 21
mcp-bom: 0.11.3

**Steps to reproduce**
1、Create an McpSyncClient.
2、Execute client.initialize(); client.listTools();
3、Suspend the debugger at client.closeGracefully();
4、Make the MCP server unavailable.
5、Resume execution of client.closeGracefully();

**Expected behavior**
this.openConnections::dispose should always be executed.

**Minimal Complete Reproducible example**
McpSyncClient client = McpClient.sync(transport)
    .requestTimeout(Duration.ofSeconds(10))
    .build();

// Initialize connection
client.initialize();

// List available tools
ListToolsResult tools = client.listTools();

// Close client
client.closeGracefully();

JunJieLiu51520:
Can you take a screenshot of the leaked stack trace?  from this code, i think it  won't be happed

<img width="2407" height="771" alt="Image" src="https://github.com/user-attachments/assets/3103a372-380f-4631-b861-bb0c957f2e63" />

CnLight:
<img width="1429" height="605" alt="Image" src="https://github.com/user-attachments/assets/5933e2a7-8217-4999-b5e5-676a5443d5e1" /> this worker  will be close but selector always running, the way of close can not shutdown the httpclient

chemicL:
Is this perhaps related to the inability to close the HttpClient properly in Java 17?

chemicL:
An estensive analysis was published in the duplicate issue: https://github.com/modelcontextprotocol/java-sdk/issues/620#issue-3512175494

andersenleo:
#566 covers the case where the DELETE fails. It does not cover the case where the DELETE never answers, and that is the one that leaks.

`closeGracefully()` is:

```java
return Mono.from(this.onClose.apply(this.sessionId.get()))
    .then(Mono.fromRunnable(this.openConnections::dispose));
```

#566 wraps the first stage in `onErrorResume` and keeps the `.then(...)`. A DELETE that hangs signals no error, so `dispose()` is never reached.

[...]

PR #566, opened 2025-09-22
fix(session): always dispose in closeGracefully

phanikpemmaraju:
## Motivation and Context

- "closeGracefully()" could skip disposing open connections when onClose threw, leading to leaked resources during server outages or abrupt disconnects.
- MCP session id is optional; when the server doesn’t return one, client flows that assumed a non-null id would throw "Session ID missing" prematurely.

## How Has This Been Tested?

"Unit tests"

- closeGracefully disposes connections when onClose throws.
- Disposal exceptions are aggregated/suppressed as appropriate; error semantics preserved.

"Integration tests"

- WebFlux/SSE: server crashes mid-session → closeGracefully() best-effort local teardown still runs (existing WebFluxSseCloseGracefullyIntegrationTests).

## Breaking Changes

- "None"

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

"What's changed"

- Wrap onClose.apply(sessionId) in try/finally and always dispose openConnections in the finally block.
- Add a null/absent session id guard before reconnect/close paths; treat missing id as “no remote teardown”, but still perform local cleanup.
- Improved test coverage (unit + integration).

"Why"

- Ensures robust cleanup even when the remote close fails or server is gone.
- Prevents spurious “Session ID missing” errors when servers legitimately omit a session id.


phanikpemmaraju:
looking for reviewers to review the same.

phanikpemmaraju:
@tzolov @chemicL can you take a look at the same ?

phanikpemmaraju:
@Kehrlann can you review the MR

andersenleo:
This handles a DELETE that errors. A DELETE that never answers still does not reach `dispose()`, because `.then(...)` waits on it and no error is signalled for `onErrorResume` to catch. I put the detail and a reproduction in #547.

A timeout on the `onClose` publisher inside this same `defer` would cover both cases, if you would rather extend this PR than have a second one.

phanikpemmaraju:
@andersenleo 

Good spot. Much appreciated for highlighting this—you're completely right about the hanging/incomplete DELETE response bypassing .then(...) and stranding open connections.

I have updated DefaultMcpTransportSession#closeGracefully() to address both the error and the hanging scenarios directly within the operator pipeline:

- Replaced the .then(...) chaining with .doFinally(...). This ensures this.openConnections.dispose() executes unconditionally across all reactive terminal signals (ON_COMPLETE, ON_ERROR, and downstream CANCEL).

- Added a .timeout(timeout) directly on the publisher returned by this.onClose.apply(...). If the server stalls or never completes the response body, the timeout triggers, cancels the upstream exchange (which prompts the JDK HttpClient / Netty to tear down the socket), and transitions straight into doFinally to dispose the connections.

- Wrapped this.onClose.apply(id) in a try-catch so that if the function throws before returning a publisher, this.openConnections.dispose() is still invoked before propagating the exception.

FYI @Kehrlann

Issue #492, opened 2025-08-14
Should McpTransportContextExtractor contain default methods for protocol version, lastEventId and sessionId


sdelamo:
I am defining my `McpTransportContextExtractor` implementation as: 

```
@Internal
@Singleton
class DefaultMcpTransportContextExtractor implements McpTransportContextExtractor<HttpRequest<?>> {
    public static final String HTTP_HEADER_MCP_PROTOCOL_VERSION = "MCP-Protocol-Version";
    public static final String DEFAULT_PROTOCOL_VERSION = "2025-03-26";
    public static final String HTTP_HEADER_MCP_SESSION_ID = "Mcp-Session-Id";
    public static final String HTTP_HEADER_DEFAULT_LAST_EVENT_ID = "Last-Event-ID";

    @Override
    public McpTransportContext extract(HttpRequest<?> request, McpTransportContext transportContext) {
        transportContext.put(HTTP_HEADER_MCP_PROTOCOL_VERSION,
            request.getHeaders().get(HTTP_HEADER_MCP_PROTOCOL_VERSION, String.class)
                .orElse(DEFAULT_PROTOCOL_VERSION));
        request.getHeaders().get(HTTP_HEADER_MCP_SESSION_ID, String.class)
            .ifPresent(v -> transportContext.put(HTTP_HEADER_MCP_SESSION_ID, v));
        request.getHeaders().get(HTTP_HEADER_DEFAULT_LAST_EVENT_ID, String.class)
            .ifPresent(v -> transportContext.put(HTTP_HEADER_DEFAULT_LAST_EVENT_ID, v));
        return transportContext;
    }
}

```

But given that the specification defines session id, last event and protocol version. I think it would be good if the `McpTransportContextExtractor` contained default methods: 

```java
default String getSessionId() {
   return null;
}

default String getLastEventId() {
   return null;
}

default String getProtocolVersion() {
   return "2025-03-26";
}
```

I can submit a PR if you think it is a good idea. With this change the TransportContext will have a clear API for those spec defined concepts.

PR #570, opened 2025-09-24
feature Add methods to McpTransportContext

sdelamo:
## Motivation and Context
This change adds methods to McpTransportContext: `protocolVersion`, `lastEventId`, `sessionId`, and `principal`. It also provides an implementation of Context extract for Servlet requests, which uses the HTTP Headers defined in the specification.

It aims to provide some structure to the `McpTransportContext` API.

## How Has This Been Tested?
Yes, the change includes tests.

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/McpTransportContextExtractor.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/McpTransportContextExtractor.java
@@ -6,6 +6,8 @@ package io.modelcontextprotocol.server;
 
 import io.modelcontextprotocol.common.McpTransportContext;
 
+import java.util.Optional;
+
 /**
  * The contract for extracting metadata from a generic transport request of type
  * {@link T}.
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/servlet/HttpServletRequestMcpTransportContextExtractor.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/servlet/HttpServletRequestMcpTransportContextExtractor.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ */
+package io.modelcontextprotocol.server.servlet;
+
+import io.modelcontextprotocol.common.McpTransportContext;
+import io.modelcontextprotocol.server.McpTransportContextExtractor;
+import io.modelcontextprotocol.spec.ProtocolVersions;
+import jakarta.servlet.http.HttpServletRequest;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * {@link McpTransportContextExtractor} implementation for {@link HttpServletRequest}.
+ */
+public class HttpServletRequestMcpTransportContextExtractor
+		implements McpTransportContextExtractor<HttpServletRequest> {
+
+	@Override
+	public McpTransportContext extract(HttpServletRequest request) {
+		return McpTransportContext.create(metadata(request));
+	}
+
+	/**
+	 * @param request Servlet Request
+	 * @return Extracts Map for MCP Transport Context
+	 */
+	protected Map<String, Object> metadata(HttpServletRequest request) {
+		Map<String, Object> metadata = new HashMap<>();
+		metadata.put(io.modelcontextprotocol.spec.HttpHeaders.PROTOCOL_VERSION,
+				Optional.ofNullable(request.getHeader(io.modelcontextprotocol.spec.HttpHeaders.PROTOCOL_VERSION))
+					.orElse(ProtocolVersions.MCP_2025_03_26));
+		Optional.ofNullable(request.getHeader(io.modelcontextprotocol.spec.HttpHeaders.MCP_SESSION_ID))
+			.ifPresent(v -> metadata.put(io.modelcontextprotocol.spec.HttpHeaders.MCP_SESSION_ID, v));
+		Optional.ofNullable(request.getHeader(io.modelcontextprotocol.spec.HttpHeaders.LAST_EVENT_ID))
+			.ifPresent(v -> metadata.put(io.modelcontextprotocol.spec.HttpHeaders.LAST_EVENT_ID, v));
+		return metadata;
+	}
+
+}
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/servlet/package-info.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/servlet/package-info.java
@@ -1,0 +1,7 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ */
+/**
+ * Classes related with servlet support.
+ */
+package io.modelcontextprotocol.server.servlet;
--- a/mcp-core/src/test/java/io/modelcontextprotocol/common/DefaultMcpTransportContextTest.java
+++ b/mcp-core/src/test/java/io/modelcontextprotocol/common/DefaultMcpTransportContextTest.java
@@ -1,0 +1,53 @@
+package io.modelcontextprotocol.common;
+
+import io.modelcontextprotocol.spec.HttpHeaders;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+class DefaultMcpTransportContextTest {
+
+	@Test
+	void protocolVersionNotPresent() {
+		var ctx = new DefaultMcpTransportContext(Collections.emptyMap());
+		assertFalse(ctx.protocolVersion().isPresent());
+	}
+
+	@Test
+	void sessionIdNotPresent() {
+		var ctx = new DefaultMcpTransportContext(Collections.emptyMap());
+		assertFalse(ctx.sessionId().isPresent());
+	}
+
+	@Test
+	void lastEventIdNotPresent() {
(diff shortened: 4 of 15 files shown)


mcp-core/src/main/java/io/modelcontextprotocol/server/McpTransportContextExtractor.java
```diff
@@ -6,6 +6,8 @@
 
 import io.modelcontextprotocol.common.McpTransportContext;
 
+import java.util.Optional;
+
```
sdelamo:
```suggestion
```

mcp-core/src/main/java/io/modelcontextprotocol/server/servlet/HttpServletRequestMcpTransportContextExtractor.java
```diff
@@ -0,0 +1,42 @@
...
+	public McpTransportContext extract(HttpServletRequest request) {
+		return McpTransportContext.create(metadata(request));
+	}
+
+	/**
+	 * @param request Servlet Request
+	 * @return Extracts Map for MCP Transport Context
+	 */
+	protected Map<String, Object> metadata(HttpServletRequest request) {
+		Map<String, Object> metadata = new HashMap<>();
```
graemerocher: might be worth passing a size to this hash map

sdelamo: I added https://github.com/modelcontextprotocol/java-sdk/pull/570/commits/f02ec6f2674c8bd1c3f3b0f51552ad2cc164ff4f

mcp-core/src/main/java/io/modelcontextprotocol/server/servlet/package-info.java
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
```
Kehrlann:
```suggestion
 * Copyright 2025-2025 the original author or authors.
```

mcp-core/src/test/java/io/modelcontextprotocol/common/DefaultMcpTransportContextTest.java
```diff
@@ -0,0 +1,53 @@
+package io.modelcontextprotocol.common;
```
Kehrlann, line 5: Please add a header.

mcp-core/src/test/java/io/modelcontextprotocol/server/servlet/HttpServletRequestMcpTransportContextExtractorTests.java
```diff
@@ -0,0 +1,81 @@
+package io.modelcontextprotocol.server.servlet;
```
Kehrlann, line 5: Please add a header.

mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/ServerRequestMcpTransportContextExtractor.java
```diff
@@ -0,0 +1,5 @@
+package io.modelcontextprotocol.server.transport;
```
Kehrlann: Please add a header.

mcp-core/src/main/java/io/modelcontextprotocol/server/servlet/HttpServletRequestMcpTransportContextExtractor.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
```
Kehrlann:
```suggestion
 * Copyright 2025-2025 the original author or authors.
```

Kehrlann (commented):
I see how that is useful on the server side.
However, initially, the `McpTransportContext` was introduced as a very thin, immutable data object (see its [javadoc](https://github.com/modelcontextprotocol/java-sdk/blob/7f16cd0b9dc72f5adc3358a903bebb0f909dda3e/mcp-core/src/main/java/io/modelcontextprotocol/common/McpTransportContext.java#L10-L16)):

```java
/**
 * Context associated with the transport layer. It allows to add transport-level metadata
 * for use further down the line. Specifically, it can be beneficial to extract HTTP
 * request metadata for use in MCP feature implementations.
 *
 * @author Dariusz Jędrzejczyk
 */
```

As such, it is used both in the server transports, to capture HTTP request information; as well as the client side to capture "context" information (notably thread-locals), and make that available inside the reactor context. See [McpSyncClient](https://github.com/modelcontextprotocol/java-sdk/blob/7f16cd0b9dc72f5adc3358a903bebb0f909dda3e/mcp-core/src/main/java/io/modelcontextprotocol/client/McpSyncClient.java#L397-L405) and then subsequent usages in transports (e.g. [HttpClientStreamableHttpTransport](https://github.com/modelcontextprotocol/java-sdk/blob/7f16cd0b9dc72f5adc3358a903bebb0f909dda3e/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java#L253-L254)).

On the client side, all these new fields would be null. On the server side, when using stateless transport, I think the `sessionId` and `lastEventId` are also always null. Maybe the core `McpTransportContext` interface is not good candidate to add those fields, at least not in the current shape.

On the Spring side, we indeed need the `principal` when authorization is in use, but that's the only element we would be using ; we have extra needs (servlet request & response objects on the client side).

We have not provided `McpTransportContextExtractor` in the SDK for now, apart from your previous request #492, there seems to be no demand for this so far.

Should we postpone this until we have a clearer view of the community needs?

graemerocher:
perhaps what can be done is a sub-interface of `McpTransportContext` called `McpServerTransportContext` for the server use cases?

sdelamo:
Sorry, but I don't understand the pushback.

The methods I added are transport concepts - [protocol version ](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#protocol-version-header), [session id](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management), [Last event Id](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#resumability-and-redelivery). Should not transport concepts be represented in Transport Context (`McpTransportContext`). 

The same with `Principal`. Should the context of the transport not contain the authenticated user, if any? I don't see any coupling, as using `Principal` does not require any extra dependency. 

The API return types are optional, and the API methods are default methods with empty options. 

As an implementer of the context API you don't have to do anything. If you use the context in the client and it doesn't apply to you, don't use these methods. They don't require extra work on the client side.

as @Kehrlann pointed out in the javadocs of `McpTransportContext`:

> Specifically, it can be beneficial to extract HTTP request metadata for use in MCP feature implementations.

That it is exactly what I am doing in `HttpServletRequestMcpTransportContextExtractor`. 

If we don't add methods to the `McpTransportContext` API, then a user will have to guess the keys of the map. How will a user check whether an authenticated user can access a resource template by checking the context? Debugging the context to see where the security is saved?

sdelamo:
I add a commit which hopefully illustrates the benefit of this PR 

https://github.com/modelcontextprotocol/java-sdk/pull/570/commits/24098f2e7026cc8eef3abcedcf158a29e94fd791

Now, whether you use a `jakarta.servlet.http.HttpServletRequest` or a springs: `org.springframework.web.servlet.function.ServerRequest` or `org.springframework.web.reactive.function.server.ServerRequest`. You create an `McpTransportContext` parsing the request the same way.

chemicL:
Thinking about it a bit more, there are further concerns.
`McpTransportContext` is a wrapper around something coming from "a transport", but it's not tied to a particular one.

You can implement, say, a tool. And in your implementation you can get the `McpAsyncServerExchange` for instance, which holds the `McpTransportContext` (or simply the context in the case of stateless servers). The thing is that this tool does not know and should not know whether it is executed in the context of an HTTP request or as a local process via STDIO pipes. The transport concern is a layer away.

Even in the HTTP context, you are suggesting we add Streamable HTTP concepts to the generic transport context, but e.g. Last-Event-Id does not belong to the legacy MCP SSE spec (it is not specified in the context of MCP despite being part of the generic SSE specification).

Not to mention these are probably of no use on the client side.

What I'm trying to say is that in order to work with a particular transport, you can create the abstraction about what the tool might need, e.g. an interface called `ServerRequestContext`. That `ServerRequestContext` can expose logical notions such as `isCallerAuthorizedToXYZ(McpTransportContext)` and depending on the transport it can decide. Then the tool is decoupled from the particular transport, but the implementation of the tool can be coupled to this particular `ServerRequestContext` implementation.

On the transport provider side, a companion `StreamableHttpServerRequestContext` interface can expose the `McpTransportContext fill(HttpRequest)` method knowing what exactly to extract.

As the instantiator of the particular transport implementation, e.g. `HttpServletStreamableServerTransportProvider`, you can both inject the relevant extractor (using `StreamableHttpServerRequestContext`) that puts the relevant data under well known keys, but can also inject the reader of that data from the `McpTransportContext` into the tool implementation so that the implementation does not need to know the actual keys but relies on the logical thing that abstracts the transport away.

Let's discuss.

---

Regarding `Principal`: by coupling I do not mean needing a 3rd party dependency library, but just logical coupling of unrelated concepts (https://en.wikipedia.org/wiki/Interface_segregation_principle).

Kehrlann:
This hasn't been active in a while, is this still relevant?

Kehrlann:
Inactive, closing.

PR #588, opened 2025-10-01
Client: handle server responses with Content-Length: 0

### LINKED ISSUE #582: Can't invoke Streamable HTTP MCP server implemented with Micronaut MCP
I implemented a Streamable HTTP MCP server with the Micronaut MCP module:
https://github.com/glaforge/mn-mcp-server

It exposes a couple methods:
* `current-moon-phase()`: to get the current phase of the moon
* `moon-phase-at-date("2025-09-28")`: to get the phase at a certain date

You can call that server here with the MCP Inspector for testing:
https://mn-mcp-server-1029513523185.europe-west1.run.app/mcp/
The MCP inspector is totally fine with it, no error, I can list all the tools, call all the methods, etc.

I tried with the LangChain4j MCP module as well (which is a reimplementation of the protocol), with the following code:
https://gist.github.com/glaforge/809ccbad223886b0db03e679adcb7e85
And it also works totally fine with my server.

However, when I can from the Agent Development Kit for Java (ADK), it fails.
As ADK is using the official MCP SDK, I tried to reproduce the problem with the SDK directly, to remove ADK from the equation, and it fails in the same way.

I reproduced the problem with the following code:
https://gist.github.com/glaforge/741edcad883b27268ab10249400ce6d6

I'm seeing this stacktrace (same as within ADK for Java):
https://gist.github.com/glaforge/afd1194ece6bc909b091a946c030dc27

i.e.
```
io.modelcontextprotocol.spec.McpTransportException: Invalid SSE response. Status code: 405 Line: {"message":"Method Not Allowed","_links":{"self":{"href":"/mcp","templated":false}},"_embedded":{"errors":[{"message":"Method [GET] not allowed for URI [/mcp]. Allowed methods: [POST]"}]}}
	at io.modelcontextprotocol.client.transport.ResponseSubscribers$SseLineSubscriber.hookOnNext(ResponseSubscribers.java:185)
	at io.modelcontextprotocol.client.transport.ResponseSubscribers$SseLineSubscriber.hookOnNext(ResponseSubscribers.java:72)
...
21:04:44.389 [HttpClient-1-Worker-0] WARN io.modelcontextprotocol.client.transport.HttpClientStreamableHttpTransport -- Unknown media type text/html returned for POST in session [missing_session_id]
21:04:44.389 [HttpClient-1-Worker-0] WARN io.modelcontextprotocol.client.LifecycleInitializer -- Handling exception
java.lang.RuntimeException: Unknown media type returned: text/html
	at io.modelcontextprotocol.client.transport.HttpClientStreamableHttpTransport.lambda$sendMessage$27(HttpClientStreamableHttpTransport.java:522)
...
Caused by: java.lang.RuntimeException: Unknown media type returned: text/html
	at io.modelcontextprotocol.client.transport.HttpClientStreamableHttpTransport.lambda$sendMessage$27(HttpClientStreamableHttpTransport.java:522)
```

So I'm not sure if it's a problem on Micronaut's side or on the MCP SDK's side.
Maybe one is too strict with the spec, or not tolerant enough, I don't know.

But when I read the spec:
```
The server MUST provide a single HTTP endpoint path (hereafter referred to as the MCP endpoint) 
that supports both POST and GET methods. For example, this could be a URL like 
https://example.com/mcp.
```
```
The client MAY issue an HTTP GET to the MCP endpoint. This can be used to open an SSE stream, 
allowing the server to communicate to the client, without the client first sending data via HTTP POST.
```
```
The server MUST either return Content-Type: text/event-stream in response to this HTTP GET, 
or else return HTTP 405 Method Not Allowed, indicating that the server does not offer an SSE 
stream at this endpoint.
```
It seems that Micronaut does the right thing: if a client tries to do a GET to do an SSE connection, 
it gets a 405, which is the error Micronaut is returning.

Any idea what could be going wrong here @tzolov ?

/cc @sdelamo @graemerocher @Poggecci

Kehrlann:
- When the client sends `notification/initalized`, servers MUST respond with HTTP 202 and an empty body. Before this PR, we checked for the absence of a Content-Type header to know whether the body was empty.
- However, some servers will send an empty body with a Content-Type header, and that header may have an unsupported, default type such as `text/html` or `text/plain`.
- Now we we also use the Content-Length header to check for an empty body. This header is optional in HTTP/2, so we do not make it our primary mechanism for detecting empty bodies.
- Fixes #582 

## Types of changes

- [ ] Documentation update

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
@@ -246,7 +246,7 @@ public class HttpClientStreamableHttpTransport implements McpClientTransport {
 				}
 
 				var builder = requestBuilder.uri(uri)
-					.header("Accept", TEXT_EVENT_STREAM)
+					.header(HttpHeaders.ACCEPT, TEXT_EVENT_STREAM)
 					.header("Cache-Control", "no-cache")
 					.header(HttpHeaders.PROTOCOL_VERSION, MCP_PROTOCOL_VERSION)
 					.GET();
@@ -371,7 +371,7 @@ public class HttpClientStreamableHttpTransport implements McpClientTransport {
 
 		BodyHandler<Void> responseBodyHandler = responseInfo -> {
 
-			String contentType = responseInfo.headers().firstValue("Content-Type").orElse("").toLowerCase();
+			String contentType = responseInfo.headers().firstValue(HttpHeaders.CONTENT_TYPE).orElse("").toLowerCase();
 
 			if (contentType.contains(TEXT_EVENT_STREAM)) {
 				// For SSE streams, use line subscriber that returns Void
@@ -420,9 +420,9 @@ public class HttpClientStreamableHttpTransport implements McpClientTransport {
 				}
 
 				var builder = requestBuilder.uri(uri)
-					.header("Accept", APPLICATION_JSON + ", " + TEXT_EVENT_STREAM)
-					.header("Content-Type", APPLICATION_JSON)
-					.header("Cache-Control", "no-cache")
+					.header(HttpHeaders.ACCEPT, APPLICATION_JSON + ", " + TEXT_EVENT_STREAM)
+					.header(HttpHeaders.CONTENT_TYPE, APPLICATION_JSON)
+					.header(HttpHeaders.CACHE_CONTROL, "no-cache")
 					.header(HttpHeaders.PROTOCOL_VERSION, MCP_PROTOCOL_VERSION)
 					.POST(HttpRequest.BodyPublishers.ofString(jsonBody));
 				var transportContext = ctx.getOrDefault(McpTransportContext.KEY, McpTransportContext.EMPTY);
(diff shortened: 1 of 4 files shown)


mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
```diff
@@ -463,11 +463,15 @@ public Mono<Void> sendMessage(McpSchema.JSONRPCMessage sentMessage) {
 						.orElse("")
 						.toLowerCase();
 
-					if (contentType.isBlank()) {
+					String contentLength = responseEvent.responseInfo()
+						.headers()
+						.firstValue("Content-Length")
```
chemicL: Please use a constant

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
```diff
@@ -463,11 +463,15 @@ public Mono<Void> sendMessage(McpSchema.JSONRPCMessage sentMessage) {
...
+						.orElse(null);
+
+					if (contentType.isBlank() || "0".equals(contentLength)) {
 						logger.debug("No content type returned for POST in session {}", sessionRepresentation);
```
chemicL: This log is inaccurate now

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
Kehrlann: Happy to use a constant but it is inconsistent with the rest of the file (e.g. instances of `Content-Type`, a `Cache-Control` and an `Access`).

In that case I think we should standardize the whole transport. Agreed?

chemicL: Yep, let's add the headers to the `HttpHeaders` class since they are either part of the specification of the HTTP-based remote transports or are simply needed to make it work. `Last-Event-ID` is not MCP specific but is part of the MCP remote transport logic so there's no need to think of other places for the common header names.

chemicL (approved):
Thanks @Kehrlann !

PR #609, opened 2025-10-08
fix: use valid JPMS module name for mcp-core

### LINKED ISSUE #560: Invalid Automatic-Module-Name in mcp-core
Hey there,

Just upgraded to 0.13.0 and ran into an invalid module name.

Current value:

`Automatic-Module-Name:io.modelcontextprotocol.sdk.mcp-core`

Module names may not contain dashes, so this should be:

`Automatic-Module-Name:io.modelcontextprotocol.sdk.mcp.core`

bjorndarri:
## Summary
Fixes #560

- Fixed `Automatic-Module-Name` in `mcp-core/pom.xml` to use a valid JPMS module name
- Changed from `io.modelcontextprotocol.sdk.mcp-core` (invalid - contains dash) to `io.modelcontextprotocol.sdk.mcp.core`
- OSGI `Bundle-SymbolicName` kept as `io.modelcontextprotocol.sdk.mcp-core` for backward compatibility

## Problem
JPMS module names must be valid Java identifiers and cannot contain dashes. The previous automatic module name `io.modelcontextprotocol.sdk.mcp-core` prevented usage in modular Java applications.

## Solution
Hard-coded the `Automatic-Module-Name` to `io.modelcontextprotocol.sdk.mcp.core`. 

Alternative: Could use BND's substitution macro `${subst;${project.groupId}.${project.artifactId};-;.}` if dynamic generation is preferred by maintainers.

## Compatibility
OSGI's `Bundle-SymbolicName` remains unchanged with the dash (`mcp-core`) as OSGI allows dashes in symbolic names. Having different names for JPMS and OSGI is acceptable and common.

## Test plan
- [x] Built `mcp-core` module successfully
- [x] Verified manifest contains `Automatic-Module-Name: io.modelcontextprotocol.sdk.mcp.core`
- [x] Verified `Bundle-SymbolicName` remains `io.modelcontextprotocol.sdk.mcp-core`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mcp-core/pom.xml
+++ b/mcp-core/pom.xml
@@ -38,7 +38,7 @@
 								version:                ${versionmask;===;${version_cleanup;${project.version}}}
 								Bundle-SymbolicName:    ${project.groupId}.${project.artifactId}
 								Bundle-Version:         ${version}
-								Automatic-Module-Name:  ${project.groupId}.${project.artifactId}
+								Automatic-Module-Name:  io.modelcontextprotocol.sdk.mcp.core
 								Import-Package:         jakarta.*;resolution:=optional, \
 								                        *;
 								Export-Package:         io.modelcontextprotocol.*;version="${version}";-noimport:=true


vorburger:
> Alternative: Could use BND's substitution macro `${subst;${project.groupId}.${project.artifactId};-;.}` if dynamic generation is preferred by maintainers.

Oh, nice!! If you would like to change it to that here, personally I actually think that would be much nicer... (But I'm NOT a maintainer here!) Saying this mostly because I have a hunch that this `bnd-maven-plugin` really belongs into the root `pom.xml` instead of the `mcp-core/pom.xml`; but perhaps that's better left for a future follow-up PR.

bjorndarri:
I agree. I would personally be fine with this PR being dropped and this automatic module name handling being added to all modules, using substitution, with `bnd-maven-plugin` in the root, like you suggest. Let's see what happens here.

sunil-sunny:
Our team is using 0.11.3 version and we are blocked due to this issue. We could not upgrade the MCP SDK to higher versions.

sunil-sunny (approved):

PR #610, opened 2025-10-09
Fix HttpClient resource leak in HTTP transports

### LINKED ISSUE #620: HttpClient resource leak causes thread accumulation and memory exhaustion
**Bug description**

When using `HttpClientStreamableHttpTransport` and `HttpClientSseClientTransport`, the application experiences continuous accumulation of `HttpClient-xxxx-SelectorManager` threads that are never cleaned up, eventually leading to memory exhaustion and application instability.

The root cause is that each transport builder creates a new `HttpClient` instance via [`HttpClient.Builder.build()`](https://github.com/modelcontextprotocol/java-sdk/blob/19a8c00a475d38c6104887d9df1ab503b4a51839/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java#L772), but these HttpClient instances are never properly closed when the transport shuts down. Each HttpClient spawns dedicated SelectorManager threads for network I/O operations, and since OpenJDK's HttpClient lacks public APIs for resource cleanup, these threads remain active indefinitely.

**Technical Details**: Tracing through `HttpClientStreamableHttpTransport#build()` reveals that each HttpClient instantiation triggers the creation of a SelectorManager thread in the OpenJDK 17 source code:

```java
SelectorManager(HttpClientImpl ref) throws IOException {
    super(null, null,
          "HttpClient-" + ref.id + "-SelectorManager",
          0, false);
    owner = ref;
    debug = ref.debug;
    debugtimeout = ref.debugtimeout;
    pool = ref.connectionPool();
    registrations = new ArrayList<>();
    deregistrations = new ArrayList<>();
    selector = Selector.open();
}
```

Source: [OpenJDK 17 HttpClientImpl.java](https://github.com/openjdk/jdk17/blob/4afbcaf55383ec2f5da53282a1547bac3d099e9d/src/java.net.http/share/classes/jdk/internal/net/http/HttpClientImpl.java#L711C1-L713C63)

This constructor shows how each HttpClient creates a uniquely named SelectorManager thread (`"HttpClient-" + ref.id + "-SelectorManager"`), which explains the observed thread naming pattern in production environments.

**Environment**

- **Spring MCP Version**: Latest (current main branch)
- **Java Version**: OpenJDK 17+ (tested on OpenJDK 17.0.14)
- **Operating System**: macOS 14.6.0 (also reproducible on Linux)
- **Transport Types**: `HttpClientStreamableHttpTransport`, `HttpClientSseClientTransport`
- **Related OpenJDK Issue**: [JDK-8308364](https://bugs.openjdk.org/browse/JDK-8308364)

**Steps to reproduce**

1. Create multiple `HttpClientStreamableHttpTransport` instances:
```java
for (int i = 0; i < 10; i++) {
    HttpClientStreamableHttpTransport transport = HttpClientStreamableHttpTransport
        .builder("http://localhost:8080")
        .build();
    
    McpSyncClient client = McpClient.sync(transport).build();
    client.initialize();
    client.closeGracefully(); // This doesn't clean up HttpClient threads
}
```

2. Monitor system threads using `jstack` or thread monitoring tools
3. Observe continuous growth of `HttpClient-xxxx-SelectorManager` threads
4. Repeat the process multiple times to see thread accumulation

**Expected behavior**

- When `transport.closeGracefully()` is called, all associated HttpClient resources should be cleaned up
- `HttpClient-xxxx-SelectorManager` threads should be terminated and not accumulate
- Memory usage should remain stable across multiple transport creation/destruction cycles
- No thread leakage should occur in long-running applications

```java
import io.modelcontextprotocol.client.McpClient;
import io.modelcontextprotocol.client.McpSyncClient;
import io.modelcontextprotocol.client.transport.HttpClientStreamableHttpTransport;

public class HttpClientLeakDemo {
    public static void main(String[] args) throws InterruptedException {
        System.out.println("Initial thread count: " + Thread.activeCount());
        
        // Create and close multiple transports
        for (int i = 0; i < 20; i++) {
            System.out.println("\n=== Creating transport " + (i + 1) + " ===");
            
            var transport = HttpClientSseClientTransport
                .builder("http://127.0.0.1:8002") // your sse mcp server base url
                .build();
            
            McpSyncClient client = McpClient.sync(transport)
                .requestTimeout(Duration.ofSeconds(5))
                .build();
            
            try {
                // This will fail but still creates the HttpClient
                client.initialize();
            } catch (Exception e) {
                System.out.println("Expected initialization failure: " + e.getMessage());
            }
            
            // Close the client - this should clean up resources but doesn't
            client.closeGracefully();
            
            System.out.println("Thread count after closing transport " + (i + 1) + ": " + Thread.activeCount());
            
            // List HttpClient threads
            Thread.getAllStackTraces().keySet().stream()
                .filter(t -> t.getName().contains("HttpClient") && t.getName().contains("SelectorManager"))
                .forEach(t -> System.out.println("  - " + t.getName()));
        }
        
        System.out.println("\nFinal thread count: " + Thread.activeCount());
        System.out.println("HttpClient SelectorManager threads are still running and will never be cleaned up!");
        
        // Force GC to confirm threads are not cleaned up
        while (true) {
            Thread.getAllStackTraces().keySet().stream()
                    .filter(t -> t.getName().contains("HttpClient") && t.getName().contains("SelectorManager"))
                    .forEach(t -> System.out.println("  - " + t.getName()));
            System.gc();
            Thread.sleep(1000);
            System.out.println("Thread count after GC: " + Thread.activeCount());
        }
    }
}
```

TsengX:
This pull request addresses the HttpClient resource leak issue described in #620.

Fixes #620

- Add support for external HttpClient instances in HttpClientStreamableHttpTransport and HttpClientSseClientTransport builders
- Implement proper HttpClient resource cleanup using reflection to close SelectorManager threads
- Add shouldCloseHttpClient flag to control resource management lifecycle
- Prevent thread leaks caused by unclosed HttpClient instances created via HttpClient.Builder.build()
- Add comprehensive tests for external HttpClient usage and resource cleanup

Fixes thread accumulation issue where HttpClient-xxxx-SelectorManager threads would continuously grow, leading to memory exhaustion. This addresses the underlying JDK issue documented in JDK-8308364.

Related: https://bugs.openjdk.org/browse/JDK-8308364

## Motivation and Context

The `HttpClientStreamableHttpTransport` and `HttpClientSseClientTransport` classes create new HttpClient instances via `HttpClient.Builder.build()` without properly managing their lifecycle. This leads to accumulation of `HttpClient-xxxx-SelectorManager` threads that are never cleaned up, eventually causing memory exhaustion.

Root cause analysis traced this to the OpenJDK 17 HttpClientImpl implementation ([JDK-8308364](https://bugs.openjdk.org/browse/JDK-8308364)), where each HttpClient spawns dedicated SelectorManager threads for network I/O but lacks public APIs for proper resource cleanup.

This change introduces two solutions:
1. **External HttpClient support** - Allow users to provide pre-created HttpClient instances for resource sharing across multiple transports
2. **Automatic resource cleanup** - Properly clean up internally-created HttpClient resources using reflection to access internal components

### References

- [JDK-8308364: HttpClient SelectorManager thread leak](https://bugs.openjdk.org/browse/JDK-8308364)
- [Related discussion: HttpClient Selector Manager Threads](https://forum.inductiveautomation.com/t/httpclient-selector-manager-threads/81664)

## How Has This Been Tested?

- Added tests for external HttpClient usage in `HttpClientStreamableHttpSyncClientTests` and `HttpSseMcpSyncClientTests`
- Updated existing transport test to verify proper resource cleanup during graceful shutdown
- Verified backward compatibility with existing test suite
- All existing tests continue to pass without modification

## Breaking Changes

No breaking changes. This is an additive change that maintains full API compatibility. Existing code will continue to work as before, but will now benefit from proper resource management and the option to use external HttpClient instances.

## Types of changes

- [ ] Documentation update

## Checklist


mcp-core/src/test/java/io/modelcontextprotocol/client/HttpClientStreamableHttpSyncClientTests.java
```diff
@@ -70,4 +73,39 @@ void customizesRequests() {
...
+
+		withClient(transport, syncSpec -> syncSpec, mcpSyncClient -> {
+			mcpSyncClient.initialize();
+			// Test should complete without errors
+		});
+
+		// External HttpClient should still be usable after transport closes
+		// (This is a basic test - in practice you'd verify the client is still
+		// functional)
+		assertThat(externalHttpClient).isNotNull();
```
Kehrlann: **[issue]** This test passes even if the client is closed.

mcp-core/src/test/java/io/modelcontextprotocol/client/HttpClientStreamableHttpSyncClientTests.java
```diff
@@ -70,4 +73,39 @@ void customizesRequests() {
...
+		// Create transport with internal HttpClient (default behavior)
+		McpClientTransport transport = HttpClientStreamableHttpTransport.builder(host).build();
+
+		withClient(transport, syncSpec -> syncSpec, mcpSyncClient -> {
+			mcpSyncClient.initialize();
+			// Test should complete and close gracefully
+		});
+
+		// This test verifies that internal HttpClient resources are cleaned up
+		// The actual verification happens during the graceful close process
```
Kehrlann: **[suggestion]** We should try to make an HTTP request here and ensure the client is closed.

TsengX:
> Hi @TsengX 👋
> 
> Thanks for the report, and for the proposed fix. A few notes on this PR:
> 
> 1. On the `HttpClientImpl`, from JDK 21 onwards, there is a `#close()` method, and we should try to use this is available:
>    ```java
>    for (var method : httpClient.getClass().getMethods()) {
>        if (method.getName().equals("close")) {
>            method.invoke(httpClient);
>            return;
>        }
>    }
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    We'd only fall back to reflection if there's no `close`.
> 2. Do you have a strong case for passing an HTTP client from the outside? It'd like to keep the API surface smaller, and not have conflicting `HttpClient.Builder` vs `HttpClient`.
> 3. If there's a strong case for the point above, I would prefer having a `Consumer<HttpClient>` in the builder for "on close" operations rather than a boolean. It would default to the "close client resources" implementation you proposed (or calling `close` if available).

Hi @Kehrlann,

Thank you for the detailed feedback! After reconsidering the design, I agree with your concerns about API complexity.

**Addressing your points:**

1. **JDK 21+ close() method**: Absolutely! I'll implement the fallback strategy you suggested.

2. **Rationale for external HttpClient support**: In enterprise AI Agent platform development scenarios, we often create 50+ (sometimes even 100+) MCP connections. Creating a shared `HttpClient` instance at application startup avoids duplicate resource initialization and connection pool setup, with proper cleanup only when the application lifecycle ends.

3. **Regarding Consumer<HttpClient> approach**: I agree this is an elegant design. However, I have concerns about your suggestion that "it would default to the 'close client resources' implementation." If this means the external `HttpClient` gets closed when `transport.closeGracefully()` is called, it could be problematic for the enterprise use case mentioned above. When one transport closes, it would shut down a globally shared `HttpClient` that may still be in use by other parts of the application. This is why I believe a boolean flag might be more appropriate to explicitly control whether the SDK should manage the HttpClient lifecycle.

**About Test improvements**: I'll add proper HTTP requests to verify client state and resource cleanup.

I'd appreciate any suggestions or feedback on this approach!

Kehrlann:
Thanks for the insights @TsengX . I'm OK with injecting a fully built `HttpClient`.

I propose we do the follwing:
1. Add an `.httpClient` method to both builders.
2. Remove `.clientBuilder`, `.customizeClient` and `.connectTimeout` methods entirely.
3. Add a `.onCloseClient(Consumer<HttpClient>)` which defaults to your reflection closing.

---

However, there's an unresolved issue here. The current `closeGracefully()` implementation actually creates a new session. So with the current implementation, it's possible to do the following:


```java
HttpClientStreamableHttpTransport transport = ...;
transport.sendMessage(...)
    .then(transport.closeGracefully())
    .then(transport.sendMessage(...));
```

And it works. With you proposed change, this would not work anymore ... But I think the proposed implementation respects the contract from `McpTransport` better. Let me discuss with the team.

TsengX:
> Thanks for the insights @TsengX . I'm OK with injecting a fully built `HttpClient`.
> 
> I propose we do the follwing:
> 
> 1. Add an `.httpClient` method to both builders.
> 2. Remove `.clientBuilder`, `.customizeClient` and `.connectTimeout` methods entirely.
> 3. Add a `.onCloseClient(Consumer<HttpClient>)` which defaults to your reflection closing.
> 
> However, there's an unresolved issue here. The current `closeGracefully()` implementation actually creates a new session. So with the current implementation, it's possible to do the following:
> 
> ```java
> HttpClientStreamableHttpTransport transport = ...;
> transport.sendMessage(...)
>     .then(transport.closeGracefully())
>     .then(transport.sendMessage(...));
> ```
> 
> And it works. With you proposed change, this would not work anymore ... But I think the proposed implementation respects the contract from `McpTransport` better. Let me discuss with the team.

Thanks for being patient @Kehrlann,

I've implemented the changes you suggested:

## Implemented Changes

- Remove `.clientBuilder`, `.customizeClient` and `.connectTimeout` methods entirely
- Add `.onCloseClient(Consumer<HttpClient>)` which defaults to reflection closing
- Replace shouldCloseHttpClient boolean with Consumer pattern

**API Design:**
- External HttpClient: onCloseClient = null (no cleanup)
- Internal HttpClient: onCloseClient = default reflection cleanup
- Custom cleanup: user-provided Consumer

**Solution implemented:**
- Use `sun.misc.Unsafe` to bypass module system temporarily
- Fallback strategy: JDK 21+ close() → JDK 17 SelectorManager reflection
- Comprehensive error handling for stability

## Test Updates

- Added proper HTTP requests to verify client functionality
- Enhanced validation for both external and internal HttpClient scenarios
- All tests pass successfully

## Discussion Point
After further investigation of JDK 21's `close()` method, I discovered that while the method exists, it's still on internal classes (`HttpClientFacade`) that are not accessible due to module restrictions. We can use sun.misc.Unsafe to bypass module system temporarily, and it works.

However, I think we need to focus on discussing whether this poses security risks. Is this approach acceptable for the SDK context, or do you have better suggestions?

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
```diff
@@ -177,6 +190,8 @@ public static class Builder {
 
 		private Duration connectTimeout = Duration.ofSeconds(10);
 
+		private Consumer<HttpClient> onCloseClient;
```
Kehrlann: Use a default value here.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
```diff
@@ -325,9 +331,23 @@ public Builder connectTimeout(Duration connectTimeout) {
...
 		 */
 		public HttpClientSseClientTransport build() {
-			HttpClient httpClient = this.clientBuilder.connectTimeout(this.connectTimeout).build();
+			HttpClient httpClient;
+			Consumer<HttpClient> closeHandler;
+
+			if (externalHttpClient != null) {
+				// Use external HttpClient, use custom close handler or no-op
+				httpClient = externalHttpClient;
+				closeHandler = onCloseClient; // null means no cleanup
```
Kehrlann: Please do not use null, use a no-op lambda instead, `client -> {}`.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
```diff
@@ -142,6 +152,7 @@ public class HttpClientSseClientTransport implements McpClientTransport {
 		this.httpClient = httpClient;
 		this.requestBuilder = requestBuilder;
 		this.httpRequestCustomizer = httpRequestCustomizer;
+		this.onCloseClient = onCloseClient;
```
Kehrlann, line 156: Assert that onCloseClient is not null.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransport.java
```diff
@@ -495,7 +515,58 @@ public Mono<Void> closeGracefully() {
...
+		try {
+			// unsafe
+			Class<?> UnsafeClass = Class.forName("sun.misc.Unsafe");
+			Field unsafeField = UnsafeClass.getDeclaredField("theUnsafe");
+			unsafeField.setAccessible(true);
+			Unsafe unsafe = (Unsafe) unsafeField.get(null);
+			Module ObjectModule = Object.class.getModule();
+			Class<HttpClientSseClientTransport> currentClass = HttpClientSseClientTransport.class;
+			long addr = unsafe.objectFieldOffset(Class.class.getDeclaredField("module"));
+			unsafe.getAndSetObject(currentClass, addr, ObjectModule);
```
Kehrlann: I don't think we should need unsafe for this. Is there another way?

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
```diff
@@ -219,14 +230,70 @@ private void handleException(Throwable t) {
 	public Mono<Void> closeGracefully() {
 		return Mono.defer(() -> {
 			logger.debug("Graceful close triggered");
+			DefaultMcpTransportSession currentSession = this.activeSession.getAndSet(createTransportSession());
+			Mono<Void> sessionClose = currentSession != null ? currentSession.closeGracefully() : Mono.empty();
+
+			if (onCloseClient != null) {
+				return sessionClose.then(Mono.fromRunnable(() -> onCloseClient.accept(httpClient)));
```
Kehrlann: merge error.

Kehrlann (commented):
Your code does not compile. The merge went wrong. Please review the merge, and then I'll do a second review.

Kehrlann:
@TsengX are you still interested in contributing this?

TsengX:
> @TsengX are you still interested in contributing this?您是否仍然对贡献这个感兴趣？

Yes, I'm currently exploring alternative approaches to address the thread leak issue and need to do more comprehensive testing before submitting an updated implementation.


I'll continue with additional testing to ensure the solution is robust and addresses all the concerns raised in your review. Once I have a solid implementation that meets the requirements, I'll update the PR accordingly.

Thanks for your patience!

TsengX:
Hi @Kehrlann ,

I've updated the implementation to address all your review comments.

**Changes Made**

Removed sun.misc.Unsafe and reflection code, and implemented ExecutorService-based resource management instead. Internal HttpClient instances now use a custom ExecutorService that is shut down during closeGracefully(), allowing GC to reclaim HttpClient resources. This approach uses standard Java APIs and works with Java 17+.

Also addressed your specific feedback: added default no-op lambda for onCloseClient, added Assert.notNull validation, and use client -> {} instead of null.

**Why ExecutorService Shutdown?**

The HttpClient in Java 11-20 does not implement AutoCloseable (this was only added in Java 21 per JDK-8304165). The recommended approach for earlier versions is to provide a custom ExecutorService via the builder and shut it down when done, as discussed in this Stack Overflow thread.

When the ExecutorService is shut down, the JVM can reclaim the HttpClient and its SelectorManager threads during the next GC cycle. This approach is clean, uses standard APIs, and doesn't require reflection or module system bypasses.

Kehrlann:
Hey @TsengX - sorry I never got back to this. 

It has drifted by a lot. Could you please fix conflicts?

TsengX:
> Hey @TsengX - sorry I never got back to this.
> 
> It has drifted by a lot. Could you please fix conflicts?

Thanks for following up.

After revisiting the problem, I ended up going with a different fix strategy than the one originally explored in this PR. Since the implementation changed substantially, I opened a new PR from the current `main` branch instead of rebasing this older branch.

The new approach aims to preserve the existing public API, stay compatible with Java 17, and avoid reflection / JDK-internal shutdown hooks, while fixing the same SelectorManager leak discussed here (#547 / #620).

The new PR is here: #904

So this PR is effectively superseded by #904, and I do not plan to continue updating #610.

Kehrlann:
Got it, thank you.

Closing, superseded by #904.

PR #632, opened 2025-10-23
Close McpTransportSession on transport close

Kehrlann:
Introduce a `ClosedMcpTransportSession`, representing what remains of a session after calling `.closeGracefully()` on a client transport.

## Motivation and Context

The `McpTransport#closeGracefully` docs say that it "releases any associated resources asynchronously". This was not the case, as a new session was created. The current implementation from streamable transports would instead recreate a session.

See also #610


tzolov (approved):
LGTM, thanks @Kehrlann

PR #654, opened 2025-11-06
chore: Make generateRequestId a protected method.

He-Pin:
## Motivation and Context
Control how the client session ID is generated.

## Types of changes

- [ ] Documentation update

## Checklist


Kehrlann:
@He-Pin What's the use-case for this? What are you trying to accomplish?

He-Pin:
we are building an mcp hub, we are encoding informations inside the sessionId, otherwise we have to change it with a varhandle

Kehrlann:
What kind of info are you encoding?

This PR is modifying the request ID generation scheme within a given session, not the session ID itself. As I understand it, the request ID is for request-response reconciliation within a given session. So I guess you need client-related info in the request ID ... is there are reason to use the request ID rather than `meta` ?

He-Pin:
Oh, I finally recovered my memory, I was think that's the session id, will close and resubmit a new pr. thanks

PR #655, opened 2025-11-06
Enable request timeout when handling requests

mcebanupgrade:
## Motivation and Context
Looks like request timeout functionality was meant to be offerred, but the `requestTimeout` parma was never hooked up to actually enforce the timeout.

## How Has This Been Tested?
Unit tests + manual tests with MCP Inspector
image

## Breaking Changes
This is backwards compatible, but some users may get surprises since previously configured request timeout was never enforced up until now.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/DefaultMcpStatelessServerHandler.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/DefaultMcpStatelessServerHandler.java
@@ -10,7 +10,9 @@ import io.modelcontextprotocol.spec.McpSchema;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
 
+import java.time.Duration;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


cbfaucherupgrade (approved):

Kehrlann:
I think there's a misconception. 

In a regular async, non-stateless server, `requestTimeout` is used for requests from the server to the client, i.e. "list roots", "elicitation", "sampling" and "ping". Stateless does not support any of this, so it has, in fact, no use for a `requestTimeout` property. I think we should remove it entirely ; let me check with the team.

BTW there's a javadoc issue with sync/async server builders, probably due to some copy-paste:

It currently is:

```
 * Sets the duration to wait for server responses before timing out requests. This
 * timeout applies to all requests made through the client, including tool calls,
 * resource access, and prompt operations.
```

It should be:

```
 * Sets the duration to wait for client responses before timing out requests. This
 * timeout applies to all requests made through the server, including ping,
 * list roots, elicitation and sampling.
```

Kehrlann:
Confirmed that we should remove this timeout in the first place.
Closing this,  but please feel free to reopen a PR to remove the timeout.

PR #657, opened 2025-11-08
fix: JSONRPCResponse JavaDoc

vorburger:
Fix the clearly wrong JavaDoc on the `McpSchema.JSONRPCResponse`.

## Motivation and Context

Triggered by reading it while debugging #605.

Complimentary to #607.

## How Has This Been Tested?

N/A, as it's just a JavaDoc fix.

## Breaking Changes

No, as it's just a JavaDoc fix.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [X] Documentation update

## Checklist
- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [X] I have added appropriate error handling
- [X] I have added or updated documentation as needed

PR #607, opened 2025-10-07
fix: Avoid NPE in McpClientSession

vorburger:
## Motivation and Context

Fixes NPE seen in the wild as described in #605.

## How Has This Been Tested?

As described in #605, unfortunately I'm unable to reproduce that NPE locally even myself, but I am consistently seeing this on a failing CI build. However I'm confident that this change improves the robustness of `McpClientSession` and will provide a much more clear and actionable error message when this occurs again next time.

## Breaking Changes

No.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally? No, see #606 - but that's unrelated to this change.
- [X] I have added appropriate error handling
- [X] I have added or updated documentation as needed

PR #658, opened 2025-11-09
fix: Validate JSONRPCResponse in constructor (fixes #605)

### LINKED ISSUE #605: NullPointerException in McpClientSession

**Bug description**

In https://github.com/enola-dev/enola/issues/1862, I've run into:

```
java.lang.NullPointerException: data
	at java.base/java.util.Objects.requireNonNull(Objects.java:259)
	at reactor.core.publisher.FluxHandle$HandleSubscriber.next(FluxHandle.java:251)
	at io.modelcontextprotocol.spec.McpClientSession.lambda$sendRequest$12(McpClientSession.java:280)
	at reactor.core.publisher.FluxHandle$HandleSubscriber.onNext(FluxHandle.java:113)
	at reactor.core.publisher.SerializedSubscriber.onNext(SerializedSubscriber.java:99)
	at reactor.core.publisher.FluxTimeout$TimeoutMainSubscriber.onNext(FluxTimeout.java:181)
	at reactor.core.publisher.MonoCreate$DefaultMonoSink.success(MonoCreate.java:176)
```

I guess that this means that [this](https://github.com/modelcontextprotocol/java-sdk/blob/19a8c00a475d38c6104887d9df1ab503b4a51839/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpClientSession.java#L280) is `null`?

**Environment**

0.14.1

**Steps to reproduce**

Unfortunately I'm unable to reproduce it locally even myself, but I am consistently seeing this on a failing CI build.

**Expected behavior**

Clear error message instead of NPE.

N/A

vorburger:
## Motivation and Context

Catch the root cause of #605 earlier if it ever happens again in the future.

## How Has This Been Tested?

Not really, but hopefully it's safe; I would welcome code review!

## Breaking Changes

Nope, this is an internal change for safety, which (hopefully) should have no impact.

## Types of changes
- [ ] Documentation update

## Checklist
- [X] I have added appropriate error handling
- [X] I have added or updated documentation as needed

## Additional context

@tzolov do you think this makes sense? I'm proposing this as a follow-up (NOT an alternative) to #607 (and #657), which IMHO makes sense either way as well.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -250,13 +250,17 @@ public final class McpSchema {
 		/**
 		 * Constructor that validates MCP-specific ID requirements. Unlike base JSON-RPC,
 		 * MCP requires that: (1) Requests MUST include a string or integer ID; (2) The ID
-		 * MUST NOT be null
+		 * MUST NOT be null.
 		 */
 		public JSONRPCRequest {
-			Assert.notNull(id, "MCP requests MUST include an ID - null IDs are not allowed");
-			Assert.isTrue(id instanceof String || id instanceof Integer || id instanceof Long,
-					"MCP requests MUST have an ID that is either a string or integer");
+			validateID(id);
 		}
+	}
+
+	private static void validateID(Object id) {
+		Assert.notNull(id, "MCP requests/responses MUST include an ID - null IDs are not allowed");
+		Assert.isTrue(id instanceof String || id instanceof Integer || id instanceof Long,
+				"MCP requests/responses MUST have an ID that is either a string or integer");
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


vorburger:
@Kehrlann @tzolov respectful friendly ping on this one?

Kehrlann:
Hey @vorburger 

Sorry it took us so long to get to this. The ID is not mandatory in Json RPC error responses ([spec](https://modelcontextprotocol.io/specification/2025-11-25/schema#jsonrpcerrorresponse)), so this could break parsing in valid cases.

I don't think users are supposed to create a `JSONRPCResponse` on their own, it's only done internally, either through direct object instantiation or through parsing with a guard ([ref](https://github.com/modelcontextprotocol/java-sdk/blob/26304a7a39f80311f4128ba8a94d2301bb0b3f95/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java#L214-L216)). So we have a good control on where this should happen, I don't think checking adds a lot of value.

If we wanted to do something like this, we'd rather do sub-types of `JSONRPCReponse`, but I don't think it would add a lot of value to the SDK, again, due to purely internal usage.

PR #669, opened 2025-11-17
Client transports: make #protocolVersions() configurable

Kehrlann:
The SDK does not support full version negotiation. However, for compatibility reasons, we want to make our client transports configurable so that users can set the range of "acceptable" spec versions


mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
```diff
@@ -105,20 +104,29 @@ public class WebClientStreamableHttpTransport implements McpClientTransport {
 
 	private final AtomicReference<Consumer<Throwable>> exceptionHandler = new AtomicReference<>();
 
+	private final List<String> supportedProtocolVersions;
+
+	private final String latestSupportedProtocolVersion;
```
tzolov, line 109: Perhaps we should call this `selectedProtocolVersion` or `negotiatedProtocolVersion`.  
After we implement the negotiation this value can change if the server request a lower version.

Kehrlann: The negotiated protocol version is not a property of the transport itself. I think it will be a property of the `io.modelcontextprotocol.client.LifecycleInitializer.Initialization`, and we'll pass it through the reactor context.

tzolov: i guess you are right. Anyway will think about this when we start the work. Will mere the PR as it is

PR #453, opened 2025-08-05
Rename (Async)HttpRequestCustomizer to Mcp(Async)HttpRequestCustomizer, add delegating implementation

Kehrlann:
## Motivation and Context
The name `(Async)HttpRequestCustomizer` is too broad when used in a broader project. We prefix it with `Mcp` to make it clearer in what context this should exist. Originally raised in: https://github.com/modelcontextprotocol/java-sdk/pull/388#issuecomment-3148568336

Additionally, users may want to apply mutliple customizers. We provide a convenient "delegating" implementation, that goes through a list of customizers in order. Originally raised in: https://github.com/spring-projects/spring-ai/pull/3994#discussion_r2252866936

## How Has This Been Tested?
n/a

## Breaking Changes
Yes, renaming:
- `HttpRequestCustomizer` -> `McpHttpRequestCustomizer`
-  `AsyncHttpRequestCustomizer` -> `McpAsyncHttpRequestCustomizer`

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #672, opened 2025-11-18
chore: Extract McpSchema to a dedicated module

He-Pin:
## Motivation and Context

This will help the implementations who want to handle everything theirself

## Types of changes

- [x] Documentation update

## Checklist


Kehrlann:
The code is MIT licensed. If an implementer only wants to keep the schema, they are free to use the Java class directly, provided they include the license.

We don't want to deal with the overhead of a separate module.

He-Pin:
But we already added a `json` module for this; this can help the kt and scala community too.

Kehrlann:
The reason we added `json` is to offer an extension point so that any JSON serialization library could be used with the SDK, and not just Jackson. See #453 

It's not for consumption outside of MCP, rather for having a "json" SPI

He-Pin:
There's also an advantage here: users can provide their own McpSchema implementation in a binary-compatible way, adding fields while reusing all the capabilities of the existing SDK. Currently, we cannot update the model independently when using mcp-core.

@Kehrlann, wdyt, this will enable users to iterate on the schema and implementation separately.

Kehrlann:
I don't think there is a lot of demand for this.

I'm going to close the PR, please open an issue instead. Please give detailed examples of what you're trying to achieve exactly. We'll see if the issue gains traction, and we'll sync about it.

PR #674, opened 2025-11-18
chore: Make SyncSpec and AsyncSpec extensible

He-Pin:
## Motivation and Context

We need to extend `McpSpec` to add more functions.

NOTE: I think we should try to make use of lombok's `@superBuilder` here, which can make our life easier, instead of a copy.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/client/McpClient.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/client/McpClient.java
@@ -121,7 +121,7 @@ public interface McpClient {
 	 * @throws IllegalArgumentException if transport is null
 	 */
 	static SyncSpec sync(McpClientTransport transport) {
-		return new SyncSpec(transport);
+		return new McpClient.SyncSpec(transport);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Kehrlann:
We tend to prefer composition over inheritance. Please document _what_ you are trying to achieve before proposing a change.

We are not planning to add Lombok to this codebase.

He-Pin:
Thanks. In our current implementation, we need to return our own McpAsyncClient implementation (which we've partially customized to support our scenario), but we also want users to maintain their previous usage habits. Since we currently cannot extend McpClient's AsyncSpec (which has a private constructor), we have to copy the AsyncSpec code and make some modifications. Although the license allows us to do this, it's extremely inconvenient for maintainability. Therefore, we hope to have this extension so that users with similar needs can better maintain their custom platform capabilities.

Randgalt:
This would be a bad direction for the project. There are better ways to allow custom clients than exposing the internals via inheritance.

Kehrlann:
Agreed, we want to avoid inheritance where possible, see discussion in #793

PR #677, opened 2025-11-19
WebMvcSseServerTransportProvider: do not include session ID in SSE event

Kehrlann:
Currently, the WebFlux SSE Server transport and WebMVC SSE Server transport do not align in behavior.
The WebMVC transport gets an ID generated an injected, instead of relying on the SessionFactory's generation mechanism.

In this PR, we put the session ID generation back into the factory. This means the "transport session" knows nothing about the session ID, so we remove the id from the SSE events and all the logging data.

### DIFF
--- a/mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcSseServerTransportProvider.java
+++ b/mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcSseServerTransportProvider.java
@@ -7,7 +7,6 @@ package io.modelcontextprotocol.server.transport;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.List;
-import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantLock;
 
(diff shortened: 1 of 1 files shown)


mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcSseServerTransportProvider.java
```diff
@@ -256,39 +255,31 @@ private ServerResponse handleSseConnection(ServerRequest request) {
...
-					logger.debug("SSE connection timed out for session: {}", sessionId);
-					sessions.remove(sessionId);
-				});
-
-				WebMvcMcpSessionTransport sessionTransport = new WebMvcMcpSessionTransport(sessionId, sseBuilder);
-				McpServerSession session = sessionFactory.create(sessionTransport);
-				this.sessions.put(sessionId, session);
+		return ServerResponse.sse(sseBuilder -> {
+			WebMvcMcpSessionTransport sessionTransport = new WebMvcMcpSessionTransport(sseBuilder);
+			McpServerSession session = sessionFactory.create(sessionTransport);
```
He-Pin, line 261: nice to have this.

chemicL (approved):
This is a useful improvement to the mess that existed there, thank you :)

He-Pin:
@Kehrlann @chemicL is it possible we extract servlet and mvc support to dedicated module?

Kehrlann:
@He-Pin I'm not sure I understand the ask. MVC is already in a separate module.

He-Pin:
@Kehrlann oops, wrong memory, what about `servlet` one

Kehrlann:
The idea is to have the basics, SDK+Jakarta base in the core. And you'll notice that `jakarta-servlet` is a provided dep, so we're not even bringing it in.

PR #649, opened 2025-10-31
chore: Spring 5 compatibility

He-Pin:
## Motivation and Context

https://github.com/modelcontextprotocol/java-sdk/issues/648

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #696, opened 2025-11-26
Revert "WebClientStreamableHttpTransport: use Spring-5 compatible met…hods (#649)"

tzolov:
Unfortunately the #649 breaks compatibility with Spring Framework 7.  

Spring Framework 7 compatibility takes priority over Spring Framework 5 support.

Remove deprecated rawStatusCode() API usage in WebClientStreamableHttpTransport
that breaks Spring Framework 7 compatibility. Replace with modern statusCode()
API methods including is2xxSuccessful() and isSameCodeAs().

### DIFF
--- a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
+++ b/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
@@ -313,7 +313,7 @@ public class WebClientStreamableHttpTransport implements McpClientTransport {
 
 					// The spec mentions only ACCEPTED, but the existing SDKs can return
 					// 200 OK for notifications
-					if (is2xx(response)) {
+					if (response.statusCode().is2xxSuccessful()) {
 						Optional<MediaType> contentType = response.headers().contentType();
 						long contentLength = response.headers().contentLength().orElse(-1);
 						// Existing SDKs consume notifications with no response body nor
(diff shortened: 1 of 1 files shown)


tzolov:
@He-Pin unfortunately we can't support SF5 and SF7 and support for later is obviously our priority.

He-Pin:
Thank you. Can we make these protected methods, then I think I can override them with subtyping

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebClientStreamableHttpTransport.java
```diff
@@ -425,8 +424,16 @@ private Flux<McpSchema.JSONRPCMessage> eventStream(McpTransportStream<Disposable
 		return Flux.from(sessionStream.consumeSseStream(idWithMessages));
 	}
 
+	private static boolean isNotFound(ClientResponse response) {
```
He-Pin, line 427: Please make these methods a protected class method, then I can override them via subtyping.

tzolov: @He-Pin if framework 5 is important, perhaps you can copy the WebClientStreamableHttpTransport.java and modify and build it locally for your needs.

He-Pin: Yes, but that will require me to diff every time there is a release :(

PR #739, opened 2025-12-31
fix: ensuring the automatic module name is a valid name by replacing the - to . in artifactId

themaherkhalil:
Currently the jar's automatic module name is invalid as it contains a - in the artifactId.  If you run:

```shell
➜  0.17.0 jar --describe-module --file mcp-core-0.17.0.jar
Unable to derive module descriptor for: mcp-core-0.17.0.jar
Automatic-Module-Name: io.modelcontextprotocol.sdk.mcp-core: Invalid module name: 'mcp-core' is not a Java identifier
```

This prevents other projects using modules to import the jar. 

The change is to replace the - with a .

After compiling the code to build the jar, running the command will produce:

```shell
➜  target git:(main) jar --describe-module --file mcp-core-0.18.0-SNAPSHOT.jar
No module descriptor found. Derived automatic module.

io.modelcontextprotocol.sdk.mcp.core@0.18.0-SNAPSHOT automatic
requires java.base mandated
contains io.modelcontextprotocol.client
contains io.modelcontextprotocol.client.transport
contains io.modelcontextprotocol.client.transport.customizer
contains io.modelcontextprotocol.common
contains io.modelcontextprotocol.json
contains io.modelcontextprotocol.json.schema
contains io.modelcontextprotocol.server
contains io.modelcontextprotocol.server.transport
contains io.modelcontextprotocol.spec
contains io.modelcontextprotocol.util
```

### DIFF
--- a/mcp-core/pom.xml
+++ b/mcp-core/pom.xml
@@ -35,14 +35,13 @@
 						<configuration>
 							<bnd><![CDATA[
 								Bundle-Name:            Bundle ${project.groupId} : ${project.artifactId}
-								version:                ${versionmask;===;${version_cleanup;${project.version}}}
 								Bundle-SymbolicName:    ${project.groupId}.${project.artifactId}
-								Bundle-Version:         ${version}
+								Bundle-Version:         ${project.version}
 								Automatic-Module-Name:  ${project.groupId}.${project.artifactId}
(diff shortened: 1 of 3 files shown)


gbrail:
+1 to this! It's not possible to use anything that depends on this library, like the Google ADK, with Java modules without a fix. Thanks for opening it!

Issue #562, opened 2025-09-19
mcp no longer works in OSGi


benken-parasoft:
**Bug description**
In version 0.12, there was a single "mcp" jar which happens to be a valid OSGi bundle having a valid OSGi manifest.  Its dependencies all happen to be OSGi bundles as well.  So, mcp and all its dependencies have been loading in OSGi without issue.

In version 0.13, the "mcp" jar has been replaced with "mcp-core" and "mcp-json-jackson2".  Although "mcp-core" is still an OSGi bundle, "mcp-json-jackson2" is not.

**Environment**
I have been using an Apache Felix Servlet Bridge setup but the OSGi container shouldn't matter.

**Steps to reproduce**
Building an McpServer from an OSGi container results in the following:
```
java.lang.IllegalStateException: No default JsonSchemaValidatorSupplier implementation found
 at io.modelcontextprotocol.json.schema.JsonSchemaInternal.lambda$createDefaultValidator$2(JsonSchemaInternal.java:66)
 at java.base/java.util.Optional.orElseThrow(Optional.java:403)
 at io.modelcontextprotocol.json.schema.JsonSchemaInternal.createDefaultValidator(JsonSchemaInternal.java:61)
 at io.modelcontextprotocol.json.schema.JsonSchemaInternal.getDefaultValidator(JsonSchemaInternal.java:30)
 at io.modelcontextprotocol.json.schema.JsonSchemaValidator.getDefault(JsonSchemaValidator.java:61)
 at io.modelcontextprotocol.server.McpServer$StreamableSyncSpecification.build(McpServer.java:848)
```

**Expected behavior**

The "mcp-json-jackson2" jar should have an OSGi manifest.  The mcp bundles should have proper "Require-Capability" and "Provide-Capability" headers to support the OSGi Service Loader mechanism, like what's implemented by Apache SPI Fly: https://aries.apache.org/documentation/modules/spi-fly.html#specconf

**Minimal Complete Reproducible example**

The original Servlet Bridge example is here:
https://github.com/apache/felix-dev/tree/8c10c79ff48b6a8938e9fc8088964a2facb42da1/http/samples/bridge

You would need to add your own OSGi bundle with a bundle Activator that starts an MCP server and registers its transport using the OSGi HTTP service.  Some detail: https://github.com/apache/felix-dev/blob/master/http/README.md#using-the-httpservice

**Workaround**

[...]

Issue #127, opened 2025-04-08
Consider moving the mcp-spring-* modules


chemicL:
Since the base module (`mcp`) can provide the basic functionality for:

* creating a STDIO-based server and client
* creating a JDK HttpClient-based SSE client
* creating a Servlet-based SSE server

It would probably be a good idea to not have framework-specific transport implementations inside the java-sdk.

In the case of `mcp-spring-*` modules, they would be moved to the spring-ai project and can be used as part of the Spring AI integration with the MCP Java SDK. The question remains whether there are existing users of these modules that do not rely on Spring AI itself. If there are, their input can be used to consider publishing these modules as independent artifacts as they currently are but under a different group.

Inside the `mcp` core module we can focus on providing the core protocol implementation and maintain these basic transport implementations. While they are quite verbose and require a lot of ceremony, they can provide the minimal set of functionalities to allow users to be framework-agnostic if they so choose.

noear:
That's a good IDEA!! We want mcp modules to focus on protocol description and keep implementation separate.

noear:
Suggest HttpClientSseClientTransport HttpServletSseServerTransportProvider, StdioServerTransportProvider independence for the module. The user can choose according to the freedom.


With these implementation adaptations out of the way, we expect the cmp module to support jdk8

noear:
Projects like solon and vert.x. They don't need java-ee, they don't need servlets and HttpClient... The mcp module integrates servlets and HttpClient, which is redundant and limits the jdk version

Aliaksie:
I think it makes sense to make more granular modules, for example 
- `mcp-spec`
- `mcp-core` (optional)
- `mcp-client`
- `mcp-server`

noear:
@Aliaksie Yes. Like langchain4j, might only need mcp-core.

Also, there is a real need for jdk8 support in mcp-core. After splitting, mcp can achieve this effect better. There can be adaptations for jdk8, as well as adaptations for jdk17 and jdk21.

chemicL:
https://github.com/spring-projects/spring-ai/pull/5423

chemicL:
#804

PR #742, opened 2026-01-12
Add Support for Jackson 3 and use it by default

filiphr:
The goal of this PR is to add support for Jackson 3.

## Motivation and Context
Jackson 3 is the latest version from Jackson and it is the next iteration of Jackson.

## How Has This Been Tested?
The existing tests were enough

## Breaking Changes

The Java MCP SDK now pulls in Jackson 3 instead of Jackson 2. This can be potentially a breaking change in case someone was using `io.modelcontextprotocol.json.jackson.JacksonMcpJsonMapper` then they would need to switch to `io.modelcontextprotocol.json.jackson3.JacksonMcpJsonMapper` and / or adjust the dependencies so that they pull `mcp-core` and `mcp-json-jackson3`

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-spring/mcp-spring-webflux/pom.xml
+++ b/mcp-spring/mcp-spring-webflux/pom.xml
@@ -24,7 +24,7 @@
 	<dependencies>
         <dependency>
             <groupId>io.modelcontextprotocol.sdk</groupId>
-            <artifactId>mcp-json-jackson3</artifactId>
+            <artifactId>mcp-json-jackson2</artifactId>
             <version>0.18.0-SNAPSHOT</version>
         </dependency>
 
--- a/mcp-spring/mcp-spring-webmvc/pom.xml
+++ b/mcp-spring/mcp-spring-webmvc/pom.xml
@@ -24,7 +24,7 @@
 	<dependencies>
         <dependency>
             <groupId>io.modelcontextprotocol.sdk</groupId>
-            <artifactId>mcp-json-jackson3</artifactId>
+            <artifactId>mcp-json-jackson2</artifactId>
             <version>0.18.0-SNAPSHOT</version>
         </dependency>
 


sdeleuze:
@filiphr I was about to submit such PR, so thanks for creating it.

I think I would prefer to have consistency between `mcp-json-jackson2` and `mcp-json-jackson3` package naming strategy (version-less or with version), see my related https://github.com/modelcontextprotocol/java-sdk/issues/562#issuecomment-3738789147 comment, but that would be a breaking change, so let see with the team what is the preferred solution.

filiphr:
Thanks for your comment @sdeleuze. I've read your comment https://github.com/modelcontextprotocol/java-sdk/issues/562#issuecomment-3738789147 and I've added an alternative idea in https://github.com/modelcontextprotocol/java-sdk/issues/562#issuecomment-3740375276. 

In any case, I stumbled onto this since I wanted to contribute Jackson 3 support for Spring AI, but that depends on this project, so I started here first.

sdeleuze:
@filiphr Thanks, just be aware I am already working on the Jackson 3 support for Spring AI, and have it pretty advanced locally, so no need to contribute it. That said, I will welcome your review on my upcoming PR.

sdeleuze:
After the discussion on https://github.com/modelcontextprotocol/java-sdk/issues/562, my proposal is the following:

0.18:
 - Introduce `mcp-json-jackson3` as implemented in this PR
 - In `mcp-json-jackson2`, deprecate classes for removal in `io.modelcontextprotocol.json.jackson`/`io.modelcontextprotocol.json.schema.jackson` and copy them to `io.modelcontextprotocol.json.jackson2`/`io.modelcontextprotocol.json.schema.jackson2` in non-deprecated form.

0.19:
- Remove deprecated classes in `io.modelcontextprotocol.json.jackson`/`io.modelcontextprotocol.json.schema.jackson` from `mcp-json-jackson2`

@filiphr @tzolov @chemicL @graemerocher @sdelamo Are you ok with that? If we are, this PR could be refined with the proposed `mcp-json-jackson2` changes for 0.18.

filiphr:
I'm OK with your proposal @sdeleuze. In an hour or two I can adjust this PR and the the changes in relation to `mcp-json-jackson2`.

---

>@filiphr Thanks, just be aware I am already working on the Jackson 3 support for Spring AI, and have it pretty advanced locally, so no need to contribute it. That said, I will welcome your review on my upcoming PR.

Thanks for letting me know @sdeleuze. I have not spend a lot of time on it anyways, I stumbled fast on the different dependencies. I would be happy to review your PR for Spring AI, feel free to ping me when you have it ready.

tzolov:
> After the discussion on #562, my proposal is the following:
> 
> 0.18:
> 
> * Introduce `mcp-json-jackson3` as implemented in this PR
> * In `mcp-json-jackson2`, deprecate classes for removal in `io.modelcontextprotocol.json.jackson`/`io.modelcontextprotocol.json.schema.jackson` and copy them to `io.modelcontextprotocol.json.jackson2`/`io.modelcontextprotocol.json.schema.jackson2` in non-deprecated form.
> 
> 0.19:
> 
> * Remove deprecated classes in `io.modelcontextprotocol.json.jackson`/`io.modelcontextprotocol.json.schema.jackson` from `mcp-json-jackson2`
> 
> @filiphr @tzolov @chemicL @graemerocher @sdelamo Are you ok with that? If we are, this PR could be refined with the proposed `mcp-json-jackson2` changes for 0.18.

@sdeleuze @filiphr 

This plan makes sense. Please go ahead

filiphr:
Thanks for the review. @sdeleuze @tzolov I have implemented the plan like we discussed.

mcp-json-jackson3/src/main/java/io/modelcontextprotocol/json/jackson3/JacksonMcpJsonMapperSupplier.java
```diff
@@ -0,0 +1,34 @@
...
+ * {@link JsonMapper}.
+ */
+public class JacksonMcpJsonMapperSupplier implements McpJsonMapperSupplier {
+
+	/**
+	 * Returns a new instance of {@link McpJsonMapper} that uses the Jackson library for
+	 * JSON serialization and deserialization.
+	 * <p>
+	 * The returned {@link McpJsonMapper} is backed by a new instance of
+	 * {@link tools.jackson.databind.json.JsonMapper}.
```
chemicL: It's not a new one, right? It's using the shared global instance. This javadoc can be improved to make a good note of that.

chemicL (dismissed):
Thanks. I added a tiny javadoc remark, probably worth rebasing the PR against main and fixing that one.
As a follow up, we should look into unifying the tests between modules as currently they're a copy.

sdelamo (approved):
it looks good to me.

mcp-json-jackson3/src/main/java/io/modelcontextprotocol/json/jackson3/JacksonMcpJsonMapperSupplier.java
filiphr: Good point @chemicL. I adjusted the javadoc

filiphr:
Thanks everyone for the review. I had to do some additional changes to make sure everything works with Jackson 3.

* Upgrade Spring Framework to 7.x. Otherwise, the webmvc / webflux support does not work with Jackson 3.
* In `io.modelcontextprotocol.json.jackson3.JacksonMcpJsonMapper` catch the `JacksonException` and throw `IOException`. Otherwise, the webflux tests and most likely some more were failing.  In Jackson 3 the base `JacksonException` no longer extends `IOException` and it's a runtime exception. I do not like this that much, the alternative would be to go through all the places that are calling the methods of `McpJsonMapper` and adjust them to capture `Exception` instead of `IOException`. Another alternative would be to not throw `IOException`, but rather have a dedicated `McpJsonException` and each underlying library can throw that if there is some error in the mapping.

>As a follow up, we should look into unifying the tests between modules as currently they're a copy.

I agree with you @chemicL it would be good to have some unified tests, maybe tests where all the different json mappers are tested. Btw, I also had some problems with some of the tests in `StdioMcpAsyncClientTests` and `StdioMcpSyncClientTests` with errors like:

```
java.lang.AssertionError: 
Expecting all elements of:
  [TextContent[annotations=null, text=MCP error -32602: Tool sampleLLM not found, meta=null]]
to satisfy given requirements, but these elements did not:

TextContent[annotations=null, text=MCP error -32602: Tool sampleLLM not found, meta=null]
error: java.lang.AssertionError: 
Expecting actual:
  "MCP error -32602: Tool sampleLLM not found"
to end with:
  "Goodbye, world!"
```

chemicL:
IMO we can't just bump spring-framework from 6.x to 7.x and render the mcp-spring-* modules unusable by anyone with spring-framework 6. I'll gather ideas and come back with feedback.

chemicL:
What I can suggest is the following:
* rollback mcp-spring/* modules to use spring-framework 6.2.x
* make jackson2 a test-only dependency for mcp-spring/* modules (potentially actually removing test-compile-time dependency on it if mocks are used instead of an actual `JacksonMcpJsonMapper` in the test for the builder that you needed to modify).

@filiphr Would this work to make the tests pass?

@sdeleuze would you be able to verify that with these changes the java-sdk can be used both with Spring Framework 6 in Spring AI 1.0 and Framework 7 in Spring AI 2.0?

As a next step, we can deprecate mcp-spring/* modules in MCP Java SDK and have the implementation for Spring in the Spring portfolio as suggested in https://github.com/modelcontextprotocol/java-sdk/issues/127

filiphr:
@chemicL thanks for pointing it out. The fact that it does not work with Spring 7 is actually a bug in the `mcp-spring` implementations. It relies on the underlying Jackson implementation and does not use the `McpJsonMapper`.

I've updated the PR to also fix the places in the `mcp-spring` implementations that were not using the `McpJsonMapper`. I actually thing it was a good exercise to run Jackson 3 with Spring 6 (which does not support Jackson 3 natively). 

I think the tests should be passing now.

chemicL:
Thank you for taking my feedback into account.

Please do follow the [contributing guidelines](https://github.com/modelcontextprotocol/java-sdk/blob/main/CONTRIBUTING.md#submitting-changes) - force pushing every change makes reviewing the incremental changes difficult.

Can you elaborate on why in the spring modules using the underlying Jackson implementation and not utilizing the `McpJsonMapper` is a bug? It works with Spring Framework 6 – what changes in Framework 7 that makes it break and forces us to explicitly invoke the mapping from an Object to the JSON String? The content negotiation mechanism in Spring would still perform a check and could do the mapping for us, so perhaps something else is at play?

filiphr:
>Please do follow the [contributing guidelines](https://github.com/modelcontextprotocol/java-sdk/blob/main/CONTRIBUTING.md#submitting-changes) - force pushing every change makes reviewing the incremental changes difficult.

Sorry about that. I didn't realize about the force pushing. I won't do that anymore.

>Can you elaborate on why in the spring modules using the underlying Jackson implementation and not utilizing the `McpJsonMapper` is a bug?

It is a bug because it relies on how the `WebClient` is configured and / or what is available on the classpath. If a user configures their `WebClient` to use Jackson 2 then not everything might as expected, as one can see with the tests failing when using Spring 6 with Jackson 3.

The 2 changes I did it more or less aligns with what is being done in other places in the `mcp-spring` modules e.g. 

* [`WebFluxSseClientTransport`](https://github.com/modelcontextprotocol/java-sdk/blob/f7a460fb505aeba239165f9bc4ef508a77ba102a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/client/transport/WebFluxSseClientTransport.java#L246)
* [`WebFluxSseServerTransportProvider`](https://github.com/modelcontextprotocol/java-sdk/blob/f7a460fb505aeba239165f9bc4ef508a77ba102a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java#L379)
* [`WebFluxStatelessServerTransport`](https://github.com/modelcontextprotocol/java-sdk/blob/f7a460fb505aeba239165f9bc4ef508a77ba102a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStatelessServerTransport.java#L114)
* [`WebFluxStreamableServerTransportProvider`](https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java#L248)
* [`WebFluxStreamableServerTransportProvider.WebFluxStreamableMcpSessionTransport`](https://github.com/modelcontextprotocol/java-sdk/blob/f7a460fb505aeba239165f9bc4ef508a77ba102a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxStreamableServerTransportProvider.java#L354)
* [`WebFluxSseServerTransportProvider.WebFluxMcpSessionTransport`](https://github.com/modelcontextprotocol/java-sdk/blob/f7a460fb505aeba239165f9bc4ef508a77ba102a/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java#L379)
* [`WebMvcSseServerTransportProvider`](https://github.com/modelcontextprotocol/java-sdk/blob/f7a460fb505aeba239165f9bc4ef508a77ba102a/mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcSseServerTransportProvider.java#L383)
* [`WebMvcStreamableServerTransportProvider.WebMvcStreamableMcpSessionTransport`](https://github.com/modelcontextprotocol/java-sdk/blob/f7a460fb505aeba239165f9bc4ef508a77ba102a/mcp-spring/mcp-spring-webmvc/src/main/java/io/modelcontextprotocol/server/transport/WebMvcStreamableServerTransportProvider.java#L520)

The approach fully decouples how the Spring WebClient is configured and whether or not the users have decided to use Jackson 2 or Jackson 3 in their global configuration.

chemicL:
Thanks for explaining, that does make sense to me. In such case, let's bring back jackson2 to the mcp-spring modules and I think we are good to go.

sdelamo:
> * rather have a dedicated `McpJsonException` and each underlying library can throw that if there is some error in the mapping.

A dedicated `McpJsonException` sounds good to me.

filiphr:
>A dedicated `McpJsonException` sounds good to me.

@sdelamo, Is it OK to do this in a subsequent PR or you want it as part of this one?

>Thanks for explaining, that does make sense to me. In such case, let's bring back jackson2 to the mcp-spring modules and I think we are good to go.

Thanks @chemicL. I'll do that adjustment. I'll also see if I can adjust the tests to run both with jackson2 and jackson3 present.

sdeleuze:
IMO ok to do it as part of this PR unless somebody objects.

filiphr:
I've added some more tests to the `mcp-json-*` modules and I also brought back `jackson2` in the mcp-spring modules.

I didn't yet do the `McpJsonException` as that would be a breaking change, since I would need to remove the `IOException` from the `McpJsonMapper`. If everyone agrees on `McpJsonException`, I am more than happy to contribute that independently.

> I'll also see if I can adjust the tests to run both with jackson2 and jackson3 present.

I also couldn't make this work since the schema validation is using the same artifact with different versions, so it is not possible to have both jackson2 and jackson3 in a single module. This could be further improved as part of #127

filiphr:
I am not sure why those tests are failing. The same tests are failing for me on main as well.

chemicL:
> I am not sure why those tests are failing. The same tests are failing for me on main as well.

I'll take care of it. The test setup is clunky and an update to the "everything server" broke our CI ;)

filiphr:
Gotcha, thanks. Please let me know if there is anything more that I can do for this.

chemicL:
@filiphr please rebase against `main`. I believe the integration tests setup should be fine now.

filiphr:
@chemicL, done. All tests are now passing locally. Thanks for the fix.

chemicL:
@filiphr thank you :)
@sdeleuze I am ok to merge this, please let me know once you confirm on the Spring AI side that your changes work fine with this PR, thanks!

sdeleuze:
Great, I will check that early next week then we will probably be able to merge it.

sdeleuze:
@tzolov @chemicL The changes look good from Spring AI Jackson 3 perpective, please merge this PR.

scottslewis:
I've just sync'd my fork of the mcp-java-sdk project with these merged changes  (1/26/2025 here).

I'm getting some compile failures in test classes:

mcp-core//mcp-core/src/test/java/io/modelcontextprotocol/spec/McpSchemaTests.java

import tools.jackson.databind.exc.InvalidTypeIdException;

/mcp-spring-webflux/src/test/java/io/modelcontextprotocol/client/transport/WebFluxSseClientTransportTests.java

import tools.jackson.databind.json.JsonMapper;

Where is the library with these (and other) classes in tools.jackson.databind.exc and .json?   

Strange thing is that maven compile seems to find it, but I can't figure out how it's finding these classes.

chemicL:
Thanks @scottslewis for the catch.

As I understand, `mcp-core//mcp-core/src/test/java/io/modelcontextprotocol/spec/McpSchemaTests.java` sees the `tools.jackson.databind.exc.InvalidTypeIdException` because the `pom.xml` declares the dependency on:
```
<dependency>
            <groupId>io.modelcontextprotocol.sdk</groupId>
            <artifactId>mcp-json-jackson3</artifactId>
            <version>0.18.0-SNAPSHOT</version>
            <scope>test</scope>
        </dependency>
```

which transitively brings in
```
<dependency>
            <groupId>tools.jackson.core</groupId>
            <artifactId>jackson-databind</artifactId>
            <version>${jackson3.version}</version>
        </dependency>
```

I don't see that as a big issue though.

For the `mcp-spring-webflux` module, it has a dependency on the `mcp` module which brings in `mcp-json-jackson3`, hence it also compiles. However, I think that is not what was intended. I'll aim to open a PR to address this.

@filiphr do you have any other insights?

chemicL:
My proposal to address the issues caused by adding Jackson 3 to the picture is in #760. Please share your thoughts.

filiphr:
@scottslewis if it compiles through maven then maybe your IDE is not properly configuring the dependencies. Not sure why it is not compiling for you.

I agree with your proposal @chemicL, it makes sense

scottslewis:
Thanks @filiphr and @chemicL ...thanks, I got the tools.* dependencies worked out now.  There is something strange going on wrt your maven config wrt m2e/eclipse, but I'm not going to be concerned with it out now.

Question:   If one wants to continue to use Jackson2 with the sdk are there any special build, test, or deployment procedures required?  ...other than using the jackson2 jar rather than jackson3 jar at runtime?

filiphr:
>Question: If one wants to continue to use Jackson2 with the sdk are there any special build, test, or deployment procedures required? ...other than using the jackson2 jar rather than jackson3 jar at runtime?

There are no special procedures required. You should only make sure that you have either `mcp-json-jackson2` or `mcp-json-jackson3` available.

PR #682, opened 2025-11-20
Refactor of json + json schema validation api into mcp-core

scottslewis:
into mcp-core.

Moved classes from mcp-json jar into new packages io.modelcontextprotocol.json, io.modelcontextprotocol.json.schema, and io.modelcontextprotocol.json.internal.  Updated dependencies on moved classes in mcp-core, mcp-jackson2 and other maven modules.

## Motivation and Context

This provides a fix for #612 and removed mcp-core dependencies on mcp-json.   After these refactoring, mcp-json project/jar is not needed and only mcp-core and mcp-jackson2 have to be deployed to have a functioning sdk.

## How Has This Been Tested?

Have run test suite locally.  Some of the mcp-core tests are failing...apparently because of the environment setup for docker...for example:

[ERROR]   HttpSseMcpAsyncClientLostConnectionTests.testPingWithExactExceptionType ┬╗ ExceptionInInitializer
[ERROR]   HttpSseMcpAsyncClientTests.startContainer:41 ┬╗ IllegalState Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration

## Breaking Changes

As per #612, previously mcp-core had compile-time dependencies on mcp-json, meaning that mcp-core, mcp-json, and mcp-jackson2 had to be deployed together or nothing would work...even code that just used mcp-core.

## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X ] Breaking change (fix or feature that would cause existing functionality to change)
- [ X] Documentation update

## Checklist

- [X ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X ] My code follows the repository's style guidelines
- [X ] New and existing tests pass locally
- [X ] I have added appropriate error handling
- [ X] I have added or updated documentation as needed

Issue #612, opened 2025-10-09
Mcp-core has compile-time dependency on mcp-json


scottslewis:
mcp-core project has a compile-time dependency on mcp-json due to the location of these three classes:

/mcp-json/src/main/java/io/modelcontextprotocol/json/McpJsonMapper.java

/mcp-json/src/main/java/io/modelcontextprotocol/json/TypeRef.java

/mcp-json/src/main/java/io/modelcontextprotocol/json/schema/JsonSchemaValidator.java

I would argue that since these classes are required at all times, it doesn't make sense for them to be in a separate jar...as mcp-core jar is useless without mcp-json jar also present.

I would suggest moving these three classes (at least) back to mcp-core (e.g. in package io.modelcontextprotocol.sdk.util or io.modelcontextprotocol.sdk.json.   This would allow a clear separation between mapper and schema validator interfaces (in core) and their implementation (default or other).

tbohr-actico:
May I add another finding that is related to this issue?
Due to the specific bnd-maven-plugin configuration in mcp-core, the class files from mcp-json are assembled into the mcp-core jar. The situation you describe is therefore already evident, albeit in a very unexpected way.

scottslewis:
I've done an initial refactor in pr #682.   This pr moves the json  mapper + json schema validation api (without implementation) back into mcp-core, in new packages .json, .json.schema, and .json.internal.   With these changes mcp-core no longer has a compile time dependency on mcp-json nor mcp-jackson2.   Now that the api that was in mcp-json is in mcp-core packages, mcp-json module is not needed at all (and is removed from the maven build.

mcp-core requires mcp-jackson2 or a new implementation of the a McpJsonMapperSupplier and JsonSchemaValidatorSupplier.   IF not configured another way (e.g. via osgi declarative services or 'manually'), the java ServiceLoader is used to load the default McpJsonMapperSupplier and JsonSchemaValidatorSupplier.

I'm planning on adding the OSGi metadata to use an OSGi service impls of McpJsonMapperSupplier and JsonSchemaValidatorSupplier, so that the the osgi service registry can be used instead of the ServiceLoader in osgi environmemts.

scottslewis:
> I'm planning on adding the OSGi metadata to use an OSGi service impls of McpJsonMapperSupplier and JsonSchemaValidatorSupplier, so that the the osgi service registry can be used instead of the ServiceLoader in osgi environmemts.

[...]

PR #762, opened 2026-01-28
Moves mcp-json API back into mcp-core for simplified dependencies and support of osgi runtimes

scottslewis:
mcp-core so that mcp-json API is no longer needed and mcp-core has fewer dependencies.

See issue #612 and previous pr #682 

## Motivation and Context

Simplifies dependency structure, and reenables ability to use OSGi runtimes #562 

## How Has This Been Tested?

Have tested in OSGi environments running OSGi application.

## Breaking Changes

This eliminates the need to deploy mcp-json jar, meaning that users will only have to (minimally) deploy mcp-core and either mcp-json-jackson2 or mcp-json-jackson3

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X ] My code follows the repository's style guidelines
- [X ] I have added appropriate error handling
- [X ] I have added or updated documentation as needed

## Additional context

To deal with the testing dependencies issues as described here:  https://github.com/modelcontextprotocol/java-sdk/pull/682#issuecomment-3814459498

it may be necessary to add further changes/refactoring/moving of the test code (in mcp-core specifically).  I'm willing to do or contribute to this refactoring if needed, once consensus is achieved.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
@@ -57,7 +57,7 @@ public class McpServiceLoader<S extends Supplier<R>, R> {
 				Optional<?> sl = serviceLoad(this.supplierType);
 				if (sl.isEmpty()) {
 					throw new ServiceConfigurationError(
-							"No %s available for creating McpJsonMapper".format(this.supplierType.getSimpleName()));
+							"No %s available for creating McpJsonMapper".formatted(this.supplierType.getSimpleName()));
 				}
 				this.supplier = (S) sl.get();
 			}


mcp-core/pom.xml
```diff
@@ -97,21 +92,18 @@
...
 			<artifactId>spring-webmvc</artifactId>
 			<version>${springframework.version}</version>
 			<scope>test</scope>
 		</dependency>
 
+        <dependency>
+            <groupId>tools.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${jackson3.version}</version>
+        </dependency>
```
LucaButBoring, line 108: Are we missing a `<scope>test</scope>` here? Otherwise I believe this forces Jackson 3 across the board.

mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	@SuppressWarnings("unchecked")
+	public synchronized R getDefault() {
+		if (this.supplierResult == null) {
+			if (this.supplier == null) {
+				// Use serviceloader
+				Optional<?> sl = serviceLoad(this.supplierType);
+				if (sl.isEmpty()) {
+					throw new ServiceConfigurationError("No JsonMapperSupplier available for creating McpJsonMapper");
```
LucaButBoring: nit for debuggability
```suggestion
					throw new ServiceConfigurationError("No %s available for creating McpJsonMapper".format(this.supplierType.getSimpleName()));
```

mcp-core/src/main/java/io/modelcontextprotocol/json/McpJsonDefaults.java
```diff
@@ -0,0 +1,50 @@
...
+	void unsetMcpJsonMapperSupplier(McpJsonMapperSupplier supplier) {
+		mcpMapperServiceLoader.unsetSupplier(supplier);
+	}
+
+	public synchronized static McpJsonMapper getDefaultMcpJsonMapper() {
+		if (mcpMapperServiceLoader == null) {
+			new McpJsonDefaults();
+		}
+		return mcpMapperServiceLoader.getDefault();
+	}
```
LucaButBoring, line 65: Is there a reason this leverages ctor side-effects instead of using static initializers for `mcpMapperServiceLoader` and `mcpValidatorServiceLoader`?

mcp-core/src/main/java/io/modelcontextprotocol/json/internal/DefaultMcpJsonMapperSupplier.java
```diff
@@ -0,0 +1,40 @@
+package io.modelcontextprotocol.json.internal;
+
+import java.util.Optional;
+import java.util.ServiceConfigurationError;
+import java.util.ServiceLoader;
+
+import io.modelcontextprotocol.json.McpJsonMapper;
+import io.modelcontextprotocol.json.McpJsonMapperSupplier;
+
+public class DefaultMcpJsonMapperSupplier {
```
LucaButBoring: I think this and `DefaultMcpJsonSchemaValidatorSupplier` are unused?

mcp-core/pom.xml
scottslewis: Yes that's right thanks.  Fixed in commit:  https://github.com/modelcontextprotocol/java-sdk/pull/762/changes/bedf075f56c182477e4539753bfddc0c756a91fc

mcp-core/src/main/java/io/modelcontextprotocol/json/McpJsonDefaults.java
scottslewis: Generally speaking, static initializers cannot be extended or replaced in the future...and instances (via subclass constructors) can (e.g. if a new version of McpJsonDefaults is desired).

It's also a personal habit from osgi-land, where it's possible to have multiple instances (in different bundles) of the same class...again making static initializers for default overly 'static' :-).

The larger point here is that since osgi can have multiple versions of a single class (e.g. McpJsonDefaults) in the same runtime at the same time (different classloaders/bundles), static initializers are avoided whenever possible in favor of instances.

In osgi environments a new instance of McpJsonDefaults is created by scr, *and* that instance's setMcpJsonMapperSupplier, and setJsonSchemaValidatorSupplier are called by scr injection *before* getDefaultJsonMapper or getDefaultSchemaValidator *can* be called by any other code.  OSGi classloading and scr/service registry lifecycle contracts guarantee that.   

So in other words, the if (mcpMapperServiceLoader == null) always is false in osgi environments (because it's already been set via scr on startup)....and so the McpJsonDefaults constructor is only used in non-osgi/serviceloader environments.

Putting the initialization code in the constructor allows the same code to be used for initialization in both OSGI and non-OSGI environments.

mcp-core/src/main/java/io/modelcontextprotocol/json/internal/DefaultMcpJsonMapperSupplier.java
scottslewis: Yes thank you....my git mistake.  Removed in:  https://github.com/modelcontextprotocol/java-sdk/pull/762/changes/76ce84df4e7fdf1548a701777e914897429e3742

mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
scottslewis: applied/committed directly from suggestion.  Thanks.

filiphr:
@scottslewis, I'm not familiar with OSGi that much, but I have some general questions about the PR.

* Why not delete `mcp-json` in this PR as well?
* What is the difference between the mechanism that was being used before in the `McpJsonInternal` vs the one in this PR with the `McpJsonDefaults`? Of course taking into consideration the fact the move into `mcp-core`

scottslewis:
> @scottslewis, I'm not familiar with OSGi that much, but I have some general questions about the PR.
> 
> * Why not delete `mcp-json` in this PR as well?

I could have...but chose to leave it so that the pr was not as large.  If doing it in one step is preferable then that can be done.

> * What is the difference between the mechanism that was being used before in the `McpJsonInternal` vs the one in this PR with the `McpJsonDefaults`? Of course taking into consideration the fact the move into `mcp-core`

The impl of McpJsonDefaults is key to working (for json defaults) in non osgi and osgi environments.  In non osgi environments, the serviceloader is used to find and load jackson2 or jackson3 impl for defaults as done previously.

In osgi, the serviceloader does not work the same...because of classloader-per-bundle as per osgi issue #562.  In osgi,  mcpjsondefaults is created and configured on start by the osgi sevice registry using scr metadata in manifests.  

For clarity [here is the McpJsonDefaults scr metadata](https://github.com/modelcontextprotocol/java-sdk/pull/762/changes#diff-21b8cf3851b8939861110e2fb5444e90d368a3eb39de9920cac852c7a3f2fd03)

filiphr:
Thanks for the explanation @scottslewis, always good to learn a bit more about it.

>I could have...but chose to leave it so that the pr was not as large. If doing it in one step is preferable then that can be done.

I think the PR would actually be smaller then, since the classes that are removed, they'll show up as modified moved in git with no changes.

scottslewis:
> Thanks for the explanation @scottslewis, always good to learn a bit more about it.
> 
> > I could have...but chose to leave it so that the pr was not as large. If doing it in one step is preferable then that can be done.
> 
> I think the PR would actually be smaller then, since the classes that are removed, they'll show up as modified moved in git with no changes.

Ok, honestly I don't care that much.  Let's get the code and build/test/dependency changes reviewed/complete/consensus, etc...without breaking the build and test/CI if possible...and I'll delete or someone else can...whatever the choice is fine by me.

mcp-core/src/main/java/io/modelcontextprotocol/json/McpJsonDefaults.java
```diff
@@ -0,0 +1,50 @@
...
+
+	void setMcpJsonMapperSupplier(McpJsonMapperSupplier supplier) {
+		mcpMapperServiceLoader.setSupplier(supplier);
+	}
+
+	void unsetMcpJsonMapperSupplier(McpJsonMapperSupplier supplier) {
+		mcpMapperServiceLoader.unsetSupplier(supplier);
+	}
+
+	public synchronized static McpJsonMapper getDefaultMcpJsonMapper() {
```
Kehrlann, line 60: **[issue]** There are two separate `synchronized` methods, each using constructor side-effects. My understanding is that the `synchronized` is to avoid calling the constructor twice. If that's the case, in the current state, calling both methods at the same time triggers a race condition.

mcp-core/src/main/java/io/modelcontextprotocol/json/McpJsonDefaults.java
```diff
@@ -0,0 +1,50 @@
+package io.modelcontextprotocol.json;
```
Kehrlann, line 4: Missing license header

mcp-core/src/main/java/io/modelcontextprotocol/json/McpJsonDefaults.java
```diff
@@ -0,0 +1,50 @@
...
+
+import io.modelcontextprotocol.json.schema.JsonSchemaValidator;
+import io.modelcontextprotocol.json.schema.JsonSchemaValidatorSupplier;
+import io.modelcontextprotocol.util.McpServiceLoader;
+
+public class McpJsonDefaults {
```
Kehrlann, line 39: The patterns used here are uncommon in non-OSGi codebases. Please add a note in the javadoc mentioning OSGi compatibility.

mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
```diff
@@ -0,0 +1,50 @@
+package io.modelcontextprotocol.util;
```
Kehrlann, line 4: Missing license header.

mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.Optional;
+import java.util.ServiceConfigurationError;
+import java.util.ServiceLoader;
+import java.util.function.Supplier;
+
+public class McpServiceLoader<S extends Supplier<R>, R> {
```
Kehrlann, line 26: Same as McpJsonDefaults, mention OSGi here.

mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	@SuppressWarnings("unchecked")
+	public synchronized R getDefault() {
+		if (this.supplierResult == null) {
+			if (this.supplier == null) {
+				// Use serviceloader
+				Optional<?> sl = serviceLoad(this.supplierType);
+				if (sl.isEmpty()) {
+					throw new ServiceConfigurationError("No %s available for creating McpJsonMapper".format(this.supplierType.getSimpleName()));
```
Kehrlann: Did you mean to use `"...".formatted(...)`, or `String.format("...", ...)` ?

sdeleuze:
> > I think the PR would actually be smaller then, since the classes that are removed, they'll show up as modified moved in git with no changes.
> 
> Ok, honestly I don't care that much. Let's get the code and build/test/dependency changes reviewed/complete/consensus, etc...without breaking the build and test/CI if possible...and I'll delete or someone else can...whatever the choice is fine by me.

@scottslewis Happy to help collaborating with @Kehrlann to get this PR merged, but please squash/rework the commits to have files moved instead of copied as suggested by @filiphr, otherwise the diff is useless and the signal/noise ratio horrible. We do strongly care about that as reviewers, thanks for your understanding.

scottslewis:
> * I agree with @filiphr , remove `mcp-json`

Ok, I've deleted.

> 
>     * Please run `./mvnw spring-javaformat:apply`

I've been doing this, and did so again.

> 
>     * We need test passing before we can merge this

Of course.   I will need assistance from you/current committers with this for a couple of reasons:

1) For the tests that are going to move out of mcp-core, I don't know where they should go...to keep your build/test/CI stuff going smoothly.  I don't have any visibility into how all that works.

2) When I run the mvn clean install with tests, it fails on localhost because of the test dependency structure as discussed (can't find classes).  So 1 is blocking me on fixing the test/dependency structure.

mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
scottslewis: This was actually added by @LucaButBoring in his review

https://github.com/modelcontextprotocol/java-sdk/pull/762#discussion_r2743632987

So to answer to your question, I need to defer to @LucaButBoring

mcp-core/src/main/java/io/modelcontextprotocol/json/McpJsonDefaults.java
scottslewis: The use of static synchronized means that the lock is on the class rather than the instance, so the two static synchronized methods are using the same lock (the class itself).   So no race condition.

mcp-core/src/main/java/io/modelcontextprotocol/json/McpJsonDefaults.java
scottslewis: Thanks.  Fixed.  I'm currently just using:

/**
 * Copyright 2026 - 2026 the original author or authors.
 */

As per what's there in the existing.   That seems minimal/maybe insufficient, but I don't know what a better alternative would be.

mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
scottslewis: Also fixed.  Thanks.

mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
scottslewis: I've added non-OSGI vs. OSGi iniltialization description to javadoc.  If this is insufficient or more detail is needed, please lmk.

mcp-core/src/main/java/io/modelcontextprotocol/json/McpJsonDefaults.java
scottslewis: I've documented in class-level javadoc.  If more/other detail is desired, please just lmk.

scottslewis:
> 
> @scottslewis Happy to help collaborating with @Kehrlann to get this PR merged, but please squash/rework the commits to have files moved instead of copied as suggested by @filiphr, otherwise the diff is useless and the signal/noise ratio horrible. We do strongly care about that as reviewers, thanks for your understanding.

I've deleted mcp-json as requested.

mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
LucaButBoring: Oh, that should have been `.formatted()`, I'll add a new review correction for that, my mistake.

mcp-core/src/main/java/io/modelcontextprotocol/util/McpServiceLoader.java
```diff
@@ -0,0 +1,69 @@
...
+
+	@SuppressWarnings("unchecked")
+	public synchronized R getDefault() {
+		if (this.supplierResult == null) {
+			if (this.supplier == null) {
+				// Use serviceloader
+				Optional<?> sl = serviceLoad(this.supplierType);
+				if (sl.isEmpty()) {
+					throw new ServiceConfigurationError(
+							"No %s available for creating McpJsonMapper".format(this.supplierType.getSimpleName()));
```
LucaButBoring:
```suggestion
							"No %s available for creating McpJsonMapper".formatted(this.supplierType.getSimpleName()));
```

scottslewis:
As described [here](https://github.com/modelcontextprotocol/java-sdk/issues/562#issuecomment-3865813949) I've tested this pr...in OSGi environments and it now works properly in OSGi using either jackson2 or jackson3 jars.  This is good.

I'm willing to contribute automated tests for maintenance of the OSGi meta-data if desired.   Such tests will have to be coordinated with the current team for this project, so as to fit into the build/test/CI being used [along with test refactoring as described here](https://github.com/modelcontextprotocol/java-sdk/pull/762#issuecomment-3855573732).

Kehrlann:
Thank you for the updated PR, looks good.

I'll rebase this and move the tests so that they keep working. This will also allow us to start testing against jackson2 in addition to jackson3.

Once this is merged, could you please help us set up some automated testing for OSGi scenarios, so we can ensure compatibility in CI?

scottslewis:
> Once this is merged, could you please help us set up some automated testing for OSGi scenarios, so we can ensure compatibility in CI?

Yes, but I likely will need some access to the CI to do so.   Is the sdk project CI available somewhere?

Kehrlann:
We're using github actions - see the workflows here: https://github.com/modelcontextprotocol/java-sdk/tree/main/.github/workflows

You can probably play around on your fork of the project.

Kehrlann:
Merged through #779 . Thank you for your contribution!

scottslewis:
> Merged through https://github.com/modelcontextprotocol/java-sdk/pull/779 . Thank you for your contribution!

Thanks for the support. 

A few follow up questions:

1) When is the expected SDK release that will include these changes?

2) Could you describe here (or point to some docs or source) where the tests formerly in mcp-core ended up?....and if there is some new/other test procedure and/or maven profile for running them?

3) Finally, for the OSGi environment tests...I will need to add additional jars/java code to the test runtime environment and launch the runtime environment itself using some OSGi framework launcher.   In other OSGi projects I've used bndtools with gradle, tycho (maven plugins) and PDE many times in different CI environments.

LMK how you want to engage 2 and 3...e.g. new issue(s) with eventual pr...and/or online discussion, etc.   

Thanks.

Kehrlann:
Hey Scott!

1. I have to loop back with the team, but possibly next week.
2. I have documented the test changes in the README in [Project structure of the SDK](https://github.com/modelcontextprotocol/java-sdk/?tab=readme-ov-file#project-structure-of-the-sdk). Feel free to submit another PR if you'd like to improve the explanation.
3. I have no strong opinions on OSGi testing. If you know what you'd like to do, please open a PR and we can take a look. If you'd like to discuss before implementing, please open an issue and let's discuss approaches.

PR #764, opened 2026-01-29
feat: added tools name format validation accordingly #SEP-986

ashakirin:
Added tools name format validation accordingly #SEP-986
Note: tool duplications are already checked (in McpAsyncServer.java and in McpStatelessAsyncServer.java)

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
@@ -286,7 +286,7 @@ public interface McpServer {
 
 		String instructions;
 
-		Boolean skipStrictToolNameValidation;
+		boolean strictToolNameValidation = ToolNameValidator.isStrictByDefault();
 
 		/**
 		 * The Model Context Protocol (MCP) allows servers to expose tools that can be
@@ -405,14 +405,14 @@ public interface McpServer {
 		}
 
 		/**
-		 * Sets whether to skip strict tool name validation for this server. When set,
-		 * this takes priority over the system property
-		 * {@code io.modelcontextprotocol.skipStrictToolNameValidation}.
-		 * @param skip true to warn only, false to throw exception on invalid names
+		 * Sets whether to use strict tool name validation for this server. When set, this
+		 * takes priority over the system property
+		 * {@code io.modelcontextprotocol.strictToolNameValidation}.
+		 * @param strict true to throw exception on invalid names and false to warn only
 		 * @return This builder instance for method chaining
 		 */
-		public AsyncSpecification<S> skipStrictToolNameValidation(boolean strict) {
-			this.skipStrictToolNameValidation = strict;
+		public AsyncSpecification<S> strictToolNameValidation(boolean strict) {
+			this.strictToolNameValidation = strict;
 			return this;
 		}
 
@@ -553,7 +553,7 @@ public interface McpServer {
 		}
 
 		private void validateToolName(String toolName) {
-			ToolNameValidator.validate(toolName, this.skipStrictToolNameValidation);
+			ToolNameValidator.validate(toolName, this.strictToolNameValidation);
 		}
 
 		private void assertNoDuplicateTool(String toolName) {
@@ -905,7 +905,7 @@ public interface McpServer {
 
 		String instructions;
 
-		Boolean skipStrictToolNameValidation;
+		boolean strictToolNameValidation = ToolNameValidator.isStrictByDefault();
 
 		/**
 		 * The Model Context Protocol (MCP) allows servers to expose tools that can be
@@ -1028,14 +1028,14 @@ public interface McpServer {
 		}
 
 		/**
-		 * Sets whether to skip strict tool name validation for this server. When set,
-		 * this takes priority over the system property
-		 * {@code io.modelcontextprotocol.skipStrictToolNameValidation}.
-		 * @param skip true to warn only, false to throw exception on invalid names
+		 * Sets whether to use strict tool name validation for this server. When set, this
+		 * takes priority over the system property
+		 * {@code io.modelcontextprotocol.strictToolNameValidation}.
+		 * @param strict true to throw exception on invalid names, false to warn only
 		 * @return This builder instance for method chaining
 		 */
-		public SyncSpecification<S> skipStrictToolNameValidation(boolean strict) {
-			this.skipStrictToolNameValidation = strict;
+		public SyncSpecification<S> strictToolNameValidation(boolean strict) {
+			this.strictToolNameValidation = strict;
 			return this;
 		}
 
@@ -1175,7 +1175,7 @@ public interface McpServer {
 		}
 
 		private void validateToolName(String toolName) {
-			ToolNameValidator.validate(toolName, this.skipStrictToolNameValidation);
+			ToolNameValidator.validate(toolName, this.strictToolNameValidation);
 		}
 
 		private void assertNoDuplicateTool(String toolName) {
@@ -1473,7 +1473,7 @@ public interface McpServer {
 
 		String instructions;
 
-		Boolean skipStrictToolNameValidation;
+		boolean strictToolNameValidation = ToolNameValidator.isStrictByDefault();
 
 		/**
 		 * The Model Context Protocol (MCP) allows servers to expose tools that can be
@@ -1593,14 +1593,14 @@ public interface McpServer {
 		}
 
 		/**
-		 * Sets whether to skip strict tool name validation for this server. When set,
-		 * this takes priority over the system property
-		 * {@code io.modelcontextprotocol.skipStrictToolNameValidation}.
-		 * @param skip true to warn only, false to throw exception on invalid names
+		 * Sets whether to use strict tool name validation for this server. When set, this
+		 * takes priority over the system property
+		 * {@code io.modelcontextprotocol.strictToolNameValidation}.
+		 * @param strict true to throw exception on invalid names, false to warn only
 		 * @return This builder instance for method chaining
 		 */
-		public StatelessAsyncSpecification skipStrictToolNameValidation(boolean strict) {
-			this.skipStrictToolNameValidation = strict;
+		public StatelessAsyncSpecification strictToolNameValidation(boolean strict) {
+			this.strictToolNameValidation = strict;
 			return this;
 		}
 
@@ -1702,7 +1702,7 @@ public interface McpServer {
 		}
 
 		private void validateToolName(String toolName) {
-			ToolNameValidator.validate(toolName, this.skipStrictToolNameValidation);
+			ToolNameValidator.validate(toolName, this.strictToolNameValidation);
 		}
 
 		private void assertNoDuplicateTool(String toolName) {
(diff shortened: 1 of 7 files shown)


Kehrlann:
Thanks for your contribution @ashakirin ! A few comments.

## Optional validation

The spec specifically says that tools **SHOULD** follow the naming conventions ([Server > Tools > Tool names](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-names)). They **MAY** not follow these, and so we should allow invalid names (although not by default).
We could consider a per-server validator, that you plug in the `SyncSpecification` and `AsyncSpecification`. Thoughts?

## Client-side validation

I like the approach on the client side, validate but don't fail. The spec itself is super vague, "client can validate" which has no RFC meaning. Maybe we shouldn't even validate on the client side.

If we chose to validate, in the current implementation the log is very verbose if you're writing a client and the server (that you don't control) has an invalid tool name. Every time you make a "call tool request" you get a `WARN`-level log. I think that's too much, and the level is probably too high. In that case, keep the logs for the "list tool" request but not "call tool".

ashakirin:
@Kehrlann: Thanks for detail review, Daniel!
My thoughts:
> ## Optional validation
> The spec specifically says that tools **SHOULD** follow the naming conventions ([Server > Tools > Tool names](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-names)). They **MAY** not follow these, and so we should allow invalid names (although not by default). We could consider a per-server validator, that you plug in the `SyncSpecification` and `AsyncSpecification`. Thoughts?
>
My initial interpretation of SHOULD was that the SDK may choose whether to support validation at all, not that validation of server-side tool names should be optional. That seems to be incorrect.

Proposal: using different validation options for different MCP servers registered in the same application is probably not a very typical use case. I would therefore propose a global setting to disable validation, with an option to override it per server in SyncSpecification, AsyncSpecification, StatelessAsyncSpecification, and StatelessSyncSpecification. WDYT?
This keeps the default behavior simple while still allowing flexibility when integrating with non-compliant servers. What do you think? 

> ## Client-side validation
> I like the approach on the client side, validate but don't fail. The spec itself is super vague, "client can validate" which has no RFC meaning. Maybe we shouldn't even validate on the client side.
> 
> If we chose to validate, in the current implementation the log is very verbose if you're writing a client and the server (that you don't control) has an invalid tool name. Every time you make a "call tool request" you get a `WARN`-level log. I think that's too much, and the level is probably too high. In that case, keep the logs for the "list tool" request but not "call tool".

My idea was to validate tool names via deserialization on both the client and server sides: on the client during listTools, and on the server during tool/call.
I agree that the current implementation in compact constructors is too verbose, because:
- the client constructs the CallToolRequest itself
- the compact listTools constructor is invoked twice due calling of pagination / cursors constructor

Therefore I am going to move validation into McpAsyncClient.listToolsInternal().
Do you think it makes sense to validate tool names on the server side during tool/call? Basically a client can call server with non-existent and invalid tool name, but that will result in a “tool not found” error. And existing invalid server-side tool names will be already validated during tool registration on the server side - so additional validation seems doesn't add value.

Kehrlann:
> Proposal: using different validation options for different MCP servers registered in the same application is probably not a very typical use case. I would therefore propose a global setting to disable validation, with an option to override it per server in SyncSpecification, AsyncSpecification, StatelessAsyncSpecification, and StatelessSyncSpecification. WDYT? This keeps the default behavior simple while still allowing flexibility when integrating with non-compliant servers. What do you think?

I like the idea, something like `.skipToolNameValidation(boolean skip)`?

> Therefore I am going to move validation into McpAsyncClient.listToolsInternal(). 

Works for me.

> Do you think it makes sense to validate tool names on the server side during tool/call? Basically a client can call server with non-existent and invalid tool name, but that will result in a “tool not found” error. And existing invalid server-side tool names will be already validated during tool registration on the server side - so additional validation seems doesn't add value.

No, don't validate client calls on the server. If a tool name does not exist, it does not exist - no matter the reason.
If a tool is present but has an invalid name, that means the user configured their server with "skip validation" and therefore we assume they know what they are doing and do not want extra messages.

ashakirin:
@Kehrlann: have updated PR accordingly our discussion, could you please take a look?

mcp-core/src/main/java/io/modelcontextprotocol/util/ToolNameValidator.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
```
Kehrlann:
```suggestion
 * Copyright 2026-2026 the original author or authors.
```

mcp-core/src/main/java/io/modelcontextprotocol/util/ToolNameValidator.java
```diff
@@ -0,0 +1,88 @@
...
+ * <ul>
+ * <li>Must be between 1 and 128 characters in length</li>
+ * <li>May only contain: A-Z, a-z, 0-9, underscore (_), hyphen (-), and dot (.)</li>
+ * <li>Must not contain spaces, commas, or other special characters</li>
+ * </ul>
+ *
+ * @see <a href=
+ * "https://modelcontextprotocol.io/specification/draft/server/tools#tool-names">MCP
+ * Specification - Tool Names</a>
+ */
```
Kehrlann, line 27: Feel free to add an `@author` tag if you'd like.

mcp-core/src/test/java/io/modelcontextprotocol/util/ToolNameValidatorTests.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
```
Kehrlann:
```suggestion
 * Copyright 2026-2026 the original author or authors.
```

mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
```diff
@@ -407,6 +404,18 @@ public AsyncSpecification<S> instructions(String instructions) {
...
+		 * this takes priority over the system property
+		 * {@code io.modelcontextprotocol.skipStrictToolNameValidation}.
+		 * @param skip true to warn only, false to throw exception on invalid names
+		 * @return This builder instance for method chaining
+		 */
+		public AsyncSpecification<S> skipStrictToolNameValidation(boolean strict) {
+			this.skipStrictToolNameValidation = strict;
+			return this;
+		}
+
```
Kehrlann, line 418: Reading the javadoc, i realize that `skipStrictToolNameValidation` is probably the wrong way to put it - sorry for the earlier suggestion. A boolean == true should mean "do the thing", not "skip the thing".

Invert the boolean and use `strictToolNameValidation` instead of `skip...`

Same goes for all transports.

mcp-core/src/main/java/io/modelcontextprotocol/util/ToolNameValidator.java
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	/**
+	 * Validates a tool name according to MCP specification.
+	 * @param name the tool name to validate
+	 * @param skip if true, logs warning only; if false, throws exception; if null, uses
+	 * system property {@link #SKIP_STRICT_VALIDATION_PROPERTY} (default: false)
+	 * @throws IllegalArgumentException if validation fails and skip is false
+	 */
+	public static void validate(String name, Boolean skip) {
```
Kehrlann: See comments in `McpAsyncClient`: change `skip` to `strict`.
Use a `boolean` rather than a boxed type.

Let's enforce validation by default. It's a breaking change but that wouldn't be our first in this 0.x line.

mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
```diff
@@ -291,6 +286,8 @@ abstract class AsyncSpecification<S extends AsyncSpecification<S>> {
 
 		String instructions;
 
+		Boolean skipStrictToolNameValidation;
```
Kehrlann: Use a `boolean` with a default value.

mcp-core/src/main/java/io/modelcontextprotocol/util/ToolNameValidator.java
```diff
@@ -0,0 +1,88 @@
...
+		boolean warnOnly = skip != null ? skip : Boolean.getBoolean(SKIP_STRICT_VALIDATION_PROPERTY);
+
+		if (name == null || name.isEmpty()) {
+			handleError("Tool name must not be null or empty", name, warnOnly);
+			return;
+		}
+		if (name.length() > MAX_LENGTH) {
```
Kehrlann: Nitpick: use else-if over returns.

mcp-core/src/test/java/io/modelcontextprotocol/server/AsyncToolSpecificationBuilderTest.java
```diff
@@ -264,4 +283,42 @@ void fromSyncShouldReturnNullWhenSyncSpecIsNull() {
 		assertThat(McpServerFeatures.AsyncToolSpecification.fromSync(null)).isNull();
 	}
 
+	@Test
+	void toolNameValidation_defaultShouldThrowOnInvalidName() {
```
Kehrlann: Avoid underscores, we don't use it in our tests.

If you want to group tests, consider JUnit's `@Nested`

mcp-core/src/test/java/io/modelcontextprotocol/util/ToolNameValidatorTests.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Tests for {@link ToolNameValidator}.
+ */
+class ToolNameValidatorTests {
+
+	@ParameterizedTest
+	@ValueSource(strings = { "getUser", "DATA_EXPORT_v2", "admin.tools.list", "my-tool", "Tool123", "a", "A",
+			"_private", "tool_name", "tool-name", "tool.name", "UPPERCASE", "lowercase", "MixedCase123" })
+	void validToolNames_strictMode(String name) {
```
Kehrlann, line 22: See comments in `AsyncToolSpecificationBuilderTest` re: not using underscores.

mcp-core/src/test/java/io/modelcontextprotocol/spec/McpSchemaTests.java
```diff
@@ -1765,4 +1765,26 @@ void testProgressNotificationWithoutMessage() throws Exception {
...
+	void testListToolsResultDeserializationWithInvalidToolName() throws Exception {
+		// Deserialization should not throw, just warn
+		String json = """
+				{"tools":[{"name":"invalid tool name","description":"test"}],"nextCursor":null}""";
+
+		McpSchema.ListToolsResult result = JSON_MAPPER.readValue(json, McpSchema.ListToolsResult.class);
+
+		assertThat(result.tools()).hasSize(1);
+		assertThat(result.tools().get(0).name()).isEqualTo("invalid tool name");
+	}
```
Kehrlann: Please add a test showing it throws.

ashakirin: This test is obsolete and stays from the times when validation was in McpSchema. Removed.

Kehrlann:
Thank you for your contribution, this is now on `main`.

PR #771, opened 2026-02-04
Validate origin header

### LINKED ISSUE #695: Return HTTP 403 when Origin header is invalid
This tracks https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1439

See also: #284

Kehrlann:
Fixes #695 . Updates conformance tests.

## Contents

Add Origin header validation for HTTP-based MCP server transports so servers can restrict which origins can connect (e.g. browser or cross-origin clients).

Introduces new `ServerTransportSecurityValidator` interface, to validate transport-level security. Default behavior stays permissive: transports use `ServerTransportSecurityValidator.NOOP` unless a validator is set via the builder.

The provided implementation is `DefaultServerTransportSecurityValidator` and only validates the Origin header, for now.

## New APIs: example usage

```java
HttpServletStreamableServerTransportProvider.builder()
// ...
.securityValidator(
    DefaultServerTransportSecurityValidator.builder()
        .allowedOrigin("http://localhost:*")
        .build())
// ...
```

## Points of interest

Integration tests use the new JUnit 6 `@ParameterizedClass` API, which is another way of implementation "abstract base classes" for tests.

### DIFF
--- a/conformance-tests/VALIDATION_RESULTS.md
+++ b/conformance-tests/VALIDATION_RESULTS.md
@@ -15,13 +15,12 @@
 - **Resources (4/6):** list, read-text, read-binary, templates-read
 - **Prompts (4/4):** list, simple, with-args, embedded-resource, with-image
 - **SSE Transport (2/2):** Multiple streams
-- **Security (1/2):** Localhost validation passes
+- **Security (2/2):** Localhost validation passes, DNS rebinding protection
 
 ### Failing (3/40)
 
 1. **resources-subscribe** - Not implemented in SDK
 2. **resources-unsubscribe** - Not implemented in SDK  
-3. **dns-rebinding-protection** - Missing Host/Origin validation (1/2 checks)
 
 ## Client Test Results
 
@@ -44,7 +43,6 @@
 
 1. **Resource Subscriptions:** SDK doesn't implement `resources/subscribe` and `resources/unsubscribe` handlers
 2. **Client SSE Retry:** Client doesn't parse or respect the `retry:` field, reconnects immediately, and doesn't send Last-Event-ID header
-3. **DNS Rebinding Protection:** Missing Host/Origin header validation in server transport
 
 ## Running Tests
 
--- a/conformance-tests/conformance-baseline.yml
+++ b/conformance-tests/conformance-baseline.yml
@@ -6,9 +6,6 @@ server:
   # Resource subscription not implemented in SDK
   - resources-subscribe
   - resources-unsubscribe
-  
-  # DNS rebinding protection missing Host/Origin validation
-  - dns-rebinding-protection
 
 client:
   # SSE retry field handling not implemented
--- a/conformance-tests/server-servlet/README.md
+++ b/conformance-tests/server-servlet/README.md
@@ -32,8 +32,8 @@ The server has been validated against the official [MCP conformance test suite](
 ✅ **SSE Transport** (2/2)
 - Multiple streams support
 
-⚠️ **Security** (1/2)
-- ⚠️ DNS rebinding protection (SDK limitation)
+✅ **Security** (2/2)
+- ✅ DNS rebinding protection
 
 ## Features
 
--- a/conformance-tests/server-servlet/src/main/java/io/modelcontextprotocol/conformance/server/ConformanceServlet.java
+++ b/conformance-tests/server-servlet/src/main/java/io/modelcontextprotocol/conformance/server/ConformanceServlet.java
@@ -7,6 +7,7 @@ import java.util.Map;
 
 import io.modelcontextprotocol.server.McpServer;
 import io.modelcontextprotocol.server.McpServerFeatures;
+import io.modelcontextprotocol.server.transport.DefaultServerTransportSecurityValidator;
 import io.modelcontextprotocol.server.transport.HttpServletStreamableServerTransportProvider;
 import io.modelcontextprotocol.spec.McpSchema.AudioContent;
 import io.modelcontextprotocol.spec.McpSchema.BlobResourceContents;
@@ -66,6 +67,8 @@ public class ConformanceServlet {
 			.builder()
 			.mcpEndpoint(MCP_ENDPOINT)
 			.keepAliveInterval(Duration.ofSeconds(30))
+			.securityValidator(
+					DefaultServerTransportSecurityValidator.builder().allowedOrigin("http://localhost:*").build())
 			.build();
 
 		// Build server with all conformance test features


mcp-core/src/test/java/io/modelcontextprotocol/server/transport/DefaultServerTransportSecurityValidatorTests.java
```diff
@@ -0,0 +1,197 @@
...
+		@Test
+		void differentHostWithWildcard() {
+			var headers = originHeader("http://example.com:3000");
+
+			assertThatThrownBy(() -> wildcardValidator.validateHeaders(headers)).isEqualTo(INVALID_ORIGIN);
+		}
+
+		@Test
+		void differentSchemeWithWildcard() {
+			var headers = originHeader("https://localhost:3000");
```
tzolov, line 122: shouldn't the port be 8080 to ensure that it fails because of the schema?

tzolov (approved):
LGTM. Great work @Kehrlann !

mcp-core/src/test/java/io/modelcontextprotocol/server/transport/DefaultServerTransportSecurityValidatorTests.java
Kehrlann: It's using the the wildcard validator, so the port does not matter:


```java
private final DefaultServerTransportSecurityValidator wildcardValidator =
	DefaultServerTransportSecurityValidator
		.builder()
		.allowedOrigin("http://localhost:*")
		.build();
```

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/DefaultServerTransportSecurityValidator.java
```diff
@@ -0,0 +1,132 @@
...
+				// Wildcard port pattern: "http://example.com:*"
+				String baseOrigin = allowed.substring(0, allowed.length() - 2);
+				if (origin.equals(baseOrigin) || origin.startsWith(baseOrigin + ":")) {
+					return;
+				}
+			}
+
+		}
+
+		throw INVALID_ORIGIN;
```
chemicL, line 82: Due to the exception being a static, the stack trace is inaccurate with regards to where it is raised. I understand the optimization, however it might make it harder to investigate where the issue happened. Perhaps we can create an instance each time it's raised?

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/DefaultServerTransportSecurityValidator.java
```diff
@@ -0,0 +1,132 @@
...
+			"Invalid Origin header");
+
+	private final List<String> allowedOrigins;
+
+	/**
+	 * Creates a new validator with the specified allowed origins.
+	 * @param allowedOrigins List of allowed origin patterns. Supports exact matches
+	 * (e.g., "http://example.com:8080") and wildcard ports (e.g., "http://example.com:*")
+	 */
+	public DefaultServerTransportSecurityValidator(List<String> allowedOrigins) {
```
chemicL, line 38: There are two ways to create the validator - either via a constructor or via the builder. I'd vouch for a single version. The builder makes sense with no inheritance, but since users are encouraged to customize by inheriting from this class, perhaps we can skip the builder?

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
```diff
@@ -311,6 +330,15 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
 
+		try {
+			Map<String, List<String>> headers = extractHeaders(request);
+			this.securityValidator.validateHeaders(headers);
+		}
+		catch (ServerTransportSecurityException e) {
+			response.sendError(e.getStatusCode(), e.getMessage());
+			return;
+		}
+
```
chemicL, line 341: This bit is duplicated in doGet, we can create a helper method for it perhaps?

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStatelessServerTransport.java
```diff
@@ -122,6 +135,15 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
 		}
 
+		try {
+			Map<String, List<String>> headers = extractHeaders(request);
+			this.securityValidator.validateHeaders(headers);
+		}
+		catch (ServerTransportSecurityException e) {
+			response.sendError(e.getStatusCode(), e.getMessage());
+			return;
+		}
```
chemicL, line 145: This code is duplicated across three implementations, perhaps we can have a centralized static utility for it?

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStatelessServerTransport.java
```diff
@@ -209,6 +231,21 @@ private void responseError(HttpServletResponse response, int httpCode, McpError
 		writer.flush();
 	}
 
+	/**
+	 * Extracts all headers from the HTTP servlet request into a map.
+	 * @param request The HTTP servlet request
+	 * @return A map of header names to their values
+	 */
+	private Map<String, List<String>> extractHeaders(HttpServletRequest request) {
```
chemicL, line 239: This private method is also duplicated across implementations, another candidate for sharing.

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
```diff
@@ -246,6 +260,15 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response)
...
 		}
 
+		try {
+			Map<String, List<String>> headers = extractHeaders(request);
+			this.securityValidator.validateHeaders(headers);
+		}
+		catch (ServerTransportSecurityException e) {
+			response.sendError(e.getStatusCode(), e.getMessage());
+			return;
+		}
```
chemicL, line 270: Can reuse a shared static util.

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
```diff
@@ -373,6 +396,15 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
```
chemicL, line 406: Duplicated :)

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
```diff
@@ -536,6 +568,15 @@ protected void doDelete(HttpServletRequest request, HttpServletResponse response
...
```
chemicL, line 578: Another duplicate that can reuse a shared util.

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
```diff
@@ -586,6 +627,21 @@ public void responseError(HttpServletResponse response, int httpCode, McpError m
 		return;
 	}
 
+	/**
+	 * Extracts all headers from the HTTP servlet request into a map.
+	 * @param request The HTTP servlet request
+	 * @return A map of header names to their values
+	 */
+	private Map<String, List<String>> extractHeaders(HttpServletRequest request) {
```
chemicL, line 635: Duplicate

mcp-core/src/test/java/io/modelcontextprotocol/server/transport/ServerTransportSecurityIntegrationTests.java
```diff
@@ -0,0 +1,292 @@
...
+
+	// ----------------------------------------------------
+	// Transport servers to test
+	// ----------------------------------------------------
+
+	/**
+	 * All transport types we want to test. We use a {@link MethodSource} rather than a
+	 * {@link org.junit.jupiter.params.provider.ValueSource} to provide a readable name.
+	 */
+	static Stream<Arguments> transports() {
```
chemicL, line 153: Nice!

mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java
```diff
@@ -332,6 +350,14 @@ private Mono<ServerResponse> handleMessage(ServerRequest request) {
 			return ServerResponse.status(HttpStatus.SERVICE_UNAVAILABLE).bodyValue("Server is shutting down");
 		}
 
+		try {
+			Map<String, List<String>> headers = request.headers().asHttpHeaders();
+			this.securityValidator.validateHeaders(headers);
+		}
+		catch (ServerTransportSecurityException e) {
+			return ServerResponse.status(e.getStatusCode()).bodyValue(e.getMessage());
+		}
```
chemicL, line 359: This can be unified perhaps too.

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/DefaultServerTransportSecurityValidator.java
Kehrlann: - Keeping the builder, because I'm going to add Host header validation
- Making the class final ; it doesn't make a lot of sense to extend it.

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletSseServerTransportProvider.java
Kehrlann: No obvious solution around this. The ideal would be to have

```java
Map<String, List<String>> headers = extractHeaders(request);
this.securityValidator.validateHeaders(headers);
```

+ the error handled outside the of method handler (doGet / ...). Too much for now, so leaving it as is.

PR #775, opened 2026-02-05
DNS rebinding protection: check host header

Kehrlann:


Kehrlann:
Will merge after #782

Issue #574, opened 2025-09-25
no default JsonSchemaValidatorSupplier or default McpJsonMapper in mcp-json project


scottslewis:
As per bug #562 I'm also getting these exceptions when running the latest 0.14.0 in repo (not deployed yet).

java.lang.IllegalStateException: No default JsonSchemaValidatorSupplier implementation found
	at io.modelcontextprotocol.json.schema.JsonSchemaInternal.lambda$createDefaultValidator$2(JsonSchemaInternal.java:66)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at io.modelcontextprotocol.json.schema.JsonSchemaInternal.createDefaultValidator(JsonSchemaInternal.java:61)
	at io.modelcontextprotocol.json.schema.JsonSchemaInternal.getDefaultValidator(JsonSchemaInternal.java:30)
	at io.modelcontextprotocol.json.schema.JsonSchemaValidator.getDefault(JsonSchemaValidator.java:61)
	at io.modelcontextprotocol.server.McpServer$SingleSessionAsyncSpecification.build(McpServer.java:232)
	at ...

and

java.lang.IllegalStateException: No default McpJsonMapper implementation found
	at io.modelcontextprotocol.json.McpJsonInternal.lambda$createDefaultMapper$2(McpJsonInternal.java:67)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at io.modelcontextprotocol.json.McpJsonInternal.createDefaultMapper(McpJsonInternal.java:62)
	at io.modelcontextprotocol.json.McpJsonInternal.getDefaultMapper(McpJsonInternal.java:30)
	at io.modelcontextprotocol.json.McpJsonMapper.getDefault(McpJsonMapper.java:97)
	at io.modelcontextprotocol.server.McpServer$SingleSessionSyncSpecification.build(McpServer.java:818)
	at ...

This problem does not appear to me to be an OSGi problem as when I look in the [mcp-json project](https://github.com/modelcontextprotocol/java-sdk/tree/main/mcp-json) there does not appear to be any META-INF/services/* files.   I can see that the ServiceLoader is used in io.modelcontextprotocol.json.schema.JsonSchemaInternal line 41, and in io.modelcontextprotocol.json.McpJsonInternal line 45, so I suspect that the META-INF/services/* files in mcp-json were accidently not checked in.

garpinc:
[...]

PR #779, opened 2026-02-11
Moves mcp-json API back into mcp-core for simplified dependencies and support of osgi runtimes

Kehrlann:
This PR supersedes #762 

## Context

We are bringing proper OSGi support to the SDK. See #562 , #574.

As part of this effort, we are bringing `mcp-json` back into `mcp-core`, see:  #612 and #682 

This means integrations tests inside `mcp-core`, which used to run with `mcp-json-jackson3`, cannot run as-is, as `mcp-jackson-3` now depends on `mcp-core` and maven can't handle the cyclic dependencies.

## Contents

1. Add OSGi relevant metadata + bnd plugin 
2. Move mcp-json back into mcp-core
3. Move integration tests to `mcp-test`
4. Add `jackson2` profile for running integration tests with `jackson2`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,11 +83,11 @@ The following sections explain what we chose, why it made sense, and how the cho
 
 ### 1. JSON Serialization
 
-* **SDK Choice**: Jackson for JSON serialization and deserialization, behind an SDK abstraction (`mcp-json`)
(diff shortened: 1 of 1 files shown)


scottslewis:
I've create pr #792 for small fixes and simplifications for the OSGI manifest.mf contents generation done at build time.  Please merge prior to release.

Issue #793, opened 2026-02-14
API:  declarations prevent subclassing/extending public classes


scottslewis:
I would like to extend the McpAsyncServer, McpSyncServer and the stateless versions to extend the behavior of these servers via a new MCP extension under development.

Currently, these classes, and a couple of classes that their constructors depend upon (McpServerFeatures.Sync and Async) are package protected meaning that only the code in the sdk can use these constructors, effectively preventing the creation of subclasses.

**Environment**

version 0.18.0.SNAPSHOT of sdk as of 2/14/2026
Please provide as many details as possible: Spring MCP version, Java version, which vector store you use if any, etc

**Expected behavior**

Ability to declare subclasses of McpSyncServer, McpAsyncServer, and stateless analogs.   Subclass constructors must call the superclass constructor and so currently cannot be from outside of io.modelcontextprotocol package namespace.

**Minimal Complete Reproducible example**

The follow code will not compile, because the superclass (McpAsyncServer) constructor is package private.

```java
	class AsyncToolGroupServer extends McpAsyncServer {

		public AsyncToolGroupServer(McpServerTransportProvider mcpTransportProvider, McpJsonMapper jsonMapper,
				Async features, Duration requestTimeout, McpUriTemplateManagerFactory uriTemplateManagerFactory,
				JsonSchemaValidator jsonSchemaValidator) {
			super(mcpTransportProvider, jsonMapper, features, requestTimeout, uriTemplateManagerFactory, jsonSchemaValidator);
		}
}

fix pr forthcoming

Kehrlann:
As I mentioned in #794 :

> As a rule of thumb, we avoid making our core implementations extensible through subclassing. Instead, we prefer composition, and introducing specific APIs.

Would you be so kind as to open an issue about your specific need, IIRC customizing params in notifications? I think you mentioned `notifyToolsListChanged`, but this could probably apply to all notifications.

scottslewis:
> 
> Would you be so kind as to open an issue about your specific need, IIRC customizing params in notifications? I think you mentioned `notifyToolsListChanged`, but this could probably apply to all notifications.

See #831

PR #794, opened 2026-02-14
Fix for #793 - public (or protected) constructors to allow subclassing for extensions

scottslewis:
Fix for issue #793 

This pr creates public constructors for McpSyncServer, McpAsyncServer and the stateless sync/async McpServer* types.

Technically the constructors could be protected rather than public, but I think for an SDK that public constructors is likely preferable.

## Motivation and Context

The main motivation is to be able to subclass the various sdk server and client types...i.e. McpSyncServer, McpAsyncServer, McpSyncClient, McpAsyncClient, etc.   I am creating an extension that has to override/customized notifyToolsListChanged() so that params can be added to the notification.

There are other use cases for extending the various server and client types in an sdk, along with making member variables protected, so that sdk external subsclasses can extend the behavior of these types.

Also, to extend these types requires that it be possible to invoke a subclass constructor, with instances of all the required arguments (e.g. McpServerFeatures.Async for McpAsyncServer).  So these types also need to have public constructors.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/json/TypeRef.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/json/TypeRef.java
@@ -32,6 +32,10 @@ public abstract class TypeRef<T> {
 		this.type = ((ParameterizedType) superClass).getActualTypeArguments()[0];
 	}
 
+	protected TypeRef(Type type) {
+		this.type = type;
+	}
+
 	/**
 	 * Returns the captured type information.
 	 * @return the Type representing the actual type argument captured by this TypeRef


Kehrlann:
As a rule of thumb, we avoid making our core implementations extensible through subclassing.  Instead, we prefer composition, and introducing specific APIs. 

In this case, some sort of handler for customizing the params seems to better fit. I'm going to close this, and we can start again from another issue.

scottslewis:
> As a rule of thumb, we avoid making our core implementations extensible through subclassing. Instead, we prefer composition, and introducing specific APIs.
> 
> In this case, some sort of handler for customizing the params seems to better fit. I'm going to close this, and we can start again from another issue.

For context:  I'm creating an [MCP extension](https://github.com/OpenMCPTools/mcp_extensions_java/tree/main/org.openmcptools.extensions.update)...that extends the behavior of server side notification sending to add params/data to the notification.   There is currently no way to do this with the current sdk class structure.  I believe this is true for all of McpSyncServer, async, stateless, etc ..as with the client clsses as well (not totally sure about that).

I'm a fan of composition also fwiw.

More generally, though, composition works only when the API of a class is actually exposed for composition at runtime...i.e. via interfaces and/or abstract super classes....which I would strongly suggest for all of the McpSync/AsyncServer classes (and the client classes as well).  I'm willing to contribute to that refactoring if desired.  

In this case though, it's server notification extension that's required, meaning some way to replace the notification that's done (via mcpTransport member variable) to include data in the form of a non-null params.  As well, the existing notify methods access mcpTransport in the current implementation (passing in null params), which is a private member variable currently, so it can't be accessed at all at runtime through inheritance or composition.  [Here is the extension impl that does that](https://github.com/OpenMCPTools/mcp_common_java/blob/main/org.openmcptools.common.impl.spring/src/main/java/org/openmcptools/common/server/toolgroup/impl/spring/McpAsyncToolGroupServer.java#L43).

I'm open to other ways of accomplishing this than inheritance, but for a viable extension a public (or perhaps just protected) constructors for the existing classes would be quick and easy...without compromising maintainability (especially protected).  FWIW, I've found that both inheritance and composition are sometimes useful for SDK consumers.

Please let me know what you want to do wrt refactoring strategy in both short and longer term to make this extension possible and I will open appropriate issues.

PR #801, opened 2026-02-18
Update SECURITY.md to use GitHub Security Advisories

localden:
Replaces Anthropic HackerOne VDP references with standard GitHub Security Advisory process.


Kehrlann:
@localden this means we should turn on private vuln disclosure on this repo, correct?

Kehrlann:
I see it's been turned on for the repo. Merged.

Issue #686, opened 2025-11-25
MCP server using StdioServerTransportProvider does not support multiple asynchronous tool calls for the same session


aoetk:
**Bug description**

On an MCP server using StdioServerTransportProvider, if multiple asynchronous tool calls are made on the same session, the following RuntimeException may be thrown on the server side, and the server may stop responding thereafter.

```
[boundedElastic-1] ERROR reactor.core.publisher.Operators - Operator called default onErrorDropped
reactor.core.Exceptions$ErrorCallbackNotImplemented: java.lang.RuntimeException: Failed to enqueue message
Caused by: java.lang.RuntimeException: Failed to enqueue message
	at io.modelcontextprotocol.server.transport.StdioServerTransportProvider$StdioMcpSessionTransport.lambda$sendMessage$0(StdioServerTransportProvider.java:154)
	at reactor.core.publisher.MonoDefer.subscribe(MonoDefer.java:45)
	at reactor.core.publisher.MonoIgnoreThen$ThenIgnoreMain.subscribeNext(MonoIgnoreThen.java:241)
	at reactor.core.publisher.MonoIgnoreThen$ThenIgnoreMain.onComplete(MonoIgnoreThen.java:204)
	at reactor.core.publisher.MonoZip$ZipInner.onComplete(MonoZip.java:536)
	at reactor.core.publisher.Operators$MonoInnerProducerBase.complete(Operators.java:2869)
	at reactor.core.publisher.SinkOneMulticast.subscribe(SinkOneMulticast.java:98)
	at reactor.core.publisher.MonoZip$ZipCoordinator.request(MonoZip.java:220)
	at reactor.core.publisher.MonoIgnoreThen$ThenIgnoreMain.onSubscribe(MonoIgnoreThen.java:135)
	at reactor.core.publisher.MonoZip.subscribe(MonoZip.java:129)
	at reactor.core.publisher.Mono.subscribe(Mono.java:4576)
	at reactor.core.publisher.MonoIgnoreThen$ThenIgnoreMain.subscribeNext(MonoIgnoreThen.java:265)
	at reactor.core.publisher.MonoIgnoreThen.subscribe(MonoIgnoreThen.java:51)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.onNext(MonoFlatMap.java:165)
	at reactor.core.publisher.FluxOnErrorResume$ResumeSubscriber.onNext(FluxOnErrorResume.java:79)
[...]

PR #834, opened 2026-02-24
fix: handle concurrent sendMessage calls in StdioServerTransportProvider

dominikozi:
Replace tryEmitNext() with emitNext() using a busy-looping retry handler in StdioMcpSessionTransport.sendMessage() to support concurrent tool call responses without crashing.

## Motivation and Context
Fixes [#686](https://github.com/modelcontextprotocol/java-sdk/issues/686). 
When multiple tool calls are executed concurrently on MCP server using StdioServerTransportProvider, the server crashes with RuntimeException: Failed to enqueue message. The reason is that sendMessage() used tryEmitNext() on the outbound sink, which fails immediately when another thread is concurrently emitting instead of retrying.

## How Has This Been Tested?
Added shouldHandleConcurrentMessages test that sends two concurrent JSON-RPC requests with delayed responses, verifying both complete without error and are written to the output stream. The test fails on main and passes with this change. Verified all existing tests continue to pass.

## Breaking Changes
None.

## Types of changes

[X] Bug fix (non-breaking change which fixes an issue)

[ ] New feature

[ ] Breaking change

[ ] Documentation update

## Checklist

[X] I have read the MCP Documentation

[X] My code follows the repository's style guidelines

[X] New and existing tests pass locally

[X] I have added appropriate error handling

[X] I have added or updated documentation as needed

## Additional context

The fix replaces tryEmitNext() with emitNext() using Sinks.EmitFailureHandler.busyLooping(Duration.ofSeconds(1)) in StdioMcpSessionTransport.sendMessage(). This is the standard Reactor pattern for thread safe emission from multiple threads. Non retryable failures still propagate immediately as Sinks.EmissionException. I dont think this requires a documentation update.

### DIFF
--- a/mcp-test/src/test/java/io/modelcontextprotocol/server/transport/StdioServerTransportProviderTests.java
+++ b/mcp-test/src/test/java/io/modelcontextprotocol/server/transport/StdioServerTransportProviderTests.java
@@ -4,15 +4,26 @@
 
 package io.modelcontextprotocol.server.transport;
 
+import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
 import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 
 import io.modelcontextprotocol.json.McpJsonDefaults;
 import io.modelcontextprotocol.spec.McpError;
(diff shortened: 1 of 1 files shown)


mcp-test/src/test/java/io/modelcontextprotocol/server/transport/StdioServerTransportProviderTests.java
```diff
@@ -220,4 +220,47 @@ void shouldHandleSessionClose() throws Exception {
...
+
+		String jsonMessage1 = "{\"jsonrpc\":\"2.0\",\"method\":\"test1\",\"params\":{},\"id\":1}\n";
+		String jsonMessage2 = "{\"jsonrpc\":\"2.0\",\"method\":\"test2\",\"params\":{},\"id\":2}\n";
+		pipedOut.write(jsonMessage1.getBytes(StandardCharsets.UTF_8));
+		pipedOut.write(jsonMessage2.getBytes(StandardCharsets.UTF_8));
+		pipedOut.flush();
+
+		// Verify both concurrent responses complete without error
+		StepVerifier
+			.create(Mono.delay(java.time.Duration.ofSeconds(2))
```
chemicL: The test passes if `Duration.ofSeconds(3)` is provided. Preferably, instead of using delays, you'd launch a swarm of threads trying to process the messages in parallel and ensure none of them fail and all the messages are processed.

dominikozi: Updated the test according to your suggestion. The delay-based test has been replaced with a swarm test, which sends 100 requests. The test verifies all response IDs are received, no failures occur, and that concurrency actually happened. Added a 15s Timeout, consistent with existing tests in the repository.

Issue #777, opened 2026-02-09
Missing handler for request type: notifications/initialized


woweow:
Please do a quick search on GitHub issues first, there might be already a duplicate issue for the one you are about to create.
If the bug is trivial, just go ahead and create the issue. Otherwise, please take a few moments and fill in the following sections:

**Bug description**
I am getting the following errors from my MCP:
```
ERROR io.modelcontextprotocol.server.transport.HttpServletStatelessServerTransport - Failed to handle request: Missing handler for request type: notifications/initialized
```

**Environment**
version 17.1 of the MCP SDK. Running kotlin

**Steps to reproduce**
Unable to reproduce - seems like a client is more up to date on the spec than we are but I haven't been able to determine which

**Expected behavior**
I wouldn't expect anything a client says to my server to fail the protocol. I think we need an additional handler? Maybe I am dumb but in the past when I had this issue for tools et all, I just needed to explicitly say that I don't support them:

```
    private val SERVER_CAPABILITIES: McpSchema.ServerCapabilities = McpSchema.ServerCapabilities.builder()
        .prompts(true)
        .tools(false)
        .resources(false, false)
        .completions()
        .build()
```

however there is not a "notifications" capability

**Minimal Complete Reproducible example**
Unable to reproduce right now. I am guessing some update was made to MCP which we don't have yet in 17.1

chemicL:
This sounds strange. Please provide a reproducer so we can investigate.

Kehrlann:
No additional activity from the community on this one, we consider it a one-off. Closing due to inactivity.

PR #835, opened 2026-02-24
Remove invalid WARN message for missing notification handler in stateless servers

bperhaps:
## Motivation and Context
https://github.com/modelcontextprotocol/java-sdk/issues/777

image

According to the MCP specification, notifications from client to server are allowed.

In the case of notification/roots/list_changed, if the client supports listChanged, sending this notification is MUST.
The initialized notification is part of the lifecycle specification.

Since the MCP spec does not define any specific error-handling requirements for notification/roots/list_changed on the server side, it seems more appropriate to accept the notification rather than return a "Missing handler" error.

Because the server is stateless, we have implemented notification/roots/list_changed as a no-op.

## How Has This Been Tested?
I wrote test code.

## Breaking Changes
nop

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
@@ -133,8 +133,24 @@ public class McpStatelessAsyncServer {
 
 		this.protocolVersions = new ArrayList<>(mcpTransport.protocolVersions());
 
-		McpStatelessServerHandler handler = new DefaultMcpStatelessServerHandler(requestHandlers, Map.of());
+		Map<String, McpStatelessNotificationHandler> notificationHandlers = prepareNotificationHandlers();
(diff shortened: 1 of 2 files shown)


Kehrlann (changes requested):
Thanks for your contribution! This is directionally good, but the tests need reworking.

Both the tests in this PR pass even without your changes, because of the server's default `handleNotification` which does return `Mono.empty`.

The only difference with and without your code is suppressing a WARN log message. 
Also, there's no need to use mock requests, as the SDK's client can trigger the notifications.

Please update your test to:
1. Use the bundled `client`
1. Read from the logs and check the absence of the specific warn message

Please also consider adding an appropriate log message for `roots/list_changed`. After all the client is sending a notification we can't handle, we should at least log it, probably at the DEBUG level, not WARN.

Kehrlann (approved):

Kehrlann:
Polished and merged, thanks for your contribution!

Issue #837, opened 2026-02-26
Implement resource subscribe/unsubscribe feature


tzolov:
Allow the client to request `resources/updated` notifications from the server whenever a particular resource, identified by it resource URI, changes: https://modelcontextprotocol.io/specification/2025-11-25/server/resources#subscriptions

tzolov:
Related issue: https://github.com/modelcontextprotocol/java-sdk/issues/776

Issue #776, opened 2026-02-06
McpStatelessSyncServer should preemptively register a no-op handler for resources/subscribe


rombert:
**Bug description**
Certain non-compliant MCP clients try to subscribe for resource updates even if the MCP server reports no support for subscription.

I have seen this behaviour with Cursor 2.4.27 and 2.4.28.

This leads to the MCP server not intialising in Cursor due to internal errors: Failed to handle request: Missing handler for request type: resources/subscribe" .

**Environment**
MCP Java SDK 0.17.0
Cursor

**Steps to reproduce**

I am using a pretty standard MCP server setup 

```java
        syncServer = McpServer.sync(transportProvider)
                .serverInfo(config.serverTitle(), serverVersion)
                .jsonMapper(jsonMapper)
                .jsonSchemaValidator(new DefaultJsonSchemaValidator())
                .instructions(config.instructions())
                .completions(completions)
                .capabilities(ServerCapabilities.builder()
                        .tools(false)
                        .prompts(false)
                        .resources(false, false)
                        .completions()
                        .build())
                .build();
````

When connecting to the server from Cursor the connection fails

> 2026-02-06 15:15:15.122 [error] Client error for command Streamable HTTP error: Error POSTing to endpoint: (...) "message":"Failed to handle request: Missing handler for request type: resources/subscribe","suppressed":[],"localizedMessage":"Failed to handle request: Missing handler for request type: resources/subscribe"}

**Expected behavior**

I would expect the server to gracefully handle non-compliant clients that try to subscribe to updates. Failing that, I should be able to register a handler myself.

However, the API does not allow that.

I had to resort to reflectiont to access HttpServletStatelessServerTransport → mcpHandler → requestHandlers and then register a no-op handler

```java
handlers.put(McpSchema.METHOD_RESOURCES_SUBSCRIBE, (context, params) -> Mono.just(Map.of()));
```

This is brittle and I would welcome either:
- a default handler that prevents this error from happening
- the ability to register handlers directly using a low-level API

**Minimal Complete Reproducible example**
N/A, sorry. But if the problem is not clear from the report let me know and I'll set something up.

chemicL:
This should be resolved via #839

rombert:
> This should be resolved via [#839](https://github.com/modelcontextprotocol/java-sdk/pull/839)

Thanks! I'll keep an eye open for the next release and will test it then.

PR #838, opened 2026-02-26
Add resource subscribe/unsubscribe support to McpAsyncServer

tzolov:
- Register resources/subscribe and resources/unsubscribe handlers when the subscribe server capability is enabled
- Track active subscriptions in a ConcurrentHashMap keyed by resource URI
- Emit notifications/resources/updated when addResource() is called for a subscribed URI; also send listChanged notification if enabled
- Add ResourceSubscriptionTests covering the capability guard, subscribe/ unsubscribe happy paths, and the updated notification flow

Note: as with some other notifications this impl. is not bound to user sessions

Resolves: #837, #776

### DIFF
--- a/conformance-tests/conformance-baseline.yml
+++ b/conformance-tests/conformance-baseline.yml
@@ -2,8 +2,6 @@
 # This file lists known failing scenarios that are expected to fail until fixed.
 # See: https://github.com/modelcontextprotocol/conformance/blob/main/SDK_INTEGRATION.md
 
-server:
-
 client:
   # SSE retry field handling not implemented
   # - Client does not parse or respect retry: field timing


mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -109,6 +109,8 @@ public class McpAsyncServer {
 
 	private final ConcurrentHashMap<String, McpServerFeatures.AsyncResourceTemplateSpecification> resourceTemplates = new ConcurrentHashMap<>();
 
+	private final Map<String, McpSchema.SubscribeRequest> resourceSubscriptions = new ConcurrentHashMap<>();
```
chemicL, line 112: 1. This construct should include a session ID so that we know which client to notify about which resource.
2. The `SubscribeRequest` is not needed right now, but perhaps the meta-information can be of use for later. For now I'd keep the field as `Map<String, Set<String>>` of uri -> set(session_id).
3. Use this map from `io.modelcontextprotocol.server.McpAsyncServer#notifyResourcesUpdated` to filter the clients to notify about a resource update (look at uri from `io.modelcontextprotocol.spec.McpSchema.ResourcesUpdatedNotification#uri`)

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -560,15 +567,26 @@ public Mono<Void> addResource(McpServerFeatures.AsyncResourceSpecification resou
...
+			}
+
+			if (this.serverCapabilities.resources().subscribe()
+					&& this.resourceSubscriptions.containsKey(resourceUri)) {
+				Mono<Void> updated = this.notifyResourcesUpdated(new ResourcesUpdatedNotification(resourceUri));
+				if (this.serverCapabilities.resources().listChanged()) {
+					return updated.then(this.notifyResourcesListChanged());
+				}
+				return updated;
 			}
```
chemicL, line 587: This logic is unnecessary. A resource is not updated when it's added but when it's modify. This code block should be removed.

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -734,6 +753,33 @@ private McpRequestHandler<McpSchema.ReadResourceResult> resourcesReadRequestHand
...
 	}
 
+	private McpRequestHandler<Object> resourcesSubscribeHandler() {
+		return (ex, params) -> {
+			McpSchema.SubscribeRequest resourceSubscribeRequest = jsonMapper.convertValue(params, new TypeRef<>() {
+			});
+
+			var resourceUri = resourceSubscribeRequest.uri();
+
+			this.resourceSubscriptions.put(resourceUri, resourceSubscribeRequest);
```
chemicL, line 763: use `ex.sessionId()` for the mapping.

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -734,6 +753,33 @@ private McpRequestHandler<McpSchema.ReadResourceResult> resourcesReadRequestHand
...
+
+	private McpRequestHandler<Object> resourcesUnsubscribeHandler() {
+
+		return (ex, params) -> {
+			McpSchema.SubscribeRequest resourceSubscribeRequest = jsonMapper.convertValue(params, new TypeRef<>() {
+			});
+
+			var resourceUri = resourceSubscribeRequest.uri();
+
+			this.resourceSubscriptions.remove(resourceUri);
```
chemicL, line 777: Only remove the particular session, not the resource uri from the mapping - only remove it if no sessions are left that have a subscription.

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
chemicL: I realized this requires the `McpTransportProvider` to expose an API that accepts the sessionId or a set of them. For subscription related notifications, it has to be the listening stream (because the act of subscribing which has to receive a response, so even if the request upgrades to an SSE stream, it has to be closed after the response). So it makes sense to expose it at this level.

chemicL:
I created a PR demonstrating the above feedback: #839 Since the approach is completely different, let's close this one if you agree @tzolov

chemicL:
Closing in favour of the merged #839.

PR #749, opened 2026-01-14
fix:  Remove JsonSchema and use a Map for inputSchema  to support json schemas dialect

bilaloumehdi:
Following the [Model Context Protocol specification](https://modelcontextprotocol.io/specification/2025-11-25/server/tools) for [JSON schemas usage](https://modelcontextprotocol.io/specification/2025-11-25/basic#json-schema-usage), the schema should, by default follow the 2020-12 dialect:  [json-schema-2020-12](https://json-schema.org/draft/2020-12/schema) if no `$schema` specified.

## Motivation and Context

The need for this change arose from the requirement for a tool that can handle one of two properties objects. This can be achieved by defining an `inputSchema` with `oneOf` at the top level, for example: 
```
 inputSchema: { 
  type: object, 
  oneOf: [ 
    { properties: {}, required: ["test"] },
    { properties: {}, required: ["test", "command"] }
  ] 
}
```

### The current state
 The schema is currently deserialized to a `JsonSchema` record and ignores `oneOf` at the top level. A specific fix for this issue would be to add a new field to the `JsonSchema` record for `oneOf`; however, this approach does not scale well if additional top-level entries (such as `allOf`) are needed. This PR shifts to using a `Map<String, Object>` to deserialize the `inputSchema`

### Next Steps and improvements
I suggest adding validation for schemas based on the schema specifications from the SDK, not only checking for serialization errors. This would help fail fast and provide feedback about any issues in the defined schemas. 

## How Has This Been Tested?

Existing Tests have been adjusted for these changes, and a new test has been introduced with json schemas that have `oneOf` at the top level to verify that deserialization do not ignore it.

## Breaking Changes

Users would need to change the tool's inputSchema data type

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #845, opened 2026-03-04
feat: implement size-limited LRU cache for JSON Schema validation

joaquinescalante23:
I noticed that the current `DefaultJsonSchemaValidator` implementations (Jackson 2 and 3) were using an unbounded `ConcurrentHashMap` for schema caching, which could lead to memory leaks in long-running processes if exposed to a large number of dynamic schemas.

This PR addresses the `TODO` in both modules by:
- Replacing the unbounded map with a thread-safe **LRU Cache** (defaulting to 1024 entries) using `LinkedHashMap`.
- Making the cache size configurable via constructors for greater flexibility.
- Improving the `generateCacheKey` logic to use a stable JSON representation of the schema instead of the default `hashCode()`, preventing potential collisions and ensuring cache hits are reliable regardless of map order.

These changes significantly improve the resilience and resource management of the SDK when validating complex MCP tool outputs.

### DIFF
--- a/mcp-json-jackson2/src/main/java/io/modelcontextprotocol/json/schema/jackson2/DefaultJsonSchemaValidator.java
+++ b/mcp-json-jackson2/src/main/java/io/modelcontextprotocol/json/schema/jackson2/DefaultJsonSchemaValidator.java
@@ -3,9 +3,10 @@
  */
 package io.modelcontextprotocol.json.schema.jackson2;
 
+import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


joaquinescalante23:
I've been following the architectural changes proposed in **PR #749** by @tzolov. While that PR focuses on the extensibility of the validator through suppliers, this PR addresses a critical **resource management issue** in the current default implementation.

Specifically, the existing `DefaultJsonSchemaValidator` (Jackson 2 and 3) lacks a cache eviction policy, which poses a memory exhaustion risk in long-running applications or when dealing with dynamic tool schemas. My proposal introduces a size-configurable **LRU Cache** that would make the new supplier-based architecture much more robust by default.

I'm happy to rebase or adjust this PR to align with the new supplier patterns once #749 is merged. Let me know if you'd like me to integrate these resilience improvements into the new supplier structure!

PR #849, opened 2026-03-04
Add all protocol versions to StdioServerTransportProvider

### LINKED ISSUE #751: Does StdioServerTransportProvider support protocol version 2025-06-18 ?
Hi there, I noticed that the StdioServerTransportProvider reports only protocol version 2024-11-05 as supported. Does that mean it does not support / respect newer protocol versions or does it mean that since that version the spec was not changed for stdio? If I understand the spec correctly, the server should respond with the latest version it supports: https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle . Is that assumption correct? Thanks! Best, Timo

https://github.com/modelcontextprotocol/java-sdk/blob/f7a460fb505aeba239165f9bc4ef508a77ba102a/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/StdioServerTransportProvider.java#L87

rameshreddy-adutla:
## Summary

Fixes #751

`StdioServerTransportProvider.protocolVersions()` only reported `MCP_2024_11_05`, causing version negotiation to fail with newer clients. Since the stdio transport layer is unchanged across all protocol versions, it should advertise support for all of them — matching the behavior of `HttpServletStreamableServerTransportProvider`.

## Changes

- **`StdioServerTransportProvider.java`**: Updated `protocolVersions()` to return all four protocol versions (`2024-11-05`, `2025-03-26`, `2025-06-18`, `2025-11-25`).

## Before/After

```java
// Before — only the original version
return List.of(ProtocolVersions.MCP_2024_11_05);

// After — all supported versions (matches Streamable HTTP transport)
return List.of(ProtocolVersions.MCP_2024_11_05, ProtocolVersions.MCP_2025_03_26,
        ProtocolVersions.MCP_2025_06_18, ProtocolVersions.MCP_2025_11_25);
```

## Testing

All 9 `StdioServerTransportProviderTests` pass.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/StdioServerTransportProvider.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/StdioServerTransportProvider.java
@@ -84,7 +84,8 @@ public class StdioServerTransportProvider implements McpServerTransportProvider 
 
 	@Override
 	public List<String> protocolVersions() {
-		return List.of(ProtocolVersions.MCP_2024_11_05);
+		return List.of(ProtocolVersions.MCP_2024_11_05, ProtocolVersions.MCP_2025_03_26,
+				ProtocolVersions.MCP_2025_06_18, ProtocolVersions.MCP_2025_11_25);
 	}
 
 	@Override


Kehrlann:
Protocol compat moved to `McpTransport` directly.

Issue #240, opened 2025-05-15
Handle 401 Unauthorized on WebFluxSseClientTransport with client.initialize


jaketothepast:
Please do a quick search on GitHub issues first, the feature you are about to request might have already been requested.

**Expected Behavior**

When calling client.initialize(), I'd like to know if a 401 happened. That way, I can initiate an oauth login for the current user.

**Current Behavior**

In the current code, 401 is captured as a fatal sse error, with no way to attach an error callback.

```java
	/**
	 * Initializes and starts the inbound SSE event processing. Establishes the SSE
	 * connection and sets up event handling for both message and endpoint events.
	 * Includes automatic retry logic for handling transient connection failures.
	 */
	// visible for tests
	protected Flux<ServerSentEvent<String>> eventStream() {// @formatter:off
		return this.webClient
			.get()
			.uri(this.sseEndpoint)
			.accept(MediaType.TEXT_EVENT_STREAM)
			.retrieve()
			.bodyToFlux(SSE_TYPE)
			.retryWhen(Retry.from(retrySignal -> retrySignal.handle(inboundRetryHandler)));
	} // @formatter:on

	/**
	 * Retry handler for the inbound SSE stream. Implements the retry logic for handling
	 * connection failures and other errors.
	 */
	private BiConsumer<RetrySignal, SynchronousSink<Object>> inboundRetryHandler = (retrySpec, sink) -> {
		if (isClosing) {
			logger.debug("SSE connection closed during shutdown");
			sink.error(retrySpec.failure());
			return;
		}
		if (retrySpec.failure() instanceof IOException) {
			logger.debug("Retrying SSE connection after IO error");
			sink.next(retrySpec);
			return;
		}
		logger.error("Fatal SSE error, not retrying: {}", retrySpec.failure().getMessage());
		sink.error(retrySpec.failure());
	};

```

I want to capture the error, so that I can initiate the oauth flow to log the user in.

**Context**

I am building a way for users of our app to connect to MCP servers, and some servers (most really) will be protected via oauth

jaketothepast:
Just as a hack to get around this, I added `errorSink` to the Client specs, so that I could report on an error within initialization and block on it in my code.

chemicL:
[...]

PR #861, opened 2026-03-10
HttpClientStreamHttpTransport: add authorization error handler

Kehrlann:
## Motivation and Context

In the context of authorization management, we need to handle authorization errors (HTTP 401 and 403) in the client transports. This PR introduces a hook to react to these specific errors. See gh-240.

The hook exposes a "retry" mechanism, so that handlers can decide the request must be replayed, for example, after performing a dynamic client registration, or after updating the scopes in the "step up" scenario. The retry mechanism is unbounded, retry limitation is the caller's responsibility.

This PR does not target the SSE transport as we don't intend to evolve it.

## How Has This Been Tested?

Test with spring's `mcp-security` project.

## Breaking Changes

n/a

## Types of changes

- [ ] Documentation update


mcp-core/src/main/java/io/modelcontextprotocol/client/transport/customizer/McpHttpClientAuthorizationErrorHandler.java
```diff
@@ -0,0 +1,98 @@
...
+public interface McpHttpClientAuthorizationErrorHandler {
+
+	/**
+	 * Handle HTTP error, and signal whether the HTTP request should be retried or not.
+	 * @param responseInfo the HTTP response information
+	 * @param context the MCP client transport context
+	 * @return {@link Publisher} emitting true if the original request should be replayed,
+	 * false otherwise.
+	 */
+	Publisher<Boolean> handle(HttpResponse.ResponseInfo responseInfo, McpTransportContext context);
```
chemicL, line 42: This method seems like it's aimed at the end user. I think the name is fine and semantics are clear. What I'd suggest documenting is what happens when there is an error from the returned `Publisher`.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/customizer/McpHttpClientAuthorizationErrorHandler.java
```diff
@@ -0,0 +1,98 @@
...
+	/**
+	 * Handle HTTP error, and optionally retry the HTTP request. Defaults to
+	 * {@link #handle(HttpResponse.ResponseInfo, McpTransportContext)}, and uses the
+	 * boolean from the return value to decide whether it should retry the request.
+	 * @param responseInfo the HTTP response information
+	 * @param context the MCP client transport context
+	 * @param retryHandler the handler to use to retry the HTTP request.
+	 * @return a {@link Publisher} to signal either an error or a retry
+	 */
+	default Publisher<Void> handle(HttpResponse.ResponseInfo responseInfo, McpTransportContext context,
```
chemicL: I think this method is rather aimed at transport implementors, not end users. It has different semantics, it returns `Publisher<Void>` instead of `Publisher<Boolean>` from the other method. I'd suggest calling it `onAuthError` and documenting what is the intention for a return `Publisher` that errors.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/customizer/McpHttpClientAuthorizationErrorHandler.java
```diff
@@ -0,0 +1,98 @@
...
+	 * Will be subscribed on {@link Schedulers#boundedElastic()}. The handler may be
+	 * blocking.
+	 * @param handler the synchronous handler
+	 * @return an async handler
+	 */
+	static McpHttpClientAuthorizationErrorHandler fromSync(Sync handler) {
+		return (info, context) -> {
+			try {
+				var shouldRetry = handler.handle(info, context);
+				return Mono.just(shouldRetry).subscribeOn(Schedulers.boundedElastic());
```
chemicL: The `subscribeOn` use is ineffective. You should wrap the `handler.handle` inside a `Mono.fromCallable` and on top of that chain add the `subscribeOn` operator to avoid blocking a non-blocking thread.

Kehrlann: Duh! Thanks a lot.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/customizer/McpHttpClientAuthorizationErrorHandler.java
```diff
@@ -0,0 +1,116 @@
...
+	 * uses the boolean from the return value to decide whether it should retry the
+	 * request.
+	 * @param responseInfo the HTTP response information
+	 * @param context the MCP client transport context
+	 * @param retryAction handler to retry the original request
+	 * @param defaultError handler to emit an error
+	 * @return a {@link Publisher} to signal either an error or a retry
+	 */
+	default Publisher<Void> onAuthorizationError(HttpResponse.ResponseInfo responseInfo, McpTransportContext context,
+			Publisher<Void> retryAction, Publisher<Void> defaultError) {
```
chemicL: For the default error:
* currently the exception is eagerly created so there is no need for wrapping it in a Mono
* instead, a Supplier<Exception> can be used which would lazily be evaluated and would capture the actual stack at invocation time.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/customizer/McpHttpClientAuthorizationErrorHandler.java
```diff
@@ -0,0 +1,116 @@
...
+	 * Handle authorization error (HTTP 401 or 403), and signal whether the HTTP request
+	 * should be retried or not. If the publisher returns true, the original transport
+	 * method (connect, sendMessage) will be replayed with the original arguments.
+	 * Otherwise, the transport will throw an {@link McpHttpClientTransportException},
+	 * indicating the error status.
+	 * <p>
+	 * If the returned {@link Publisher} errors, the error will be propagated to the
+	 * calling method, to be handled by the caller.
+	 * <p>
+	 * The caller is responsible for bounding the number of retries.
```
chemicL: Can you add a test that shows how one would limit the retries? We should follow-up with another PR for the documentation about this case I suppose.

PR #867, opened 2026-03-13
fix: ensure openConnections.dispose() runs in closeGracefully even if the handshake logic throws an exception or fails.

noxymon:
This PR ensures that internal connections in DefaultMcpTransportSession are always disposed of during the graceful shutdown process, even if the handshake logic throws an exception or fails.

  ## Motivation and Context
  
  The previous implementation of closeGracefully() would skip the connection disposal step if the onClose function emitted an error. This led to resource leaks in scenarios where the server was unresponsive or the handshaking handshake handshaked
  wrong. This fix uses Reactor's doFinally to guarantee cleanup.

  This PR closes https://github.com/modelcontextprotocol/java-sdk/issues/547.

  ## How Has This Been Tested?
  
   - Implemented DefaultMcpTransportSessionTests.
   - Verified that openConnections.dispose() is called when the graceful handshake succeeds.
   - Verified that openConnections.dispose() is called when the graceful handshake fails with an exception.

  ## Breaking Changes
  
  None.

  ## Types of changes
  
   - [ ] Documentation update

## Checklist
  
   - [x] I have read the [MCP Documentation (https://modelcontextprotocol.io)

  ## Additional context
  
  Used Mono.defer and doFinally to ensure reliable execution of the cleanup logic regardless of the reactive signal type.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/spec/DefaultMcpTransportSession.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/spec/DefaultMcpTransportSession.java
@@ -77,8 +77,9 @@ public class DefaultMcpTransportSession implements McpTransportSession<Disposabl
 
 	@Override
 	public Mono<Void> closeGracefully() {
-		return Mono.from(this.onClose.apply(this.sessionId.get()))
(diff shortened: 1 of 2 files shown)


noxymon:
i think, the solution duplicated by #868

Issue #697, opened 2025-11-26
SEP-1303: Input Validation Errors as Tool Execution Errors


Kehrlann:
Tracking: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1303

ashakirin:
started working on the enhancement (close to https://github.com/modelcontextprotocol/java-sdk/issues/689)

ashakirin:
I'd like to add some additional context: currently, the Java SDK [validates tool output parameters](https://github.com/modelcontextprotocol/java-sdk/blob/cbb235fd32ea29c93f07642a7a6f83893672c8ec/mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java#L416), against the outputSchema, but does not validate tool input parameters against the inputSchema by default. According to the [MCP specification](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#security-considerations), servers MUST validate all tool inputs.

As part of this issue, we plan to:
1. Implement input schema validation for tool arguments
2. Enable this validation by default to align with the specification requirements

Please note: This change may not be backward compatible with existing implementations. Some tools that currently work may fail validation if their input schemas are not correctly defined. However, this change is necessary to ensure compliance with the MCP specification.

PR #873, opened 2026-03-18
feat!: #697 added tool input arguments validation

ashakirin:
Added tool input arguments validation causes Tool Execution Error accordingly SEP-1303. 
It is breaking change, because tool input validation is activated by default

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/util/ToolInputValidator.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/util/ToolInputValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024-2024 the original author or authors.
+ * Copyright 2026-2026 the original author or authors.
  */
 
 package io.modelcontextprotocol.util;
--- a/mcp-core/src/main/java/io/modelcontextprotocol/util/ToolInputValidator.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/util/ToolInputValidator.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ */
+
+package io.modelcontextprotocol.util;
+
+import java.util.List;
+import java.util.Map;
+
+import io.modelcontextprotocol.json.McpJsonMapper;
+import io.modelcontextprotocol.json.TypeRef;
+import io.modelcontextprotocol.json.schema.JsonSchemaValidator;
+import io.modelcontextprotocol.spec.McpSchema;
+import io.modelcontextprotocol.spec.McpSchema.CallToolResult;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Validates tool input arguments against JSON schema.
+ *
+ * @author Andrei Shakirin
+ */
+public final class ToolInputValidator {
+
+	private static final Logger logger = LoggerFactory.getLogger(ToolInputValidator.class);
+
+	/**
+	 * System property to disable tool input validation. Set to "false" to disable.
+	 * Default is true (validation enabled).
+	 */
+	public static final String VALIDATE_TOOL_INPUTS_PROPERTY = "io.modelcontextprotocol.validateToolInputs";
+
+	private ToolInputValidator() {
+	}
+
+	/**
+	 * Returns whether validation is enabled by default based on system property.
+	 * @return true if validation is enabled (default), false if disabled
+	 */
+	public static boolean isEnabledByDefault() {
+		return !"false".equalsIgnoreCase(System.getProperty(VALIDATE_TOOL_INPUTS_PROPERTY));
+	}
+
+	/**
+	 * Validates tool arguments against the tool's input schema.
+	 * @param tool the tool definition containing the input schema
+	 * @param arguments the arguments to validate
+	 * @param validateToolInputs whether validation is enabled
+	 * @param jsonMapper the JSON mapper for schema conversion
+	 * @param validator the JSON schema validator (may be null)
+	 * @return CallToolResult with isError=true if validation fails, null if valid or
+	 * validation skipped
+	 */
+	public static CallToolResult validate(McpSchema.Tool tool, Map<String, Object> arguments,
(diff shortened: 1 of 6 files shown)


mcp-core/src/main/java/io/modelcontextprotocol/util/ToolInputValidator.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
```
Kehrlann:
```suggestion
 * Copyright 2026-2026 the original author or authors.
```

mcp-test/src/test/java/io/modelcontextprotocol/server/ToolInputValidationIntegrationTests.java
```diff
@@ -0,0 +1,270 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
```
Kehrlann:
```suggestion
 * Copyright 2026-2026 the original author or authors.
```

mcp-test/src/test/java/io/modelcontextprotocol/server/ToolInputValidationIntegrationTests.java
```diff
@@ -0,0 +1,270 @@
...
+import reactor.core.publisher.Mono;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for tool input validation against JSON schema. Validates that input validation
+ * errors are returned as Tool Execution Errors (isError=true) rather than Protocol
+ * Errors, per MCP specification.
+ *
+ * @author Alireza Khoram
```
Kehrlann: Is this author tag intended?

mcp-test/src/test/java/io/modelcontextprotocol/server/ToolInputValidationIntegrationTests.java
Kehrlann:
```suggestion
 * Copyright 2026-2026 the original author or authors.
```

mcp-core/src/main/java/io/modelcontextprotocol/util/ToolInputValidator.java
```diff
@@ -0,0 +1,74 @@
...
+	 * validation skipped
+	 */
+	public static CallToolResult validate(McpSchema.Tool tool, Map<String, Object> arguments,
+			boolean validateToolInputs, McpJsonMapper jsonMapper, JsonSchemaValidator validator) {
+		if (!validateToolInputs || tool.inputSchema() == null || validator == null) {
+			return null;
+		}
+		Map<String, Object> inputSchema = jsonMapper.convertValue(tool.inputSchema(),
+				new TypeRef<Map<String, Object>>() {
+				});
```
Kehrlann: I wonder if we should leave this marshalling on the hot path. Look at what we've done in `DefaultJsonSchemaValidator`, there's a schema cache to avoid specifically doing to much of these operations.

Maybe we should have something similar here? Or maybe we should update the `JsonSchemaValidator` to also validate `JsonSchema` objects?

Kehrlann (commented):
Now that I'm revisiting this, I'm wondering whether we really want this system property to toggle this on or off.

We're going to ship this in 2.0, let it be a breaking change - after all, [servers MUST validate tool inputs](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#security-considerations).

(And we should probably revisit tool name validation too, then)

mcp-core/src/main/java/io/modelcontextprotocol/util/ToolInputValidator.java
Kehrlann: #886 points out similar issues ; there are open PRs about it https://github.com/modelcontextprotocol/java-sdk/pull/749

mcp-test/src/test/java/io/modelcontextprotocol/server/ToolInputValidationIntegrationTests.java
ashakirin: ups, need to review headers more carefully ;)

mcp-core/src/main/java/io/modelcontextprotocol/util/ToolInputValidator.java
ashakirin: thanks for pointing it, I will wait until [#749](https://github.com/modelcontextprotocol/java-sdk/pull/749) is merged

ashakirin:
@Kehrlann regarding toggling: 
 - I think toggling makes sense for tool names, because validation is specified as [SHOULD/SHOULD NOT](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-names) in the spec;
 - For server tool input schema validation, I'm not sure about toggling. The spec says [MUST](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#security-considerations), so without validation the implementation would be non-spec-compliant. On the other hand, toggling could make SDK 2.0 adoption easier for some use cases. I agree to remove it.

Kehrlann:
For toggling, I'm in favor of leaving an option to turn it off. But let's get rid of the system property - we don't use that pattern anywhere else in the codebase.
If users don't want validation, they should indicate that in the client.

ashakirin:
@Kehrlann: Got it now — you meant we should reconsider using system properties specifically for tool name and input schema validation, not as a full toggle. That definitely makes sense for the input schema validation: I will remove the system property, but keep the toggle in the server builder. For tool names, server builder settings should be enough as well; exposing it as a system property is more the responsibility of a framework as the SDK. WDYT?

Kehrlann:
> For tool names, server builder settings should be enough as well; exposing it as a system property is more the responsibility of a framework as the SDK

Exactly.

For the tool names, you can do it in a separate PR.

ashakirin:
@Kehrlann: I reworked PR with Map<String, Object> input schema

Kehrlann:
Thanks for your contribution. For info I've simplified the integration test a bit.

Kehrlann (dismissed):

Kehrlann (approved):

PR #876, opened 2026-03-20
Fix concurrent sendMessage race in StdioClientTransport

### LINKED ISSUE #875: StdioClientTransport.sendMessage() fails with concurrent calls due to unicast sink race
## Problem

`StdioClientTransport.sendMessage()` uses `tryEmitNext()` on a unicast Reactor sink. When two threads call `sendMessage()` concurrently on the same transport, one call fails with `"Failed to enqueue message"` due to `FAIL_NON_SERIALIZED`.

This happens because `Sinks.many().unicast().onBackpressureBuffer()` is wrapped by Reactor's `SinkManySerialized`, which uses a CAS-based guard in `tryEmitNext()`. When two threads race on the CAS, the loser immediately gets `FAIL_NON_SERIALIZED` — the method does not retry.

## When this occurs

Any time two threads call `McpSyncClient.callTool()` (or any other method that calls `sendMessage()`) concurrently on the same client instance. This is a normal scenario when an MCP server proxies tool calls from parallel requests to a downstream MCP server via stdio transport.

## Current code

```java
// StdioClientTransport.java line 229
public Mono<Void> sendMessage(JSONRPCMessage message) {
    if (this.outboundSink.tryEmitNext(message).isSuccess()) {
        return Mono.empty();
    }
    else {
        return Mono.error(new RuntimeException("Failed to enqueue message"));
    }
}
```

The existing TODO comment at line 230 acknowledges this limitation:
> "we delegate the retry and the backpressure onto the caller. This might be enough for most cases."

## Proposed fix

Replace `tryEmitNext` with `emitNext` + `busyLooping`, which spin-retries on `FAIL_NON_SERIALIZED` until the competing thread finishes its CAS (microseconds):

```java
outboundSink.emitNext(message, Sinks.EmitFailureHandler.busyLooping(Duration.ofMillis(100)));
```

This is the pattern recommended by Reactor's own Javadoc for `emitNext`:
> "It would be possible for an EmitFailureHandler to busy-loop and optimistically wait for the contention to disappear"

## Reproduction

A test with two threads sending messages concurrently through the same transport reproduces the failure in ~90% of runs.

AbbasNS:
## Problem

`StdioClientTransport.sendMessage()` uses `tryEmitNext()` on a unicast Reactor sink (`Sinks.many().unicast().onBackpressureBuffer()`). When two threads call `sendMessage()` concurrently on the same transport, one fails with `"Failed to enqueue message"`.

### Root cause

Reactor wraps the unicast sink in `SinkManySerialized`, which uses a CAS-based guard in `tryEmitNext()`. When two threads race on the CAS, the loser immediately gets `FAIL_NON_SERIALIZED` — the method does not retry. The current code checks only `.isSuccess()` and returns `Mono.error` for any failure, discarding the retryable `FAIL_NON_SERIALIZED` result.

The existing TODO at line 230-235 acknowledges this:
> "we delegate the retry and the backpressure onto the caller. This might be enough for most cases."

It is not enough when an MCP server proxies concurrent tool calls to a downstream server via stdio — each call dispatches on a separate reactor thread and both call `sendMessage()` on the same transport.

## Fix

Replace `tryEmitNext` with `emitNext` + `Sinks.EmitFailureHandler.busyLooping(Duration.ofMillis(100))`.

`emitNext` with `busyLooping` spin-retries on `FAIL_NON_SERIALIZED` until the competing thread finishes its CAS (microseconds in practice). The 100ms duration is just a generous upper bound for pathological cases like GC pauses. This is the approach recommended by Reactor's own `emitNext` Javadoc:

> "It would be possible for an EmitFailureHandler to busy-loop and optimistically wait for the contention to disappear"

For non-retryable failures (`FAIL_TERMINATED`, `FAIL_CANCELLED`), `emitNext` throws `EmissionException` which is caught and converted to `Mono.error`, preserving the existing error behavior.

## Test

`StdioClientTransportConcurrencyTest` reproduces the race with two threads sending through the same transport concurrently (20 repetitions):

- **Before fix**: 19/20 fail
- **After fix**: 20/20 pass

Closes #875

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/StdioClientTransport.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/StdioClientTransport.java
@@ -227,16 +227,17 @@ public class StdioClientTransport implements McpClientTransport {
 
 	@Override
 	public Mono<Void> sendMessage(JSONRPCMessage message) {
-		if (this.outboundSink.tryEmitNext(message).isSuccess()) {
-			// TODO: essentially we could reschedule ourselves in some time and make
-			// another attempt with the already read data but pause reading until
-			// success
-			// In this approach we delegate the retry and the backpressure onto the
-			// caller. This might be enough for most cases.
+		try {
+			// busyLooping retries on FAIL_NON_SERIALIZED (concurrent tryEmitNext from
+			// another thread) instead of failing immediately. The contention window is
+			// microseconds (single CAS), so the spin resolves almost instantly; the
+			// duration is just a generous upper bound for pathological cases like GC
+			// pauses.
+			this.outboundSink.emitNext(message, Sinks.EmitFailureHandler.busyLooping(Duration.ofMillis(100)));
 			return Mono.empty();
 		}
-		else {
-			return Mono.error(new RuntimeException("Failed to enqueue message"));
+		catch (Sinks.EmissionException e) {
+			return Mono.error(new RuntimeException("Failed to enqueue message", e));
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


AbbasNS:
@chemicL @tzolov, let me know if this is the right way to contribute. your review and feedback is appreciated.

chemicL:
Addressed in #935.

PR #890, opened 2026-03-28
fix: route server-initiated requests via GET SSE in responseStream

armando-sepulveda:
When a client sends a POST request to initiate a tool/handler, the server uses ServerResponse.sse() (WebMVC) which calls request.startAsync(), putting Tomcat's OutputBuffer into 'suspended' async mode. While the request handler is executing, any writes to the POST SSE transport's OutputBuffer are buffered and never flushed — causing a deadlock for server-initiated requests like MCP Sampling.

This fix routes server-initiated requests via the GET SSE listening stream instead.

See detailed explanation in commit message.


armando-sepulveda:
@copilot apply changes based on the comments in [this thread](https://github.com/modelcontextprotocol/java-sdk/pull/890#pullrequestreview-4024592311)

PR #908, opened 2026-04-07
feat: Add SEP-1034 elicitation schema default values

### LINKED ISSUE #699: SEP-1034 : Support default values for all primitive types in elicitation schemas
Tracking: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1034

sainathreddyb:
Add client-side default application for elicitation schemas, matching the behavior specified in SEP-1034 and implemented in the TypeScript SDK.

Changes:
- Add applyDefaults field to Elicitation.Form capability record
- Add applyElicitationDefaults() in McpAsyncClient to fill missing fields from schema defaults when the client returns accept with incomplete content
- Add builder method elicitation(form, url, applyDefaults) to ClientCapabilities.Builder
- Add unit tests for applyElicitationDefaults covering all primitive types
- Add schema serialization tests for applyDefaults capability
- Add integration test verifying end-to-end default application flow

Closes modelcontextprotocol/java-sdk#699

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
@@ -15,6 +15,9 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import io.modelcontextprotocol.client.LifecycleInitializer.Initialization;
 import io.modelcontextprotocol.json.TypeRef;
 import io.modelcontextprotocol.json.schema.JsonSchemaValidator;
@@ -30,16 +33,14 @@ import io.modelcontextprotocol.spec.McpSchema.ElicitRequest;
 import io.modelcontextprotocol.spec.McpSchema.ElicitResult;
 import io.modelcontextprotocol.spec.McpSchema.GetPromptRequest;
 import io.modelcontextprotocol.spec.McpSchema.GetPromptResult;
-import io.modelcontextprotocol.util.ToolNameValidator;
 import io.modelcontextprotocol.spec.McpSchema.ListPromptsResult;
 import io.modelcontextprotocol.spec.McpSchema.LoggingLevel;
 import io.modelcontextprotocol.spec.McpSchema.LoggingMessageNotification;
 import io.modelcontextprotocol.spec.McpSchema.PaginatedRequest;
 import io.modelcontextprotocol.spec.McpSchema.Root;
 import io.modelcontextprotocol.util.Assert;
+import io.modelcontextprotocol.util.ToolNameValidator;
 import io.modelcontextprotocol.util.Utils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
@@ -561,8 +562,63 @@ public class McpAsyncClient {
 			ElicitRequest request = transport.unmarshalFrom(params, new TypeRef<>() {
 			});
 
-			return this.elicitationHandler.apply(request);
+			return this.elicitationHandler.apply(request).map(result -> {
+				// Apply defaults from schema when applyDefaults is enabled
+				if (result.action() == ElicitResult.Action.ACCEPT && result.content() != null
+						&& shouldApplyElicitationDefaults()) {
+					Map<String, Object> merged = new HashMap<>(result.content());
+					applyElicitationDefaults(request.requestedSchema(), merged);
+					return new ElicitResult(result.action(), merged, result.meta());
+				}
+				return result;
+			});
 		};
+	}
+
+	/**
+	 * Checks whether the client is configured to apply elicitation defaults.
+	 * @return true if the client capabilities indicate that defaults should be applied
+	 */
+	private boolean shouldApplyElicitationDefaults() {
+		if (this.clientCapabilities.elicitation() == null) {
+			return false;
+		}
+		McpSchema.ClientCapabilities.Elicitation.Form form = this.clientCapabilities.elicitation().form();
+		return form != null && Boolean.TRUE.equals(form.applyDefaults());
+	}
+
+	/**
+	 * Applies default values from the elicitation schema to the result content. For each
+	 * property in the schema that has a "default" value, if the corresponding key is
+	 * missing from the content map, the default value is inserted.
+	 * @param schema the requestedSchema from the ElicitRequest
+	 * @param content the mutable content map from the ElicitResult
+	 */
+	@SuppressWarnings("unchecked")
+	static void applyElicitationDefaults(Map<String, Object> schema, Map<String, Object> content) {
+		if (schema == null || content == null) {
+			return;
+		}
+
+		Object propertiesObj = schema.get("properties");
+		if (!(propertiesObj instanceof Map)) {
+			return;
+		}
+
+		Map<String, Object> properties = (Map<String, Object>) propertiesObj;
+		for (Map.Entry<String, Object> entry : properties.entrySet()) {
+			String key = entry.getKey();
+			Object propDef = entry.getValue();
+
+			if (!(propDef instanceof Map)) {
+				continue;
+			}
+
+			Map<String, Object> propMap = (Map<String, Object>) propDef;
+			if (!content.containsKey(key) && propMap.containsKey("default")) {
+				content.put(key, propMap.get("default"));
+			}
+		}
 	}
 
 	// --------------------------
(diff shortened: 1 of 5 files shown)


mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -561,10 +561,69 @@ private RequestHandler<ElicitResult> elicitationCreateHandler() {
...
 
-			return this.elicitationHandler.apply(request);
+			return this.elicitationHandler.apply(request).map(result -> {
+				// Apply defaults from schema when applyDefaults is enabled
+				if (result.action() == ElicitResult.Action.ACCEPT && result.content() != null
+						&& shouldApplyElicitationDefaults()) {
+					try {
+						applyElicitationDefaults(request.requestedSchema(), result.content());
+					}
+					catch (Exception e) {
```
utafrali: The `catch (Exception e)` silently swallows an `UnsupportedOperationException` if the user returns an unmodifiable map (e.g., `Map.of(...)` or `Collections.unmodifiableMap(...)`). Defaults are silently not applied, with only a DEBUG log. This is a real correctness hazard because nothing in the API contract prevents users from returning an unmodifiable content map.

The safer approach is to build a new merged map and return a new `ElicitResult` rather than mutating in-place:
```java
Map<String, Object> merged = new HashMap<>(result.content());
applyElicitationDefaults(request.requestedSchema(), merged);
return new ElicitResult(result.action(), merged);
```
This also removes the need for the try/catch entirely.

utafrali (commented):
The core logic is sound and test coverage is thorough, but there is a correctness bug: mutating the content map in-place silently fails when the user returns an unmodifiable map, with the exception swallowed at DEBUG level. There is also a design concern around `applyDefaults` leaking into the serialized MCP capabilities sent to the server, which risks breaking interoperability with strict servers.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -507,6 +522,20 @@ public Builder elicitation(boolean form, boolean url) {
...
+			 * Enables elicitation capability with form mode and applyDefaults setting.
+			 * @param form whether to support form-based elicitation
+			 * @param url whether to support URL-based elicitation
+			 * @param applyDefaults whether the client should apply schema defaults to
+			 * elicitation results
+			 * @return this builder
+			 */
+			public Builder elicitation(boolean form, boolean url, boolean applyDefaults) {
+				this.elicitation = new Elicitation(form ? new Elicitation.Form(applyDefaults) : null,
+						url ? new Elicitation.Url() : null);
```
utafrali, line 795: `applyDefaults` is a client-side SDK behavior flag, but because it lives inside `ClientCapabilities.Elicitation.Form`, it gets serialized and sent to the server in the `initialize` request. A strict MCP server that validates capabilities against the protocol schema will receive an unknown field it never declared. This could break interoperability.

Consider keeping `applyDefaults` purely in the SDK layer (e.g., a separate `McpClientOptions` / builder flag) rather than embedding it in the protocol capabilities record. Alternatively, document explicitly that this is an SDK extension field and that servers must tolerate unknown capability fields.

utafrali, line 795: The new builder overload `elicitation(boolean form, boolean url, boolean applyDefaults)` silently drops `applyDefaults` when `form == false`. A caller writing `elicitation(false, true, true)` gets a capability object where `applyDefaults` is never set, with no warning. Since `shouldApplyElicitationDefaults()` then returns `false`, defaults are never applied.

At minimum, add a guard:
```java
if (!form && applyDefaults) {
    throw new IllegalArgumentException("applyDefaults requires form to be true");
}
```
Or document the behavior clearly in the Javadoc.

mcp-test/src/main/java/io/modelcontextprotocol/AbstractMcpClientServerIntegrationTests.java
```diff
@@ -446,6 +446,76 @@ void testCreateElicitationSuccess(String clientType) {
...
+
+		var mcpServer = prepareAsyncServerBuilder().serverInfo("test-server", "1.0.0").tools(tool).build();
+
+		// Enable applyDefaults via the capability
+		try (var mcpClient = clientBuilder.clientInfo(new McpSchema.Implementation("Sample client", "0.0.0"))
+			.capabilities(ClientCapabilities.builder().elicitation(true, false, true).build())
+			.elicitation(elicitationHandler)
+			.build()) {
+
+			InitializeResult initResult = mcpClient.initialize();
```
utafrali, line 521: The integration test only exercises the `HashMap` path where mutation succeeds. The real-world failure mode (unmodifiable content map from `Map.of()`) has no test coverage. Consider adding a second variant:
```java
return new McpSchema.ElicitResult(McpSchema.ElicitResult.Action.ACCEPT, Map.of());
```
This would surface the `UnsupportedOperationException` bug described above, confirming the fix is needed.

mcp-test/src/test/java/io/modelcontextprotocol/spec/McpSchemaTests.java
```diff
@@ -1717,6 +1717,71 @@ void testElicitationCapabilityBuilderFormOnly() throws Exception {
...
 
+	@Test
+	void testElicitationCapabilityWithApplyDefaults() throws Exception {
+		McpSchema.ClientCapabilities capabilities = McpSchema.ClientCapabilities.builder()
+			.elicitation(true, false, true)
+			.build();
+
+		assertThat(capabilities.elicitation()).isNotNull();
+		assertThat(capabilities.elicitation().form()).isNotNull();
+		assertThat(capabilities.elicitation().form().applyDefaults()).isTrue();
```
utafrali, line 1896: `assertThat(json).contains("true")` is too broad — it will pass as long as any boolean `true` appears anywhere in the serialized JSON, not specifically as the value of `applyDefaults`. Use a more targeted assertion:
```java
assertThat(json).contains("\"applyDefaults\":true");
```
or use `assertThatJson(json).node("elicitation.form.applyDefaults").isEqualTo(true)`.

mcp-core/src/test/java/io/modelcontextprotocol/client/McpAsyncClientElicitationDefaultsTests.java
```diff
@@ -0,0 +1,151 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
```
utafrali: Copyright year says `2024-2024`. It should be `2024-2025` to match the other files in the project.

mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
sainathreddyb: Good Catch!

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
sainathreddyb: Good catch on the interoperability concern.

This follows the TypeScript SDK's design — `applyDefaults` is defined inside `FormElicitationCapabilitySchema` and gets serialized to the server during initialization:
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/core/src/types/schemas.ts#L312

The Java SDK already uses `@JsonIgnoreProperties(ignoreUnknown = true)` on capability records, so servers built with this SDK will tolerate it. I've added Javadoc documenting this as an SDK-level behavior flag, consistent with the TypeScript reference implementation.

That said, if the maintainers prefer keeping it outside the protocol record (separate builder flag), happy to refactor.

sainathreddyb: Good catch. Added an `IllegalArgumentException` guard — `elicitation(false, true, true)` now throws with "applyDefaults requires form to be true".

mcp-test/src/test/java/io/modelcontextprotocol/spec/McpSchemaTests.java
sainathreddyb: Fixed — changed to `assertThatJson(json).node("elicitation.form.applyDefaults").isEqualTo(true)`.

mcp-test/src/main/java/io/modelcontextprotocol/AbstractMcpClientServerIntegrationTests.java
sainathreddyb: Good point. Added an integration test variant where the client returns `Map.of()` (unmodifiable). This confirms the fix works since we now copy into a new HashMap before applying defaults.

chemicL:
Thank you, @sainathreddyb for the contribution. I have incorporated @utafrali feedback - thank you as well! And opened #976 that supersedes this PR.

PR #913, opened 2026-04-09
Fix SSE event classification to follow spec for missing event field

### LINKED ISSUE #885: Streamable HTTP client ignores SSE messages when event name is omitted
Disclaimer: I am not a native English speaker. This issue was drafted with AI translation assistance. I apologize for any awkward phrasing.

## Bug description

When using the Streamable HTTP transport, the Java client receives an SSE frame whose payload contains a valid JSON-RPC response, but the SSE event name is omitted by the server.

In standard SSE semantics, if the `event:` field is omitted, the event type defaults to `message`. However, the current Java SDK only parses the frame if `event.event()` is explicitly equal to `"message"`. If the event name is missing and Spring/WebFlux exposes it as `null`, the client drops the frame instead of treating it as a default `message` event.

As a result, the `initialize` response is ignored even though the client has already received the JSON-RPC result, and initialization eventually times out with:

`Client failed to initialize by explicit API call`

In my case, the server is `rmcp` over Streamable HTTP.

## Environment

- Spring AI: `2.0.0-M3`
- MCP Java SDK: `1.1.0`
- Transport: `WebClientStreamableHttpTransport` (via `spring-ai-starter-mcp-client-webflux`)
- Spring Boot: `4.0.4`
- Java: `25.0.2`
- Reactor Netty: `1.3.4`
- MCP server: `rmcp 1.2.0` (Rust, Streamable HTTP)
- OS: macOS

## Steps to reproduce

1. Run an MCP server over Streamable HTTP.
2. Make the server return a valid JSON-RPC `initialize` result as `text/event-stream`.
3. The SSE frame should contain only `data: ...` and omit the `event:` field. For example:

```text
   data: {"jsonrpc":"2.0","id":"1","result":{"protocolVersion":"2025-06-18","capabilities":{"tools":{}},"serverInfo":{"name":"test","version":"1.0.0"}}}
````

4. Start a Spring AI / MCP Java SDK client using `WebClientStreamableHttpTransport`.
5. Call `client.initialize()` (or let Spring AI auto-initialize the MCP client during startup).

## Expected behavior

The client should treat an SSE frame without an explicit `event:` field as a default `message` event and parse the JSON-RPC payload normally.

So the `initialize` response above should be accepted, and client initialization should complete successfully.

## Actual behavior

The client receives the SSE payload, but the event type is exposed as `null` and is not parsed as a JSON-RPC message.

As a result, the initialize response is effectively ignored, and initialization times out.

From debug logs:

```text
Received SSE event with type: ServerSentEvent [id = 'null', event='null', retry=null, comment='null', data={"jsonrpc":"2.0","id":"9642da98-0","result":{"protocolVersion":"2025-06-18","capabilities":{"tools":{}},"serverInfo":{"name":"rmcp","version":"1.2.0"},"instructions":"..."}}]
...
Caused by: java.util.concurrent.TimeoutException: Did not observe any item or terminal signal within 20000ms
...
Client failed to initialize by explicit API call

### PR DESCRIPTION
**Pull request by iaJingda**
Fix `HttpClientStreamableHttpTransport` silently dropping SSE frames that omit the `event:` field, in violation of the SSE specification.

## Motivation and Context

Per the [SSE specification (WHATWG HTML Living Standard §9.2.6)](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation):

> If the field name is "event": Set the `event type` buffer to `field value`. If the `event:` line is absent, the event type buffer is empty, and the event is dispatched as a `message` event.

The current `reconnect()` / GET SSE stream path in `HttpClientStreamableHttpTransport` uses a strict equality check:

```java
if (MESSAGE_EVENT_TYPE.equals(sseResponseEvent.sseEvent().event())) {
```

When a server emits a bare `data:` frame without an accompanying `event:` line (which is valid per the spec), `event()` is `null` and the frame is silently dropped. This manifests as server-initiated notifications never reaching the handler after initialization completes.

Note the asymmetry: the `sendMessage()` POST path does not check the event type at all and is unaffected, which is why `initialize()` succeeds on HttpClient. The bug only impacts server-pushed messages on the open GET stream.

Originally reported in #885 against the Spring `WebClientStreamableHttpTransport` variant, which was moved to Spring AI 2.0 in #805. This PR fixes the `HttpClient` variant that remains in this repository.

## How Has This Been Tested?

Added `HttpClientStreamableHttpTransportSseEventTypeTest` with parameterized unit tests covering:

- `null` and empty event names → parsed as message (the bug scenario)
- Explicit `"message"` → parsed as message (regression)
- Non-message events (`ping`, `error`, `notification`, `MESSAGE`/`Message` case variants, custom names) → not parsed as message

Verified locally:

```
./mvnw test -pl mcp-core -Dtest=HttpClientStreamableHttpTransportSseEventTypeTest
→ Tests run: 9, Failures: 0, Errors: 0, Skipped: 0

./mvnw validate -pl mcp-core
→ BUILD SUCCESS

./mvnw test -pl mcp-core
→ BUILD SUCCESS (no regressions)
```

## Breaking Changes

None. This is a pure bug fix that relaxes an overly strict check to match the SSE specification. Existing behavior for explicit `"message"` events is preserved, and non-message events are still ignored.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

**Implementation note**: I extracted the classification into a package-private static helper `isMessageEvent(String)` rather than inlining the null/empty check. This keeps the call site readable, documents the spec rule in one place (with a link to WHATWG §9.2.6 in the Javadoc), and enables direct unit testing without setting up an HTTP server.

**Scope**: This PR only addresses the `HttpClient` variant in this repository. The `WebClient` variant mentioned in #885 lives in Spring AI 2.0 after #805 and would need a separate fix there. See my [comment on #885](https://github.com/modelcontextprotocol/java-sdk/issues/885) for details on the asymmetry between the POST and GET paths.

Closes gh-885

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
@@ -112,6 +112,29 @@ public class HttpClientStreamableHttpTransport implements McpClientTransport {
 
 	public static int BAD_REQUEST = 400;
 
+	/**
+	 * Determines whether an SSE event should be treated as a "message" event carrying a
+	 * JSON-RPC payload.
+	 *
+	 * <p>
+	 * Per the <a href=
+	 * "https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation">
+	 * SSE specification (WHATWG HTML Living Standard §9.2.6)</a>, an event with no
+	 * explicit {@code event:} field MUST be dispatched as a {@code message} event by
+	 * default. This method applies that rule by treating {@code null} or empty event
+	 * names as equivalent to {@link #MESSAGE_EVENT_TYPE}.
+	 *
+	 * <p>
+	 * This alignment ensures interoperability with MCP servers that emit bare
+	 * {@code data:} frames without an accompanying {@code event:} line, which are valid
+	 * per the SSE spec.
+	 * @param eventName the SSE event name, which may be {@code null} or empty
+	 * @return {@code true} if the event should be parsed as a JSON-RPC message
+	 */
+	static boolean isMessageEvent(String eventName) {
+		return eventName == null || eventName.isEmpty() || MESSAGE_EVENT_TYPE.equals(eventName);
+	}
+
 	private final McpJsonMapper jsonMapper;
 
 	private final URI baseUri;
@@ -311,7 +334,7 @@ public class HttpClientStreamableHttpTransport implements McpClientTransport {
 											+ statusCode));
 						}
 						else if (statusCode >= 200 && statusCode < 300) {
-							if (MESSAGE_EVENT_TYPE.equals(sseResponseEvent.sseEvent().event())) {
+							if (isMessageEvent(sseResponseEvent.sseEvent().event())) {
 								String data = sseResponseEvent.sseEvent().data();
 								// Per 2025-11-25 spec (SEP-1699), servers may
 								// send SSE events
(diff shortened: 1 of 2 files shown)


Kehrlann (approved):
Thank you for your contribution!

Issue #738, opened 2025-12-31
I would like to maintain MCP server state persisted (sessionId)


poornimas-iyer:
Now the session id is maintained in memory in a map. But since I have my MCP server multiple pods running, I would like the state to be persisted in redis or in a DB.

I would like to override the session Validator as a parameter in the HttpServletStreamableServerTransportProvider

**Current Behavior**

Currently only the session creation can be customised but not maintaining the state fully.

**Context**

I am not able to implement stateful session in a distributed setup

amaan75:
@chemicL  is this waiting on [Stateless MCP #1442 ](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1442)
or are we free to raise a PR for this?
Also what is the general guideline for implementation,  are we free to use something like Spring data Redis, or is the expectation that the SDK exposes/defines interfaces/contracts while the default implementation creates inMemoryVariants for them while expecting the actual users to implement the Distributed Storage?


What store( i would assume redis (cluster enabled/disabled) is a preferred choice for writing a horizontal adapter, since there would be cases when SSE requests in MCP tie things down to  single sever/node and we would want to still make it so that  server initiated notifications are going to respond back with the correct SSE channel via the correct server/node/instance.

amaan75:
@chemicL I read how other sdks are planning  to tackle this, turns out instead of a horizontal adapter, which imho will probably also work for websocket assuming that some one does add some kind of a callback for task status update once the task kind completes, but that's beside the point, most sdks are banking on saving whatever state can be serialised in Redis or some store  like mongo/redis or something similar and for the pieces where a SSE or some other transport is required, they expect that to be recreated, if its missing against a node. what route will the java sdk take here?
Oh and Btw, what I mean by the horizontal adapter is some kind of a Redis pub/sub model, where instead of directly sending the server Side initiated thing ( like a notification) to the client, we put in some kind of a Redis like system which also keeps  using which we will then be able to decide  which instance/node/pod needs to act on it.
this is fairly common with systems like soketi, sockudo, which mostly deal with web sockets, in our case SSE, but also have other apis that are http. there is however a slight issue where, we probably need to think about how do we abstract out cases, where some responses can/should be sent in a sync way, but others probably need to go back via SSE, 

I could be wrong about this bit
```
[...]

Issue #376, opened 2025-07-06
Proposal: Add client state persistence API to allow MCP clients to survive restarts


asaikali:
# Motivation

## Problem Statement

The **current Java MCP SDK provides no mechanism for persisting client runtime state**. Without SDK support, it is **impossible** for applications to resume a prior session, reuse an existing access token, or skip redundant protocol negotiation after a restart.

## Example Failure Scenario

1. **Startup** – An `McpClient` sends an `initialize` request to `https://example.com/mcp`.  
2. **Session established** – The server responds with `Mcp-Session-Id: 1868a90c…`; the client continues issuing requests within that session. The server may create associated state (e.g., cached prompts, resources, intermediate results) tied to the session ID.  
3. **Crash / redeploy** – The JVM crashes or the container restarts.  
4. **Restart** – The new client instance knows nothing about the previous session ID, access token, or negotiated metadata. As a result, it must repeat the OAuth flow and send a new `initialize` request, forcing the server to discard prior state and degrading the user experience.

## Impact

Introducing a **first-class SDK abstraction** for state persistence will deliver:

* **Improve reliability** – Maintain continuity for long-running operations that depend on server-side session state
* **Better user experience** – Seamless resumption without re-authentication delays
* **Production readiness** – Essential for containerized environments with frequent deployments
* **Operational simplicity** – Let the host application decide *where* and *how* to store this data (e.g., files, JDBC, Redis, secret vault) while keeping the SDK storage-agnostic

## Use Cases

This feature is particularly valuable for:
- **Long-running applications** that need to survive restarts without losing context
- **Containerized environments** with frequent deployments and rolling updates
- **Enterprise applications** where remote MCP servers are desirable because they are easier to manage across thousands of employees
- **Batch processing systems** that require session continuity across job restarts

# Proposed API

Here is an example API to support discussions around this request. 

```java
package io.modelcontextprotocol.spec;

/** Wraps the OAuth 2.1 access token used in Authorization headers. */
public record TokenInfo(String accessToken) { }

/**
* Negotiated protocol version and server capabilities returned by the server.
*
* @param protocolVersion the agreed MCP protocol version
* @param capabilitiesJson server capabilities serialized as JSON (or any text format)
*/
public record ServerMetadata(String protocolVersion, String capabilitiesJson) { }
```

```java
package io.modelcontextprotocol.spec;

import java.util.Optional;

/**
 * Persists MCP client state so a Streamable-HTTP client can resume seamlessly
[...]

Issue #107, opened 2025-04-04
Session not found handling


tikal:
**Bug description**
After a server restart, a client calling the message endpoint with its session id it obtained via the previous running server instance gets a HTTP/1.1 500 Internal Server Error

**Environment**
java-sdk: 0.8.1 via Spring AI

**Steps to reproduce**
Call server message endpoint after a server restart, using previous sessionId

**Expected behavior**
Not too sure but it seems that the client should not interpret that there was indeed an internal server error because of a restart. It could either get a message which leads to recreating a session or there should be some sessions persistence between restart. There is a TODO around the code which returns the 500: https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java#L288

tikal:
Actually the TODO is irrelevant here, the error occurs before and comes from a NullPointerException due to the null value coming out of the concurrent hash map get method: https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-spring/mcp-spring-webflux/src/main/java/io/modelcontextprotocol/server/transport/WebFluxSseServerTransportProvider.java#L281

autoexebat:
**Environment**
Spring AI M8
Mcp SDK 0.9.0 

I am currently using Spring AI where the MCP client and MCP server communicate via SSE. When the MCP server restarts, the SSE connection between the client and server is naturally broken.

The issue arises because the MCP client holds a session ID, but when the MCP server restarts, it doesn’t receive that session ID from the client — meaning the connection can’t be properly re-established.

At the moment, the only way I have found to resolve this is to also restart the MCP client so it can initiate a new SSE connection and synchronize the session ID with the server again.

I am wondering if there’s a better approach to handle this situation — ideally, allowing the client to detect the server restart and re-send the existing session ID, so the connection can resume without requiring a full client restart.

Any insights or suggestions on improving this behavior would be greatly appreciated.

Kehrlann:
Sessions are gone from 2026-07-28. We don't plan on evolving the 2025-11-25 version of this project. Closing as `wont-fix`.

PR #914, opened 2026-04-09
feat: add McpSessionStore SPI for pluggable session storage

### LINKED ISSUE #274: McpServerSession lifecycle doesn't support distributed services

vlin529:
## Summary

Introduce a `McpSessionStore` interface that abstracts session storage in `HttpServletStreamableServerTransportProvider`, enabling custom implementations (e.g., Redis, JDBC, Hazelcast) for distributed and multi-instance MCP server deployments.

### Problem

Currently, `HttpServletStreamableServerTransportProvider` stores sessions in a hardcoded `ConcurrentHashMap`. This causes:
- **Sessions lost on restart** — clients get "Session not found" errors after server restart
- **No multi-instance support** — sessions created on Pod A are invisible to Pod B behind a load balancer
- **No extension point** — the `sessions` field is `private final`, methods are `private`, Builder has no `sessionStore()` option

Related issues: #274, #107, #738, #376

### Changes

- **`McpSessionStore`** — new interface with `save/get/remove/values/isEmpty/size/clear` operations
- **`InMemoryMcpSessionStore`** — default `ConcurrentHashMap`-backed implementation (preserves existing behavior)
- **`HttpServletStreamableServerTransportProvider`** — refactored to use `McpSessionStore` instead of hardcoded `ConcurrentHashMap`
- **Builder** — added `sessionStore(McpSessionStore)` method for custom store injection; defaults to `InMemoryMcpSessionStore` when not set

### Usage

```java
// Default (backward compatible — no changes needed)
HttpServletStreamableServerTransportProvider.builder()
    .jsonMapper(mapper)
    .build();

// Custom Redis-backed store
HttpServletStreamableServerTransportProvider.builder()
    .jsonMapper(mapper)
    .sessionStore(new RedisSessionStore(redisTemplate))
    .build();
```

### Non-breaking

This is a fully backward-compatible change. Existing code continues to work without modification — the default `InMemoryMcpSessionStore` is used when no custom store is provided.

### Test

All 274 existing mcp-core tests pass.

---

> **Note:** This PR only addresses `HttpServletStreamableServerTransportProvider` in `mcp-core`. The same pattern should be applied to the Spring WebMvc/WebFlux transport providers as a follow-up.

Closes #274
Relates to #107, #738, #376

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
@@ -11,7 +11,6 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantLock;
 
 import org.slf4j.Logger;
@@ -22,8 +21,10 @@ import io.modelcontextprotocol.json.TypeRef;
 import io.modelcontextprotocol.common.McpTransportContext;
 import io.modelcontextprotocol.server.McpTransportContextExtractor;
 import io.modelcontextprotocol.spec.HttpHeaders;
+import io.modelcontextprotocol.spec.InMemoryMcpSessionStore;
 import io.modelcontextprotocol.spec.McpError;
 import io.modelcontextprotocol.spec.McpSchema;
+import io.modelcontextprotocol.spec.McpSessionStore;
 import io.modelcontextprotocol.spec.McpStreamableServerSession;
 import io.modelcontextprotocol.spec.McpStreamableServerTransport;
 import io.modelcontextprotocol.spec.McpStreamableServerTransportProvider;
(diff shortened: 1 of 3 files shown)


galm1007:
@vlin529 there's a minor conflict with a log line, I tried to push a fix but I don't have permissions for it. Can you please take care of it?

PR #918, opened 2026-04-11
docs(config): add local environment example for development

however-yir:
Follow-up contribution in the "3 PRs per project" batch.

Scope:
- env focused, low-risk, docs/config oriented
- keeps existing behavior unchanged
- improves contributor onboarding and local reproducibility


however-yir:
Thanks for reviewing this PR. I want to clarify scope to reduce review overhead:\n\n- Goal: provide a local environment example to improve reproducible setup\n- Intentionally out of scope: feature logic changes, behavior changes, and dependency upgrades\n- Risk level: low (docs/config only)\n\nIf you prefer, I can immediately split/trim this further into even smaller PRs (one file per PR) to match your contribution policy.

PR #927, opened 2026-04-20
Consistent JSON forward/backward compatibility (2.0 foundation)

chemicL:
## Motivation and Context

The MCP specification evolves continuously; domain types must absorb
new fields and subtypes without breaking existing clients or servers.
On the 1.x line this is structurally prevented by sealed interfaces,
which make it impossible to add a permitted subtype without breaking
exhaustive pattern-match switch expressions in caller code. This
commit opens the 2.0 release line, where those constraints are lifted
and serialization is made self-contained — independent of any global
ObjectMapper configuration.

## Breaking Changes

Breaking changes for users migrating from 1.x

- Sealed interfaces removed from JSONRPCMessage, Request, Result,
  Notification, ResourceContents, CompleteReference and Content.
  Exhaustive switch expressions over these types must add a default
  branch.
- Prompt(name, description, null) no longer silently coerces null
  arguments to an empty list. Use Prompt.withDefaults() to preserve
  the previous behaviour.
- CompleteCompletion.total and .hasMore are now absent from the wire
  when not set, rather than being emitted as null.
- ServerParameters no longer carries Jackson annotations; it is an
  internal configuration class, not a wire type.

## What now works that did not before

- CompleteReference polymorphic dispatch (PromptReference vs
  ResourceReference) works through a plain readValue or convertValue
  call — no hand-rolled map inspection required.
- LoggingLevel deserialization is lenient: unknown level strings
  produce null instead of throwing.
- All domain records now tolerate unknown JSON fields, so a client
  built against an older SDK version will not fail when a newer server
  sends fields it does not yet recognise.
- Null optional fields are consistently absent from serialized output
  regardless of ObjectMapper configuration.

## Documentation

- CONTRIBUTING adds an "Evolving wire-serialized records" section: a
  9-rule recipe and example for adding a field safely.
- MIGRATION-2.0 documents all breaking changes listed above.

## Follow-up coming next

Several spec-required fields (e.g. JSONRPCError.code/message,
ProgressNotification.progress, CreateMessageRequest.maxTokens,
CallToolResult.content) are stored as nullable Java types without a
null guard. If constructed with null, the NON_ABSENT rule silently
omits them, producing invalid wire JSON without throwing. Fix: compact
canonical constructors with Assert.notNull, following the pattern
already in JSONRPCRequest.

## Types of changes

- [x] Documentation update

## Checklist


mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -1172,11 +1171,11 @@ public record Prompt( // @formatter:off
 		@JsonProperty("_meta") Map<String, Object> meta) implements Identifier { // @formatter:on
 
 		public Prompt(String name, String description, List<PromptArgument> arguments) {
-			this(name, null, description, arguments != null ? arguments : new ArrayList<>());
+			this(name, null, description, arguments, null);
 		}
 
 		public Prompt(String name, String title, String description, List<PromptArgument> arguments) {
-			this(name, title, description, arguments != null ? arguments : new ArrayList<>(), null);
+			this(name, title, description, arguments, null);
```
sainathreddyb: ---
**Potential NPE in the SDK's own completion handlers**

The `Prompt` constructors no longer coerce `null` arguments to an empty list (intentional, documented in MIGRATION-2.0). However, the completion validation code in both `McpAsyncServer` and `McpStatelessAsyncServer` still assumes `arguments()` is non-null:

```java
// McpAsyncServer.java ~line 987, McpStatelessAsyncServer.java ~line 734
if (!promptSpec.prompt()
    .arguments()   // can now return null
    .stream()      // NPE
    ...
```

This is the SDK's own code, so it would crash at runtime if a prompt is registered with null arguments and a completion request comes in. Might need a null guard here, something like:

```java
List<PromptArgument> args = promptSpec.prompt().arguments();
if (args != null && args.stream()
        .filter(arg -> arg.name().equals(argumentName))
        .findFirst()
        .isPresent()) {
```

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -1845,11 +1844,12 @@ public CreateMessageRequest(List<SamplingMessage> messages, ModelPreferences mod
 
 		public enum ContextInclusionStrategy {
 
-		// @formatter:off
-			@JsonProperty("none") NONE,
-			@JsonProperty("thisServer") THIS_SERVER,
-			@JsonProperty("allServers")ALL_SERVERS
-		} // @formatter:on
+			@JsonProperty("none")
```
tzolov, line 1847: the @formatter:off was used on purpose. Without it the content is unreadable.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -1959,29 +1959,36 @@ public record CreateMessageResult( // @formatter:off
 
 		public enum StopReason {
 
-		// @formatter:off
```
tzolov, line 1962: the @formatter:off was used on purpose. Without it the content is unreadable.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -2122,11 +2129,12 @@ public record ElicitResult( // @formatter:off
 
 		public enum Action {
 
-		// @formatter:off
```
tzolov, line 2125: same

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/ServerParameters.java
```diff
@@ -32,13 +30,10 @@ public class ServerParameters {
 						"SYSTEMDRIVE", "SYSTEMROOT", "TEMP", "USERNAME", "USERPROFILE")
 				: Arrays.asList("HOME", "LOGNAME", "PATH", "SHELL", "TERM", "USER");
 
-	@JsonProperty("command")
```
tzolov, line 35: If not mistaken the ServerParameters are modeling the Claude Desktop MCP json configuration (or subset of it) : https://modelcontextprotocol.io/docs/develop/connect-local-servers
Therefore the annotations serve this purpose.

tzolov (commented):
Please return the `@formatter:off` for the enum types because combined with the json properties the values are not readable. I've marked few places perhaps there are more. 

Additionally the JSONRPCMessage, JSONRPCResponse, JSONRPCError, JSONRPCNotification, JSONRPCRequest and the deserializeJsonRpcMessage are not really MCP Schema related but internal sdk implementation detail and can be moved in a separate class. I guess we can do it in another follow up PR.

tzolov:
Rebased, edited, squashed and merged at 4c8596306de503d374504e0fb80d6effada7d5f6

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/ServerParameters.java
chemicL: That's just a Claude thing, not specification thing. We don't use it to serialize/deserialize and we don't expect that kind of usage. These annotations should be removed.

chemicL:
> Please return the `@formatter:off` for the enum types because combined with the json properties the values are not readable. I've marked few places perhaps there are more.

I didn't catch that. In fact, there is an open issue to address this in the formatter: https://github.com/spring-io/spring-javaformat/issues/393 - would be great to have that.

> Additionally the JSONRPCMessage, JSONRPCResponse, JSONRPCError, JSONRPCNotification, JSONRPCRequest and the deserializeJsonRpcMessage are not really MCP Schema related but internal sdk implementation detail and can be moved in a separate class. I guess we can do it in another follow up PR.

That would make sense, I'll try to address it.

PR #933, opened 2026-04-23
fix: guard against null Prompt.arguments() in completion handlers

### LINKED ISSUE #932: NullPointerException in completion handler when Prompt has null arguments
Bug description

The completion validation code in `McpAsyncServer` and `McpStatelessAsyncServer` throws a `NullPointerException` when a `Prompt` is registered with `null` arguments and a `completion/complete` request is received for that prompt.

This was introduced in commit `4c85963` (#972, the 2.0 forward-compat refactor), which changed `Prompt` constructors to stop coercing `null` arguments to an empty list — but did not update the SDK's own completion handler code that calls `.arguments().stream()` without a null guard.

Additionally, `MIGRATION-2.0.md` references a `Prompt.withDefaults()` factory method as the migration path, but this method was never implemented.

### Environment

- MCP Java SDK: `main` branch, commit `4c85963`
- Java: 17+
- Transport: stdio (also affects servlet-based transports)

### Steps to reproduce

1. Create a server with a prompt that has null arguments:
```java
Prompt prompt = new Prompt("summarize", "Summarize a document", (List<PromptArgument>) null);
```
2. Register a completion handler for that prompt
3. Send a `completion/complete` request:
```json
{"jsonrpc":"2.0","id":2,"method":"completion/complete","params":{"ref":{"type":"ref/prompt","name":"summarize"},"argument":{"name":"query","value":"sum"}}}
```

### Expected behavior

An empty completion result or a proper error response.

### Actual behavior

```json
{"jsonrpc":"2.0","id":2,"error":{"code":-32603,"message":"Cannot invoke \"java.util.List.stream()\" because the return value of \"io.modelcontextprotocol.spec.McpSchema$Prompt.arguments()\" is null"}}
```

### Minimal Complete Reproducible example
https://github.com/sainathreddyb/mcp-test-servers/tree/main/bug-repro-server

### Root cause

`McpAsyncServer.java` ~line 1002 and `McpStatelessAsyncServer.java` ~line 747:
```java
if (!promptSpec.prompt()
    .arguments()   // returns null since 2.0
    .stream()      // NPE
    ...
```

### Fix
- Add null guard before `.stream()` in both server classes
- Implement the `Prompt.withDefaults()` factory referenced in `MIGRATION-2.0.md`
---

sainathreddyb:
---

Fixes #932

The 2.0 forward-compat refactor (#972) changed `Prompt` constructors to stop coercing `null` arguments to an empty list, but the completion validation code in `McpAsyncServer` and `McpStatelessAsyncServer` still called `.arguments().stream()` without a null guard, causing a `NullPointerException` when a `Prompt` with null arguments receives a `completion/complete` request.

Additionally, `MIGRATION-2.0.md` references `Prompt.withDefaults()` as the migration path, but this factory method was never implemented.

This commit:
- Adds a null check before `.arguments().stream()` in both `McpAsyncServer` and `McpStatelessAsyncServer`
- Adds `Prompt.withDefaults(name, description, arguments)` factory that coerces `null` to an empty list, matching the 1.x behaviour

## Motivation and Context

After #972 merged, any MCP server that registers a `Prompt` with `null` arguments and a completion handler will crash with a `NullPointerException` when a `completion/complete` request arrives. This is the SDK's own internal code failing to handle the null-semantics change it introduced — not a user error.

## How Has This Been Tested?

- Reproduced the bug with a standalone stdio MCP server that registers a prompt with null arguments and sends a `completion/complete` request — confirmed the NPE
- Applied the fix and verified the server returns an empty completion result instead of crashing
- All existing tests pass (`./mvnw clean test`)

## Breaking Changes

None. This is a bugfix that makes the SDK more tolerant.

## Types of changes

- [ ] Documentation update

## Checklist

- [x] I have read the [[MCP Documentation](https://modelcontextprotocol.io/)](https://modelcontextprotocol.io)

## Additional context

The `Prompt.withDefaults()` factory was documented in `MIGRATION-2.0.md` as the recommended migration path but never implemented. This PR adds it so the migration guide is accurate.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
@@ -999,9 +999,8 @@ public class McpAsyncServer {
 						.message("Prompt not found: " + promptReference.name())
 						.build());
 				}
-				if (!promptSpec.prompt()
-					.arguments()
-					.stream()
+				List<McpSchema.PromptArgument> promptArgs = promptSpec.prompt().arguments();
+				if (promptArgs != null && !promptArgs.stream()
 					.filter(arg -> arg.name().equals(argumentName))
 					.findFirst()
 					.isPresent()) {
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
@@ -744,9 +744,8 @@ public class McpStatelessAsyncServer {
 						.message("Prompt not found: " + promptReference.name())
 						.build());
 				}
-				if (!promptSpec.prompt()
-					.arguments()
-					.stream()
+				List<McpSchema.PromptArgument> promptArgs = promptSpec.prompt().arguments();
+				if (promptArgs != null && !promptArgs.stream()
 					.filter(arg -> arg.name().equals(argumentName))
 					.findFirst()
 					.isPresent()) {
--- a/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -10,6 +10,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.slf4j.Logger;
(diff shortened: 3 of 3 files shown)


chemicL:
Thanks for the proposal to fix this, however the fix here would result in actually executing the handler while the previous behavior was to return an empty response without calling the handler. I addressed this together with a test in #934

Issue #304, opened 2025-06-11
STDIO Message sending is not thread-safe


damien-urruty-sonarsource:
For a few days we are noticing flaky tests in our CI. After some investigation, I noticed that this call to `tryEmitNext` fails:

https://github.com/modelcontextprotocol/java-sdk/blob/2f944349cc77009d020ebddc8b9967b8e1b14baf/mcp/src/main/java/io/modelcontextprotocol/server/transport/StdioServerTransportProvider.java#L153

The EmitResult object that is returned is [FAIL_NON_SERIALIZED](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Sinks.EmitResult.html#FAIL_NON_SERIALIZED).

I am not experienced with Reactor at all, but I think the reason is that this code block is not "thread-safe", as sinks cannot try to emit more than one message at a time from different threads. 

We get this error from time to time in our MCP server tests. We rely on the deprecated `McpSyncServer#loggingNotification notification`, which can be called from a background thread in our case, and this sometimes causes the message sending of tool results to fail. 

I am not sure if it's only a problem when tools and logging happen at the same time, or if other concurrent messages can be sent back to the client from `StdioServerTransport` in other cases.

I don't know how to fix, I just added a synchronized block in this lambda, which is probably a terrible fix, but at least our tests are more stable now.

Let me know if you need more details

chemicL:
Thank you for the report. You are correct, the implementation has to handle `FAIL_NON_SERIALIZED` (most probably using `reactor.core.publisher.Sinks.EmitFailureHandler#busyLooping`).

chemicL:
Resolved via #935

PR #935, opened 2026-04-25
fix: make STDIO transport sendMessage thread-safe (#304)

nquinquenel:
Hello, we've had this [issue #304](https://github.com/modelcontextprotocol/java-sdk/issues/304) for a while in our MCP server, and I implemented this fix following your recommendation.

As a summary, concurrent calls to `sendMessage` on the STDIO transports raced on the unicast `Sinks.Many` and sporadically failed with `FAIL_NON_SERIALIZED`, dropping the message with `RuntimeException("Failed to enqueue message")`. This happens in practice whenever tools and `loggingNotification` are emitted from different threads on the server, and can happen on the client under any concurrent sender.

I switched `tryEmitNext` to `emitNext(message, Sinks.EmitFailureHandler.busyLooping(Duration.ofMillis(100)))` as suggested by the maintainers in the issue, so concurrent emissions are serialized. Applied on both sides:

- `mcp-core/src/main/java/io/modelcontextprotocol/server/transport/StdioServerTransportProvider.java`
- `mcp-core/src/main/java/io/modelcontextprotocol/client/transport/StdioClientTransport.java`

The remaining `catch (Sinks.EmissionException)` path still returns a failed `Mono` for the non-retryable modes (`FAIL_CANCELLED`, `FAIL_TERMINATED`, busyLooping timeout), preserving the existing error semantics.

I also removed the TODO, as I believe that the only realistic failure (`FAIL_NON_SERIALIZED`) is now handled in-place by busyLooping. The other scenario such as `FAIL_OVERFLOW` is unreachable with the current sink - `onBackpressureBuffer()` is unbounded so there's no overflow to reschedule. Happy to revert if you'd rather keep the hint for a future bounded-buffer refactor.

## Test plan

- [x] Added `shouldHandleConcurrentSendMessage` regression test that fans 500 `sendMessage` calls across 16 parallel rails and verifies all complete and every message lands on the output stream. The test fails on `main` (`RuntimeException: Failed to enqueue message` after ~44 emissions) and passes with the fix.
- [x] `./mvnw -pl mcp-test -am test -Dtest='StdioServerTransportProviderTests,StdioMcpSyncClientTests'` - 52/52 pass.
- [x] `StdioMcpAsyncClientTests` pass in isolation with the fix.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/StdioClientTransport.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/StdioClientTransport.java
@@ -227,16 +227,13 @@ public class StdioClientTransport implements McpClientTransport {
 
 	@Override
 	public Mono<Void> sendMessage(JSONRPCMessage message) {
-		if (this.outboundSink.tryEmitNext(message).isSuccess()) {
-			// TODO: essentially we could reschedule ourselves in some time and make
-			// another attempt with the already read data but pause reading until
-			// success
-			// In this approach we delegate the retry and the backpressure onto the
-			// caller. This might be enough for most cases.
+		try {
+			// busyLooping retries FAIL_NON_SERIALIZED under concurrent senders
+			this.outboundSink.emitNext(message, Sinks.EmitFailureHandler.busyLooping(Duration.ofMillis(100)));
 			return Mono.empty();
 		}
-		else {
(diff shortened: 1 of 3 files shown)


chemicL:
Thank you for the contribution! I improved the test to avoid potential flakiness and avoid sleeping (and finish in a coulple ms instead of always 500+) to make it deterministic. The outcome is merged in commit 2bb1481e.

PR #943, opened 2026-04-30
Validate message endpoint in SSE client transport

Kehrlann:
Add an API for validating the message endpoint that comes back as an SSE event in the SSE client transport.

### DIFF
--- a/mcp-test/src/test/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransportTests.java
+++ b/mcp-test/src/test/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransportTests.java
@@ -11,7 +11,6 @@ import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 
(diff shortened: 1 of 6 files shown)


mcp-test/src/test/java/io/modelcontextprotocol/client/transport/HttpClientSseClientTransportTests.java
```diff
@@ -417,4 +418,40 @@ void testAsyncRequestCustomizer() {
...
+				}
+				return false;
+			});
+
+			// Since connection failed, there is no message endpoint, and no message can
+			// be sent
+			JSONRPCRequest testMessage = new JSONRPCRequest(McpSchema.JSONRPC_VERSION, "test-method", "test-id",
+					Map.of("key", "value"));
+
+			StepVerifier.create(transport.sendMessage(testMessage)).verifyTimeout(Duration.ofMillis(500));
```
chemicL: Instead of timing out, can the messageEndpoint Sink get an error so that we immediately fail?

chemicL (commented):
Thanks, overall looks good, just one test related suggestion to avoid waiting. Perhaps this can have more useful consequences actually impacting other areas where we timeout instead of failing upon misconfiguration.

Kehrlann:
Implemented change, merged without re-requesting approval.

Issue #701, opened 2025-11-26
Utils.resolveUri incorrectly resolves nested paths breaking JDK HttpClient transports


chemicL:
The following reproducer:

```java
    public static void main(String[] args) {
        System.out.println("URI: " + resolveUri(URI.create("https://example.com/some"), "/path").toString());
    }
```

prints

```
https://example.com/path
```

while it should be

```
https://example.com/some/path
```

This affects usages of `HttpClientSseClientTransport` and `HttpClientStreamableHttpTransport` with base

It's not used for WebClient-based transports which are free from this defect.

java-sdk version: 0.16.0

chemicL:
I think this behaviour was introduced in #150 and noted in a follow-up comment: https://github.com/modelcontextprotocol/java-sdk/pull/150#issuecomment-2873867910

The tests added in the PR assume this behaviour is correct. I am not certain whether it's expected to behave this way. [SO users also have issues with this behaviour](https://stackoverflow.com/questions/7498030/append-relative-url-to-java-net-url).

If the baseUri from the reproducer has a trailing slash (`https://example.com/some/`) and the endpoint doesn't have the leading slash (`path`), the resolution looks correct.

This is a delicate subject so I won't be providing a fix at this time. Hoping for some pointers from someone more knowledgeable in URI specs.

chemicL:
The behaviour with the URI class probably is correct according to [the spec](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.3), but our expectations probably are wrong. Maybe we should change the paradigm a bit instead of assuming that the endpoint is appended to the base path or express some expectations (e.g. baseUri has to end with trailing slash and the mcp endpoint can't be absolute). I don't know. The goal probably was not to build something that external libraries can provide :)

chemicL:
This is actually a duplicate of #257.

PR #974, opened 2026-05-19
Fix nested path URI resolution

### LINKED ISSUE #257: io.modelcontextprotocol.util.Utils#resolveUri incorrectly handles nested paths

**Bug description**

It took me four hours to locate and fix this Bug，
When I create a simple link, it always reminds me of http status 302. Finally, it was located that the SSE MCP Server of the business side was redirected.

**Environment**
Please provide as many details as possible: Spring MCP version, Java version, which vector store you use if any, etc

Java SDK

**Steps to reproduce**
It's very simple. When you have an SSE MCP SERVER but the interface does not end with /sse, an error will definitely be reported when creating the connection.

**Expected behavior**

Adapt to the SSE MCP SERVER with non-standard users

I first fork and change the IO. Modelcontextprotocol. Util. Utils# resolveUri logic is as follows

old code
```java
	public static URI resolveUri(URI baseUrl, String endpointUrl) {
		URI endpointUri = URI.create(endpointUrl);
		if (endpointUri.isAbsolute() && !isUnderBaseUri(baseUrl, endpointUri)) {
			throw new IllegalArgumentException("Absolute endpoint URL does not match the base URL.");
		}
		else {
			return baseUrl.resolve(endpointUri);
		}
	}
```

new code
```java
	public static URI resolveUri(URI baseUrl, String endpointUrl) {
		if (endpointUrl.equals(DEFAULT_SSE_ENDPOINT)) {
			return baseUrl;
		}
		else {
			URI endpointUri = URI.create(endpointUrl);
			return baseUrl.resolve(endpointUri);
		}
	}
```

You only need to replace the following URL with a URL that does not end with SSE, and it can be reproduced

```java
import io.modelcontextprotocol.client.McpClient;
import io.modelcontextprotocol.client.McpSyncClient;
import io.modelcontextprotocol.client.transport.HttpClientSseClientTransport;
import io.modelcontextprotocol.spec.McpSchema;
import org.junit.Test;

import java.time.Duration;

public class SSEClientTest {
	private String youUrl = "https://host:port/aaa/bbb/cccc";

	@Test
	public void testSSE() {
		mockSSEClient(youUrl);
	}

	/**
	 * Test MCP Client
	 */
	private void mockSSEClient(String url) {
		McpSyncClient client = McpClient.sync(create(url))
			.requestTimeout(Duration.ofSeconds(3))
			.initializationTimeout(Duration.ofSeconds(1))
			.capabilities(McpSchema.ClientCapabilities.builder().build())
			.build();

		client.initialize();
		for (McpSchema.Tool tool : client.listTools().tools()) {
			System.out.println(" Tool :" + tool);
		}

	}

	private HttpClientSseClientTransport create(String url) {
		return HttpClientSseClientTransport.builder(url).build();
	}
}
```

Qshuai0213:
## Summary

Fixes #257

`Utils.resolveUri` incorrectly handled nested base paths — relative endpoints with a leading slash (e.g. `/api/v1`) would replace the base path entirely instead of being appended to it. This affected `HttpClientSseClientTransport` and `HttpClientStreamableHttpTransport` when the base URL had a non-root path (e.g. `https://example.com/some`).

### Root cause

`URI.resolve()` follows RFC 3986 §5.2.3: when the relative reference starts with `/`, it replaces the entire path of the base URI. So `URI.create("https://example.com/some").resolve("/path")` yields `https://example.com/path`, dropping `/some`.

### Fix approach

- Ensure the base URL always has a trailing slash (e.g. `/root` → `/root/`)
- Strip leading slashes from the endpoint (e.g. `/api/v1` → `api/v1`)
- `URI.resolve()` then performs path concatenation (`/root/` + `api/v1` = `/root/api/v1`), consistent with RFC 3986
- Absolute endpoint validation behavior is preserved unchanged

## Testing

- `./mvnw.cmd -pl mcp-core -Dtest=UtilsTests test`
- `./mvnw.cmd -pl mcp-core test`
- Added parameterized test cases covering: nested base paths, leading/trailing slashes, and query strings

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/util/Utils.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/util/Utils.java
@@ -5,6 +5,7 @@
 package io.modelcontextprotocol.util;
 
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Collection;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


Qshuai0213:
@chemicL Hi, since you've looked into this issue in both #257 and #701, could you take a look at this PR when you have time? Happy to adjust the approach if needed. Thanks!

PR #977, opened 2026-05-21
Release to maven central: do not run the tests twice

Kehrlann:
When releasing to maven central, `deploy` runs the tests before deploying.
No need to run the tests twice.


tzolov (approved):
LGTM

PR #981, opened 2026-05-24
fix: replace Mono.zip with Mono.when in StdioServerTransportProvider#sendMessage

### LINKED ISSUE #303: StdioServerTransportProvider uses incorrect Mono.zip operator
Hello,

I was debugging another issue and looking at the code, and I noticed this line:  [#[[line](https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp/src/main/java/io/modelcontextprotocol/server/transport/StdioServerTransportProvider.java#L152)]](https://github.com/modelcontextprotocol/java-sdk/blob/2f944349cc77009d020ebddc8b9967b8e1b14baf/mcp/src/main/java/io/modelcontextprotocol/server/transport/StdioServerTransportProvider.java#L152)

When opened in IntelliJ, I noticed this warning being reported: 

Image

I am not experienced at all with Mono so I don't know if it's a real problem or not, sharing just in case...

neerajbhatt:
## Summary
- `Mono.zip()` in `sendMessage()` never completes because `inboundReady` and `outboundReady` are `Sinks.One<Void>` that emit `null` via `tryEmitValue(null)` — `zip` requires actual non-null values to produce a `Tuple2`
- Replaced with `Mono.when()` which correctly waits for completion signals without requiring values

## Test plan
- [x] All 9 existing `StdioServerTransportProviderTests` pass
- [ ] CI pipeline passes

Closes #303

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/StdioServerTransportProvider.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/StdioServerTransportProvider.java
@@ -160,7 +160,7 @@ public class StdioServerTransportProvider implements McpServerTransportProvider 
 		@Override
 		public Mono<Void> sendMessage(McpSchema.JSONRPCMessage message) {
 
-			return Mono.zip(inboundReady.asMono(), outboundReady.asMono()).then(Mono.defer(() -> {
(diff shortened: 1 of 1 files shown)


neerajbhatt:
Closing as duplicate of #846 which already addresses this issue.

PR #983, opened 2026-05-24
feat: add duplicateStructuredContent option to server builders

### LINKED ISSUE #688: Allow disabling `structuredContent` duplication in McpAsyncServer
**Expected Behavior**

Make `io.modelcontextprotocol.server.McpAsyncServer.StructuredOutputCallToolHandler` configurable to allow users to decide whether they want to duplicate the `structuredContent` in the `content` property.

This configuration could be introduced via a bean or another appropriate mechanism.

A possible solution implementation might look like this:

```java
@Override
public Mono<CallToolResult> apply(McpAsyncServerExchange exchange, McpSchema.CallToolRequest request) {

    return this.delegateCallToolResult.apply(exchange, request).map(result -> {

        // Skipping some code for brevity

        if (Utils.isEmpty(result.content()) && keepCompatibility()) {  // <-- HERE keepCompatibility() METHOD IS NEW
            // For backwards compatibility, a tool that returns structured
            // content SHOULD also return functionally equivalent unstructured
            // content. (For example, serialized JSON can be returned in a
            // TextContent block.)
            // https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content

            return CallToolResult.builder()
                .content(List.of(new McpSchema.TextContent(validation.jsonStructuredOutput())))
                .isError(result.isError())
                .structuredContent(result.structuredContent())
                .build();
            }

            return result;
    });
}

private boolean keepCompatibility() {
    // Provide a way for users to specify whether to keep compatibility or not
}
```

**Current Behavior**

Currently, `StructuredOutputCallToolHandler` always duplicates `structuredContent` into the `content` property. This behavior is not mandated by the MCP specification.

**Context**

MCP tool users may be dissatisfied receiving duplicate content because:
- It causes unnecessary network traffic.
- It reduces performance.
- **Most importantly, it consumes unnecessary space in the LLM context window on the client side.**

There is no current way to prevent `StructuredOutputCallToolHandler#apply()` from copying `structuredContent` into `content` because the `StructuredOutputCallToolHandler` class is private and cannot be extended or customized externally.

neerajbhatt:
## Summary

- Add `duplicateStructuredContent(boolean)` builder method to all server specifications (async, sync, stateless async, stateless sync)
- When `true` (default), structured content is auto-duplicated into a `TextContent` block in the `content` field for backwards compatibility — preserving existing behavior
- When `false`, the duplication is skipped, reducing response payload size for clients that fully support `structuredContent`
- Schema validation of structured output is always performed regardless of this setting

## Usage

```java
McpServer.async(transportProvider)
    .duplicateStructuredContent(false) // skip backwards-compat duplication
    .tool(myTool, myHandler)
    .build();
```

## Changes

- **`McpServer.java`** — Added `duplicateStructuredContent` field and builder method to `AsyncSpecification`, `SyncSpecification`, `StatelessAsyncSpecification`, and `StatelessSyncSpecification`; passed through to server constructors
- **`McpAsyncServer.java`** — Added field, updated constructors and `StructuredOutputCallToolHandler` to conditionally duplicate
- **`McpStatelessAsyncServer.java`** — Same changes as `McpAsyncServer`

## Test plan

- [x] All 315 mcp-core unit tests pass
- [x] `HttpServletStatelessIntegrationTests` structured output tests pass (1 pre-existing flaky `testToolCallSuccess` timeout)
- [ ] CI pipeline passes

Closes #688

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
@@ -101,6 +101,8 @@ public class McpAsyncServer {
 
 	private final boolean validateToolInputs;
 
+	private final boolean duplicateStructuredContent;
+
 	private final McpSchema.ServerCapabilities serverCapabilities;
 
 	private final McpSchema.Implementation serverInfo;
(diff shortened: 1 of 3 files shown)


neerajbhatt:
@Kehrlann Would you mind reviewing this PR when you get a chance? It adds a configurable flag to control structured content duplication in tool responses (issue #688). Thanks!

PR #939, opened 2026-04-27
feat(core): add URL elicitation support (SEP-1036)

sainathreddyb:
## Summary

Add URL-type elicitation schema support allowing servers to request URL input from users during tool execution (SEP-1036).

**Changes:**
- `UrlElicitationSchema` and `UrlElicitationResult` types in `McpSchema.java`
- Client-side URL elicitation handler registration and dispatch in `McpAsyncClient`, `McpClient`, and `McpClientFeatures`
- Server exchange methods for URL elicitation requests in `McpAsyncServerExchange` and `McpSyncServerExchange`
- Comprehensive tests for schema serialization, client handler dispatch, and server exchange round-trips

## Motivation and Context

SEP-1036 introduces a URL elicitation type, enabling MCP servers to request a URL from the user during tool execution — for example, asking the user to provide a callback URL, a webhook endpoint, or a resource link. This complements the existing form-based elicitation with a purpose-built schema for URL input.

## How Has This Been Tested?

- `McpSchemaTests`: Serialization/deserialization round-trip tests for `UrlElicitationSchema` and `UrlElicitationResult`
- `McpAsyncClientResponseHandlerTests`: Client-side handler registration and dispatch for URL elicitation requests
- `McpAsyncServerExchangeTests` / `McpSyncServerExchangeTests`: Server exchange URL elicitation request/response flow
- Full build passes: `./mvnw clean test` — all modules green (BUILD SUCCESS)
- Tested by creating server that supports [url elicitation](https://github.com/sainathreddyb/mcp-test-servers/tree/main/elicitation-server)

## Breaking Changes

None. This is a purely additive change — new types and methods alongside existing elicitation support.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [[MCP Documentation](https://modelcontextprotocol.io/)](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

## Additional context

This implements the Java SDK side of SEP-1036. The schema types follow the same pattern as the existing `ElicitationSchema`/`ElicitationResult` for form-based elicitation, keeping the API consistent.

PR #993, opened 2026-06-02
Add URL elicitation support (SEP-1036)

### LINKED ISSUE #692: SEP-1036: URL Mode Elicitation for secure out-of-band interactions
This is a tracking issue for https://github.com/modelcontextprotocol/modelcontextprotocol/pull/887

Kehrlann:
Add URL-type elicitation schema support allowing servers to request URL input from users during tool execution. This enables out-of-band interactions like payment processing or API key entry.

Breaking changes:

- `ElicitRequest` changed from a `record` to an `interface`.
- The original `ElicitRequest` record was renamed to `ElicitFormRequest`.
- `McpClient` builder `elicitation()` methods now accept `ElicitFormRequest` instead of `ElicitRequest`.

New APIs:

- `ElicitUrlRequest` record for URL-mode elicitation.
- `urlElicitation()` builder methods in `McpClient`.
- `elicitationCompleteConsumer()` and `elicitationCompleteConsumers()` builder methods in `McpClient`.
- `sendElicitationComplete()` methods in `McpAsyncServer` and `McpSyncServer`.
- `McpError.URL_ELICITATION_REQUIRED` and `McpSchema.ErrorCodes.URL_ELICITATION_REQUIRED`.
- `ElicitationCompleteNotification` record and `METHOD_NOTIFICATION_ELICITATION_COMPLETE` constant.

Supersedes #939 
Closes #692

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -715,9 +715,7 @@ var tool = SyncToolSpecification.builder()
         }
 
         // Request user confirmation
-        ElicitRequest elicitRequest = ElicitRequest.builder()
-            .message("Do you want to proceed with this action?")
-            .requestedSchema(Map.of(
+        ElicitRequest elicitRequest = ElicitFormRequest.builder("Do you want to proceed with this action?", Map.of(
                 "type", "object",
                 "properties", Map.of("confirmed", Map.of("type", "boolean"))
             ))
@@ -733,6 +731,34 @@ var tool = SyncToolSpecification.builder()
         } else {
             return CallToolResult.builder()
                 .content(List.of(new McpSchema.TextContent("Action declined")))
+                .build();
+        }
+    })
+    .build();
+```
+
+To request out-of-band URL elicitation, such as a user authorizing an OAuth flow:
+
+```java
+var urlTool = SyncToolSpecification.builder()
+    .tool(Tool.builder()
+        .name("oauth-auth")
+        .description("Authenticates via OAuth")
+        .inputSchema(schema)
+        .build())
+    .callHandler((exchange, request) -> {
+        // Request URL elicitation from client
+        ElicitRequest urlRequest = McpSchema.ElicitUrlRequest.builder("Please authenticate", "https://example.com/oauth", "oauth-123").build();
+
+        ElicitResult result = exchange.elicit(urlRequest);
+
+        if (result.action() == ElicitResult.Action.ACCEPT) {
+            return CallToolResult.builder()
+                .content(List.of(new McpSchema.TextContent("Authentication successful")))
+                .build();
+        } else {
+            return CallToolResult.builder()
+                .content(List.of(new McpSchema.TextContent("Authentication failed or cancelled")))
                 .build();
         }
     })
--- a/docs/server.md
+++ b/docs/server.md
@@ -748,19 +748,25 @@ var urlTool = SyncToolSpecification.builder()
         .build())
     .callHandler((exchange, request) -> {
         // Request URL elicitation from client
-        ElicitRequest urlRequest = McpSchema.ElicitUrlRequest.builder("Please authenticate", "https://example.com/oauth", "oauth-123").build();
+        if (
+                exchange.getClientCapabilities().elicitation() != null 
+                && exchange.getClientCapabilities().elicitation().url() != null
+        ) {
+            ElicitRequest urlRequest = McpSchema.ElicitUrlRequest
+                    .builder("Please authenticate", "https://example.com/oauth", "oauth-123").build();
+            ElicitResult result = exchange.elicit(urlRequest);
+            // handle result.action == CANCELLED or DENIED
+            if (result.action() != ElicitResult.Action.ACCEPT) {
+                return CallToolResult.builder()
+                        .content(List.of(new McpSchema.TextContent("Authentication failed or cancelled")))
+                        .build();
+            }
+        }
 
-        ElicitResult result = exchange.elicit(urlRequest);
-
-        if (result.action() == ElicitResult.Action.ACCEPT) {
-            return CallToolResult.builder()
+        // wait for user to visit the URL
+        return CallToolResult.builder()
                 .content(List.of(new McpSchema.TextContent("Authentication successful")))
                 .build();
-        } else {
-            return CallToolResult.builder()
-                .content(List.of(new McpSchema.TextContent("Authentication failed or cancelled")))
-                .build();
-        }
     })
     .build();
 ```
--- a/mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
@@ -15,6 +15,8 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
+import io.modelcontextprotocol.spec.McpSchema.ElicitFormRequest;
+import io.modelcontextprotocol.spec.McpSchema.ElicitUrlRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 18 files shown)


Kehrlann:
@sainathreddyb I reworked your PR, would you have a look?

docs/server.md
```diff
@@ -739,6 +737,34 @@ var tool = SyncToolSpecification.builder()
...
+    .tool(Tool.builder()
+        .name("oauth-auth")
+        .description("Authenticates via OAuth")
+        .inputSchema(schema)
+        .build())
+    .callHandler((exchange, request) -> {
+        // Request URL elicitation from client
+        ElicitRequest urlRequest = McpSchema.ElicitUrlRequest.builder("Please authenticate", "https://example.com/oauth", "oauth-123").build();
+
+        ElicitResult result = exchange.elicit(urlRequest);
```
chemicL: We probably should add a check that elicitation is supported by the client and show an example for handling non supported case to return an error.

mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -226,11 +238,18 @@ public class McpAsyncClient {
 
 		// Elicitation Handler
 		if (this.clientCapabilities.elicitation() != null) {
-			if (features.elicitationHandler() == null) {
+			if ((this.clientCapabilities.elicitation().url() == null
+					|| this.clientCapabilities.elicitation().form() != null)
```
chemicL: This check is interesting. Can you add a comment (if my understanding is correct) that this is a backwards-compatible check for lack of a nested specifier to mean form elicitation?

mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -552,23 +581,48 @@ private RequestHandler<CreateMessageResult> samplingCreateMessageHandler() {
...
+							&& result.content() != null) {
+						Map<String, Object> merged = new HashMap<>(result.content());
+						applyElicitationDefaults(formRequest.requestedSchema(), merged);
+						return new ElicitResult(result.action(), merged, result.meta());
+					}
+					return result;
+				});
+			}
+
+			return Mono.error(new IllegalStateException("Unknown elictation type deserialized"));
```
chemicL, line 614: This is not an illegal state, rather the input from the server is wrong, we should fail it gracefully.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpError.java
```diff
@@ -20,6 +20,15 @@ public class McpError extends RuntimeException {
...
 
+	/**
+	 * <a href=
+	 * "https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation#url-elicitation-required-error">URL
+	 * Elicitation Required</a>
+	 */
+	public static final Function<java.util.List<McpSchema.ElicitUrlRequest>, McpError> URL_ELICITATION_REQUIRED = elicitations -> new McpError(
+			new JSONRPCError(McpSchema.ErrorCodes.URL_ELICITATION_REQUIRED, "URL elicitation required",
+					Map.of("elicitations", elicitations)));
+
```
chemicL, line 31: Would be cool to see an integration test with this.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3996,10 +4047,135 @@ public Builder progressToken(Object progressToken) {
...
+			public Builder url(String url) {
+				Assert.notNull(url, "url must not be null");
+				this.url = url;
+				return this;
+			}
+
+			public Builder elicitationId(String elicitationId) {
+				Assert.notNull(elicitationId, "elicitationId must not be null");
+				this.elicitationId = elicitationId;
+				return this;
```
chemicL: Since these are required at creation time, do we need to allow modification?

chemicL: I see that in the form variant we also can modify, but this is just a genuine question if we should consider limiting the possibilities to modify the builder.

chemicL (changes requested):
Overall this looks great! Just a few minor requests. Also, please add the breaking changes information into the migration md file.

mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
Kehrlann: I think this _is_ an illegal state exception.

It only happens when:
1. Deserialization succeeded (otherwise `transport.unmarshalFrom(...)` throws)
2. The result of the deserialization is an `ElicitRequest` but neither `ElicitFormRequest` (the default) nor `ElicitUrlRequest`This is not possible because of our JSON annotations, it will always be either of those, defaulting to `ElicitFormRequest` when the input is "wrong" (e.g. missing fields) but parseable.

There _may_ be ways to trick the deserializer into a producing a new subtype by injecting purposefully designed mixins (though I'm not 100% sure its possible). But this is developers crafting code that _cannot_ work, so I'd stick with illegal state.

Thoughts?

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpError.java
Kehrlann: There's `io.modelcontextprotocol.AbstractMcpClientServerIntegrationTests#testElicitationRequiredError` . Is that enough?

docs/server.md
Kehrlann: I had intentionally left it out for the brevity of the example, but this makes sense.

It have reworked the example with a few comments.

mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
chemicL: Makes sense. IllegalState it is!

docs/server.md
```diff
@@ -748,19 +748,25 @@ var urlTool = SyncToolSpecification.builder()
         .build())
     .callHandler((exchange, request) -> {
         // Request URL elicitation from client
-        ElicitRequest urlRequest = McpSchema.ElicitUrlRequest.builder("Please authenticate", "https://example.com/oauth", "oauth-123").build();
-
-        ElicitResult result = exchange.elicit(urlRequest);
+        if (
+                exchange.getClientCapabilities().elicitation() != null 
+                && exchange.getClientCapabilities().elicitation().url() != null
```
chemicL, line 753: The UX is not great but we can add something to simplify the check in the future.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -4022,12 +4022,22 @@ private Builder(String message, Map<String, Object> requestedSchema) {
 				this.requestedSchema = requestedSchema;
 			}
 
+			/**
+			 * @deprecated Use {@link ElicitRequest#builder(String, Map)} factory method
+			 * instead.
+			 */
+			@Deprecated
```
chemicL, line 4017: No need to deprecate, these are new types.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -4022,12 +4022,22 @@ private Builder(String message, Map<String, Object> requestedSchema) {
...
 				Assert.notNull(message, "message must not be null");
 				this.message = message;
 				return this;
 			}
 
+			/**
+			 * @deprecated Use {@link ElicitRequest#builder(String, Map)} factory method
+			 * instead.
+			 */
+			@Deprecated
```
chemicL, line 4028: No need to deprecate.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpError.java
chemicL: I completely overlooked that test case. Looks great!

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
Kehrlann: Correct. But this builder is here just to keep ElicitRequest.builder() working with the old API. They should not be used. I want to signal "don't use these methods, you should be using ElicitFormRequest.builder(message, schema) and not change the values after."

chemicL (approved):

PR #995, opened 2026-06-03
fix: avoid dropped errors when transport is closed or uninitialized

chemicL:
Also avoid logging peer disconnect at warn level.


mcp-test/src/test/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransportTest.java
```diff
@@ -139,13 +142,14 @@ void testCloseUninitialized() {
...
 		StepVerifier.create(transport.sendMessage(testMessage))
-			.expectErrorMessage("MCP session has been closed")
+			.expectErrorMessage("Transport has already been closed.")
 			.verify();
 	}
 
 	@Test
 	void testCloseInitialized() {
 		var transport = HttpClientStreamableHttpTransport.builder(host).build();
+		// transport.connect(Function.identity()).block();
```
Kehrlann: How come we don't need to call "connect" anymore?
If we are not "connect"-ed, shouldn't `sendMessage` error rather than complete?

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
```diff
@@ -30,17 +30,7 @@
...
-import io.modelcontextprotocol.spec.DefaultMcpTransportStream;
-import io.modelcontextprotocol.spec.HttpHeaders;
-import io.modelcontextprotocol.spec.McpClientTransport;
-import io.modelcontextprotocol.spec.McpSchema;
-import io.modelcontextprotocol.spec.McpTransportException;
-import io.modelcontextprotocol.spec.McpTransportSession;
-import io.modelcontextprotocol.spec.McpTransportSessionNotFoundException;
-import io.modelcontextprotocol.spec.McpTransportStream;
-import io.modelcontextprotocol.spec.ProtocolVersions;
+import io.modelcontextprotocol.spec.*;
```
Kehrlann: nit - we typically don't do star imports.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
```diff
@@ -389,18 +385,18 @@ else if (statusCode == BAD_REQUEST) {
 								"Received unrecognized SSE event type: " + sseResponseEvent.sseEvent().event()));
 					})
 					.retryWhen(authorizationErrorRetrySpec())
-					.flatMap(jsonrpcMessage -> this.handler.get().apply(Mono.just(jsonrpcMessage)))
+					.flatMap(jsonrpcMessage -> requestHandler.apply(Mono.just(jsonrpcMessage)))
```
Kehrlann, line 399: can this leak?

mcp-test/src/test/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransportTest.java
chemicL: Yeah, notice that this commented line is added. Previously there was no connect at all and it generated drops. Now you get a warning.

The thing is that for streamable http there is no "connect", it just sets up a json-rpc handler.
If you call sendMessage it is an asynchronous operation. We don't have any feedback on the sending whether the server responded. The historical reason is the separate channel on STDIO and on the legacy SSE transport. With Streamable HTTP it should actually be possible to change this paradigm but the intention at the time was not to break the common interfaces.

Since the test case used to work without the connect call, I didn't want to suddenly break it.

Apparently this test passes despite the transport ignoring all the responses.

WDYT should be the right behaviour? I can either remove the commented code (it slipped, but maybe it's good that it stayed here so we can have the discussion) or uncomment it and have the sendMessage fail if `connect` was not setup.

Kehrlann: Ah thanks, I get it now.

Let's not break the API. The fact that `sendMessage` returns a `Mono<Void>` makes it useless without a handler, so I don't think we need a strong signal (error) when a handler is missing.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
chemicL: Not sure, but is this structure much different from the one we had before? We now have a local variable capturing the handler, before we fetched it just-in-time. So from leaking the handler perspective - not a big deal, the handler should live throughout the lifetime of the transport. The other way - leaking messages to the handler - it was always the case.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
chemicL: Fixed!

mcp-test/src/test/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransportTest.java
chemicL: I decided to include the connect call, it makes sense to have a fully initialized instance here and in the uninitialized case skip the connect.

mcp-core/src/main/java/io/modelcontextprotocol/client/transport/HttpClientStreamableHttpTransport.java
Kehrlann: fair enough, you're not supposed to call `.connect()` many times anyway.

Kehrlann (approved):

Issue #693, opened 2025-11-26
SEP-1577 - Sampling With Tools


Kehrlann:
This is a tracking issue for https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577

chemicL:
Closing due to the fact that Sampling is becoming deprecated (https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577) in the upcoming specification and it makes no sense to introduce a feature that is bound to be removed in near future.

PR #1018, opened 2026-06-10
Add SEP-1577 sampling with tools support

chemicL:
Introduces parallel V2 sampling types (`SamplingMessageV2`, `CreateMessageWithToolsRequest`/`Result`, `ToolUseContent`, `ToolResultContent`, `ToolChoice`) alongside the existing V1 types, leaving the legacy wire format byte-identical. Servers call `createMessageWithTools` on the exchange; a version gate refuses multi-content or tools payloads when the negotiated protocol version is older than 2025-11-25. Clients opt in via `samplingWithTools(handler)` on the builder, which advertises sampling.tools in the capability handshake. `StopReason` gains `TOOL_USE`.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
@@ -103,6 +103,9 @@ public class McpAsyncClient {
 	public static final TypeRef<CreateMessageRequest> CREATE_MESSAGE_REQUEST_TYPE_REF = new TypeRef<>() {
 	};
 
+	public static final TypeRef<McpSchema.CreateMessageWithToolsRequest> CREATE_MESSAGE_WITH_TOOLS_REQUEST_TYPE_REF = new TypeRef<>() {
+	};
+
 	public static final TypeRef<LoggingMessageNotification> LOGGING_MESSAGE_NOTIFICATION_TYPE_REF = new TypeRef<>() {
 	};
 
@@ -141,6 +144,8 @@ public class McpAsyncClient {
 	 * servers in their prompts.
 	 */
 	private Function<CreateMessageRequest, Mono<CreateMessageResult>> samplingHandler;
+
+	private Function<McpSchema.CreateMessageWithToolsRequest, Mono<McpSchema.CreateMessageWithToolsResult>> samplingWithToolsHandler;
 
 	/**
 	 * MCP provides a standardized way for servers to request additional information from
@@ -227,11 +232,21 @@ public class McpAsyncClient {
 
 		// Sampling Handler
 		if (this.clientCapabilities.sampling() != null) {
-			if (features.samplingHandler() == null) {
-				throw new IllegalArgumentException(
-						"Sampling handler must not be null when client capabilities include sampling");
+			boolean withTools = this.clientCapabilities.sampling().tools() != null;
+			if (withTools) {
+				if (features.samplingWithToolsHandler() == null) {
+					throw new IllegalArgumentException(
+							"Sampling-with-tools handler must not be null when client capabilities include sampling.tools");
+				}
+				this.samplingWithToolsHandler = features.samplingWithToolsHandler();
 			}
-			this.samplingHandler = features.samplingHandler();
+			else {
+				if (features.samplingHandler() == null) {
+					throw new IllegalArgumentException(
+							"Sampling handler must not be null when client capabilities include sampling");
+				}
+				this.samplingHandler = features.samplingHandler();
+			}
 			requestHandlers.put(McpSchema.METHOD_SAMPLING_CREATE_MESSAGE, samplingCreateMessageHandler());
 		}
 
@@ -575,11 +590,29 @@ public class McpAsyncClient {
 	// --------------------------
 	// Sampling
 	// --------------------------
-	private RequestHandler<CreateMessageResult> samplingCreateMessageHandler() {
+	private RequestHandler<Object> samplingCreateMessageHandler() {
 		return params -> {
-			McpSchema.CreateMessageRequest request = transport.unmarshalFrom(params, CREATE_MESSAGE_REQUEST_TYPE_REF);
-
-			return this.samplingHandler.apply(request);
+			// Dispatch to V2 handler if the request carries tools/toolChoice, otherwise
+			// V1
+			McpSchema.CreateMessageWithToolsRequest v2Request = transport.unmarshalFrom(params,
+					CREATE_MESSAGE_WITH_TOOLS_REQUEST_TYPE_REF);
+			boolean isV2Request = (v2Request.tools() != null && !v2Request.tools().isEmpty())
+					|| v2Request.toolChoice() != null;
+			if (isV2Request) {
+				if (this.samplingWithToolsHandler == null) {
+					return Mono.error(new IllegalStateException(
+							"Received sampling request with tools, but no samplingWithTools handler is registered. "
+									+ "Use McpClient.async(transport).samplingWithTools(handler) to register one."));
+				}
+				return this.samplingWithToolsHandler.apply(v2Request).cast(Object.class);
+			}
+			// V1 path: unmarshal as the legacy type for handler type-safety
+			McpSchema.CreateMessageRequest v1Request = transport.unmarshalFrom(params, CREATE_MESSAGE_REQUEST_TYPE_REF);
+			if (this.samplingHandler == null) {
+				return Mono.error(
+						new IllegalStateException("Received sampling request, but no sampling handler is registered."));
+			}
+			return this.samplingHandler.apply(v1Request).cast(Object.class);
 		};
 	}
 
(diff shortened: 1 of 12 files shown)


mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -622,10 +631,28 @@ public RootCapabilities build() {
...
 		 * from MCP servers in their prompts.
+		 *
+		 * @param tools Present when the client supports sampling with tools (SEP-1577).
+		 * When non-null the client can handle {@code ToolUseContent} and
+		 * {@code ToolResultContent} blocks and the {@code "toolUse"} stop reason.
 		 */
 		@JsonInclude(JsonInclude.Include.NON_ABSENT)
 		@JsonIgnoreProperties(ignoreUnknown = true)
-		public record Sampling() {
+		public record Sampling(@JsonProperty("tools") SamplingTools tools) {
```
Kehrlann, line 641: SEP-1577 also defines a `context` property for sampling ([schema](https://modelcontextprotocol.io/specification/2025-11-25/schema#clientcapabilities-sampling))

is this ignored on purpose?

mcp-core/src/main/java/io/modelcontextprotocol/client/McpAsyncClient.java
```diff
@@ -575,11 +590,29 @@ private RequestHandler<McpSchema.ListRootsResult> rootsListRequestHandler() {
...
+				}
+				return this.samplingWithToolsHandler.apply(v2Request).cast(Object.class);
+			}
+			// V1 path: unmarshal as the legacy type for handler type-safety
+			McpSchema.CreateMessageRequest v1Request = transport.unmarshalFrom(params, CREATE_MESSAGE_REQUEST_TYPE_REF);
+			if (this.samplingHandler == null) {
+				return Mono.error(
+						new IllegalStateException("Received sampling request, but no sampling handler is registered."));
+			}
+			return this.samplingHandler.apply(v1Request).cast(Object.class);
```
Kehrlann, line 615: If the incoming request has no `tools` or `toolChoice` (both of which are optional), then we route to v2. Otherwise, we route to v1.

A server _could_ send a request with or without tools, but the constructor only allows us to define one of `samplingWithToolsHandler` / `samplingHandler` (gated with `if (withTools) / else`). So if the server sends a request without tools to a v2-enabled client, we'll throw `IllegalStateException` because we're missing `samplingHandler`

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
```diff
@@ -150,6 +176,82 @@ public Mono<McpSchema.CreateMessageResult> createMessage(McpSchema.CreateMessage
...
+	public Mono<McpSchema.CreateMessageWithToolsResult> createMessageWithTools(
+			McpSchema.CreateMessageWithToolsRequest request) {
+		if (this.clientCapabilities == null) {
+			return Mono
+				.error(new IllegalStateException("Client must be initialized. Call the initialize method first!"));
+		}
+		if (this.clientCapabilities.sampling() == null) {
+			return Mono.error(new IllegalStateException("Client must be configured with sampling capabilities"));
+		}
+		if (this.clientCapabilities.sampling().tools() == null) {
```
Kehrlann, line 211: If `tools` is null, users cannot use this method. But this is the only way to create multi-content sampling requests, even without tools.

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServerExchange.java
```diff
@@ -150,6 +176,82 @@ public Mono<McpSchema.CreateMessageResult> createMessage(McpSchema.CreateMessage
...
+		return this.session.sendRequest(McpSchema.METHOD_SAMPLING_CREATE_MESSAGE, request,
+				CREATE_MESSAGE_WITH_TOOLS_RESULT_TYPE_REF);
+	}
+
+	/**
+	 * Returns {@code true} when the given version string is older than
+	 * {@link McpSchema#PROTOCOL_VERSION_SAMPLING_WITH_TOOLS}, or when the version is
+	 * {@code null} (unknown — treated as pre-SEP-1577).
+	 */
+	private static boolean isOlderThanSamplingWithToolsVersion(String negotiatedVersion) {
```
Kehrlann, line 248: I had to scratch my head a bit at the name.
Consider renaming to `supportsSamplingWithTools` ?

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -109,6 +110,14 @@ private McpSchema() {
 	// Sampling Methods
 	public static final String METHOD_SAMPLING_CREATE_MESSAGE = "sampling/createMessage";
 
+	/**
+	 * Minimum negotiated protocol version that supports SEP-1577 "Sampling with Tools".
+	 * Clients advertising {@code sampling.tools} and servers calling
+	 * {@link io.modelcontextprotocol.server.McpAsyncServerExchange#createMessageWithTools}
+	 * must have negotiated at least this version.
+	 */
+	public static final String PROTOCOL_VERSION_SAMPLING_WITH_TOOLS = "2025-11-25";
```
Kehrlann, line 119: I don't think we should hint at particular feature gates within the schema. The [schema reference](https://modelcontextprotocol.io/specification/2025-11-25/schema) doesn't mention this kind of behavior , I'd prefer having it closer to the feature itself.

Consider renaming to `PROTOCOL_2025_11_25` and keeping the "sampling with tools" concept in the server instead

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3055,6 +3092,47 @@ public Tool build() {
...
+
+		/** The LLM must call at least one tool. */
+		public static final String MODE_REQUIRED = "required";
+
+		/** Tool calling is disabled for this request. */
+		public static final String MODE_NONE = "none";
+
+		public ToolChoice {
+			Assert.hasText(mode, "mode must not be empty");
+		}
```
Kehrlann, line 3117: - Validate that `mode` is one of the allowed modes
- `"toolChoice": {}` is a param for a sampling request ([schema](https://modelcontextprotocol.io/specification/2025-11-25/schema#toolchoice)), and this would fail deserializing it . From what I understand, it should default to `auto` in that case?

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3521,6 +3599,85 @@ public SamplingMessage build() {
...
+	 * On deserialization, a bare JSON object is accepted in addition to a JSON array (via
+	 * {@link JsonFormat.Feature#ACCEPT_SINGLE_VALUE_AS_ARRAY}) so that legacy peers that
+	 * still emit a single-block object are handled transparently.
+	 *
+	 * @param role the role of the message sender
+	 * @param content one or more content blocks
+	 */
+	@JsonInclude(JsonInclude.Include.NON_ABSENT)
+	@JsonIgnoreProperties(ignoreUnknown = true)
+	public record SamplingMessageV2( // @formatter:off
```
Kehrlann, line 3619: The name is confusing, since the it does not match the spec, but I can't think of anything better. Consider documenting why we need this.

Is `SamplingMessage` v1 intended to be replaced by v2 in the future?

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3521,6 +3599,85 @@ public SamplingMessage build() {
...
+		@JsonProperty("role") Role role,
+		@JsonProperty("content")
+		@JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
+		List<Content> content) { // @formatter:on
```
Kehrlann, line 3623: [schema](https://modelcontextprotocol.io/specification/2025-11-25/schema#samplingmessage) seems to indicate we should extend `Meta`

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpStreamableServerSession.java
```diff
@@ -150,6 +152,10 @@ public String getId() {
 		return this.id;
 	}
 
+	void setNegotiatedProtocolVersion(String version) {
+		this.negotiatedProtocolVersion.set(version);
+	}
+
```
Kehrlann, line 158: This cannot change over the lifetime of the session, can it? Should we guard against it?

Kehrlann (commented):
There's a "MUST" about [tool use / result balance](https://modelcontextprotocol.io/specification/2025-11-25/client/sampling#tool-use-and-result-balance) that seems to be missing ; but I'm not fully sure where this should even be enforced. I guess this can be handled by the implementer?

Also, it'd be really neat to have at least one happy-path integration test.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
chemicL: Looking at this property more closely made me realize that it wasn't included because it might be removed.. Looking more closely I realized that Sampling itself is being deprecated and will be removed. Thank you for your careful review, but with the information from the upcoming spec I will close this PR...

chemicL:
Thank you for the careful review @Kehrlann. In light of the discoveries described in https://github.com/modelcontextprotocol/java-sdk/issues/693#issuecomment-4674557814 I am closing this PR.

PR #1019, opened 2026-06-10
Return void from McpStatelessSyncServer#closeGracefully instead of Mono

tzolov:


chemicL (dismissed):
Approving, thanks. Please update the MIGRATION-2.0.md doc before you merge.

PR #1020, opened 2026-06-10
Add schemas for form-based elicitation

### LINKED ISSUE #691: SEP-1330: Elicitation Enum Schema Improvements and Standards Compliance
This is a tracking issue for https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1330

Kehrlann:
Adds builder classes to all new enum schema types introduced for form-based elicitation ([SEP-1330](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1330))
ConformanceServlet is updated to use the new builders and serves as a reference.

Closes #691

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -6,6 +6,7 @@ package io.modelcontextprotocol.spec;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3888,6 +3889,636 @@ public CreateMessageResult build() {
...
+			private Builder(String constValue, String title) {
+				Assert.notNull(constValue, "constValue must not be null");
+				Assert.notNull(title, "title must not be null");
+				this.constValue = constValue;
+				this.title = title;
+			}
+
+			public EnumSchemaOption build() {
+				return new EnumSchemaOption(constValue, title);
+			}
```
chemicL: Since we have just two required fields and nothing optional, we can skip the builder perhaps?

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3888,6 +3889,636 @@ public CreateMessageResult build() {
...
+	 *
+	 * @param title Optional title for the enum field
+	 * @param description Optional description for the enum field
+	 * @param enumValues Array of enum values to choose from
+	 * @param enumNames Optional display names for enum values (non-standard per JSON
+	 * Schema 2020-12)
+	 * @param defaultValue Optional default value
+	 * @deprecated Use {@link TitledSingleSelectEnumSchema} instead
+	 */
+	@Deprecated
```
chemicL, line 3944: Does it make sense a new API that is deprecated on day 1?

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3888,6 +3889,636 @@ public CreateMessageResult build() {
...
+				return this;
+			}
+
+			public Builder description(String description) {
+				this.description = description;
+				return this;
+			}
+
+			public Builder enumValues(List<String> enumValues) {
+				Assert.notNull(enumValues, "enum must not be null");
```
chemicL:
```suggestion
				Assert.notNull(enumValues, "enumValues must not be null");
```

chemicL: Same suggestion for other places

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3888,6 +3889,636 @@ public CreateMessageResult build() {
...
+			}
+
+			public Builder enumValues(String... enumValues) {
+				Assert.notNull(enumValues, "enum must not be null");
+				this.enumValues = new ArrayList<>(Arrays.asList(enumValues));
+				return this;
+			}
+
+			public Builder enumValue(String enumValue) {
+				if (this.enumValues == null) {
```
chemicL: I noticed we don't check whether `enumValue` itself is null. Should we?

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3888,6 +3889,636 @@ public CreateMessageResult build() {
...
+				return this;
+			}
+
+			public Builder enumValues(String... enumValues) {
+				Assert.notNull(enumValues, "enum must not be null");
+				this.enumValues = new ArrayList<>(Arrays.asList(enumValues));
+				return this;
+			}
+
+			public Builder enumValue(String enumValue) {
```
chemicL: While I understand the `List` and `varargs` overloads, this one looks suspicious because it can append to the list. There is no corresponding remove operation and it can work counter-intuitively. I'd suggest we remove it.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3888,6 +3889,636 @@ public CreateMessageResult build() {
...
+				return this;
+			}
+
+			public Builder oneOf(EnumSchemaOption... oneOf) {
+				Assert.notNull(oneOf, "oneOf must not be null");
+				this.oneOf = new ArrayList<>(Arrays.asList(oneOf));
+				return this;
+			}
+
+			public Builder addOneOf(EnumSchemaOption option) {
```
chemicL: Same as before - I am against modifying the contents - let's just accept varargs and list and just replace the list.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
```diff
@@ -3888,6 +3889,636 @@ public CreateMessageResult build() {
...
+	@JsonInclude(JsonInclude.Include.NON_ABSENT)
+	@JsonIgnoreProperties(ignoreUnknown = true)
+	public record TitledSingleSelectEnumSchema( // @formatter:off
+		@JsonProperty("title") String title,
+		@JsonProperty("description") String description,
+		@JsonProperty("oneOf") List<EnumSchemaOption> oneOf,
+		@JsonProperty("default") String defaultValue) { // @formatter:on
+
+		public TitledSingleSelectEnumSchema {
+			Assert.notEmpty(oneOf, "oneOf must not be empty");
```
chemicL, line 4124: Since `oneOf` must not be empty, let's make it the builder factory argument?

chemicL (changes requested):
Thanks! I left a few comments - they apply to all occurrences of a given pattern in the new code, I didn't repeat it in all the new types.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
Kehrlann: I think so, if we plan to introduce it.
Spring Boot did this for Jackson 2 support.

Otherwise, we could just skip this altogether since users can do Map<String, Object>

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
Kehrlann: It's the case for every of one of these enums, but it makes for a super ugly API. I prefer `.oneOf(...)` with a check on `.build()`

Stylistically I would even rename `.builder(...)` itself to `.oneOf(...)` but that's inconsistent with the other builders in this file.

mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
Kehrlann: The builder allows us to enforce non-null on the `value` and `title` fields ; while allowing `null` values when deserializing from the wire.

I'm ok for removing the builder but should we:
1. Disallow null values when deserializing through the compact constructor? Really, it's invalid state, even if it comes from the server.
2. Add a JSON creator factory method to allow for null values?

chemicL (approved):
Thanks, looks excellent!

PR #1023, opened 2026-06-11
Clarify tool validation error messages

### LINKED ISSUE #986: Tool input validation returns with misleading error referencing `outputSchema`
**Bug description**

The tool input validation introduced in https://github.com/modelcontextprotocol/java-sdk/pull/873 returns with a misleading error message referencing `outputSchema` when validation fails. For example, the first test case in 
[ToolInputValidationIntegrationTests#invalidInput_withDefaultValidation_shouldReturnToolError](https://github.com/modelcontextprotocol/java-sdk/blob/c8ab3411b9191ffbbb30f0baaa56b3a26e076dbe/mcp-test/src/test/java/io/modelcontextprotocol/server/ToolInputValidationIntegrationTests.java#L174) leads to the following error message, but only tests for the presence of `age` in the error message substring:

```
Validation failed: structuredContent does not match tool outputSchema. Validation errors: [: required property 'age' not found]
```

Instead of referring to `structuredContent` and `outputSchema`, the error message should clearly indicate that the error is referring to the input data as otherwise it's misleading to consumers/LLMs.

**Environment**

MCP SDK 2.0.0-M3 (via Spring AI 2.0.0-M7)

**Steps to reproduce**

- Run `ToolInputValidationIntegrationTests#invalidInput_withDefaultValidation_shouldReturnToolError` & inspect the produced error message.
- or create an MCP server with tool input validation enabled and call a tool with invalid input data

**Expected behavior**

- A concise error response stating that the listed schema validation errors refer to input parameters.
- Better assertions in the tool input validation test, not only checking for presence of a substring.

See steps to reproduce.

tzolov:
- DefaultJsonSchemaValidator now emits a neutralmessage;
- Call sites in ToolInputValidator and the server output handlers prepend their own context prefix,
- Integration test assertions strengthened accordingly.

Closes #986

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
@@ -4,9 +4,19 @@
 
 package io.modelcontextprotocol.server;
 
-import io.modelcontextprotocol.json.TypeRef;
-import io.modelcontextprotocol.json.McpJsonMapper;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.BiFunction;
+
 import io.modelcontextprotocol.common.McpTransportContext;
+import io.modelcontextprotocol.json.McpJsonMapper;
+import io.modelcontextprotocol.json.TypeRef;
 import io.modelcontextprotocol.json.schema.JsonSchemaValidator;
 import io.modelcontextprotocol.server.McpStatelessServerFeatures.AsyncResourceTemplateSpecification;
 import io.modelcontextprotocol.spec.McpError;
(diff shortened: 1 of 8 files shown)


mcp-core/src/main/java/io/modelcontextprotocol/util/ToolInputValidator.java
```diff
@@ -42,9 +42,10 @@ public static CallToolResult validate(McpSchema.Tool tool, Map<String, Object> a
 		Map<String, Object> args = arguments != null ? arguments : Map.of();
 		var validation = validator.validate(tool.inputSchema(), args);
 		if (!validation.valid()) {
-			logger.warn("Tool '{}' input validation failed: {}", tool.name(), validation.errorMessage());
+			String message = "Tool input validation failed: " + validation.errorMessage();
+			logger.warn("Tool '{}' {}", tool.name(), message);
```
chemicL:
```suggestion
			String message = "input validation failed: " + validation.errorMessage();
			logger.warn("Tool '{}' {}", tool.name(), message);
```

mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java
```diff
@@ -293,9 +293,10 @@ public Mono<CallToolResult> apply(McpTransportContext transportContext, McpSchem
 				var validation = this.jsonSchemaValidator.validate(outputSchema, result.structuredContent());
 
 				if (!validation.valid()) {
-					logger.warn("Tool call result validation failed: {}", validation.errorMessage());
+					String message = "Tool output validation failed: " + validation.errorMessage();
```
chemicL: Perhaps we should add the tool name to the message?

chemicL (changes requested):

chemicL (approved):

PR #1027, opened 2026-06-13
Close servlet streamable HTTP transports on async lifecycle events

lxq19991111:
## Summary

Fixes #1021.

Closes servlet async lifecycle cleanup for Streamable HTTP **POST streaming responses** created by `HttpServletStreamableServerTransportProvider`, and routes SSE write failures through the transport `close()` path instead of removing the logical MCP session from the session registry.

GET listening/replay stream cleanup is now handled by [4186ca13](https://github.com/modelcontextprotocol/java-sdk/commit/4186ca13) ("Fix stream leak when resuming streams with `LAST_EVENT_ID`"); this PR covers the remaining `doPost` path and the write-failure behavior.

## Motivation and Context

The servlet Streamable HTTP transport creates async SSE responses for multiple request paths. When a client disconnects, the current HTTP/SSE response can become unusable while the SDK still keeps the async context and transport state alive. In production this can leave server-side sockets in `CLOSE-WAIT` and keep Tomcat resources tied up until an external timeout or kernel keepalive eventually reclaims them.

`4186ca13` fixed the GET (listening/replay) path. This PR applies the same `AsyncListener` approach to POST streaming responses, where the per-request transport is still not closed on client disconnect, timeout, or async error.

This issue was observed from Spring AI WebMVC usage first, but the lifecycle gap is in the MCP Java SDK core servlet transport.

## Related Issues and PRs

- Fixes #1021.
- Related to #952 and #972: a request-specific SSE write failure should not immediately remove the logical MCP session.
- Related to #920: first transient SSE write failure should not orphan the session.
- Related to #726 only by production symptom (`CLOSE-WAIT` / Tomcat resource growth). #726 targets the older `WebMvcSseServerTransportProvider` / `McpServerSession` path, while this PR targets the core `HttpServletStreamableServerTransportProvider` Streamable HTTP servlet path.
- Follow-up logical session eviction after repeated keep-alive failures is handled separately in #1028.

## Changes

- Register async lifecycle cleanup for POST streaming responses.
- Close the POST streaming transport through the transport `close()` path on handling failures.
- Close only the current transport on SSE write failure instead of removing the logical MCP session.
- Add focused regression coverage for POST streaming response cleanup and write-failure behavior.

## Rationale

A TCP/SSE stream lifecycle is not the same as an MCP logical session lifecycle.

A servlet async error, timeout, completion, or response write failure proves that the current HTTP/SSE transport is no longer usable. It does not necessarily prove that the logical MCP session should be removed from the session registry.

[...]

Issue #952, opened 2026-05-10
`HttpServletStreamableServerTransportProvider` removes session on non-fatal failure


labkey-jeckels:
**Bug description**
We use the MCP Java SDK as via our dependency on Spring AI. Our CI/CD test coverage for our MCP implementation fails regularly with a race condition when dealing with intentional error conditions. We could not reproduce it locally on dev machines, but via logging eventually found a workaround to what appears to be a problem in HttpServletStreamableTransportProvider.

**Environment**
Versions 1.1.1, 1.1.2, and 2.0.0-M2 of the Java SDK. Prerelease versions of Spring AI 2.0, including 2.0.0-M5 and 2.0.0-M4. Java 25.

**Steps to reproduce**

See attached for a standalone repro. The error manifested for us as a race condition but always of this form:

```
  java.lang.RuntimeException: MCP session with server terminated
  at io.modelcontextprotocol.spec.McpClientSession.lambda$dismissPendingResponses$1(McpClientSession.java:128)
  at java.base/java.util.concurrent.ConcurrentHashMap.forEach(ConcurrentHashMap.java:1619)
  at io.modelcontextprotocol.spec.McpClientSession.dismissPendingResponses(McpClientSession.java:126)
  at io.modelcontextprotocol.spec.McpClientSession.close(McpClientSession.java:304)
  at io.modelcontextprotocol.client.LifecycleInitializer$DefaultInitialization.close(LifecycleInitializer.java:225)
  at io.modelcontextprotocol.client.LifecycleInitializer.handleException(LifecycleInitializer.java:257)
  at io.modelcontextprotocol.client.transport.HttpClientStreamableHttpTransport.handleException(HttpClientStreamableHttpTransport.java:231)
  at io.modelcontextprotocol.client.transport.HttpClientStreamableHttpTransport.lambda$sendMessage$33(HttpClientStreamableHttpTransport.java:641)
  at reactor.core.publisher.FluxOnErrorReturn$ReturnSubscriber.onError(FluxOnErrorReturn.java:172)
  at reactor.core.publisher.FluxOnErrorResume$ResumeSubscriber.onError(FluxOnErrorResume.java:107)
  at reactor.core.publisher.Operators.error(Operators.java:198)
  at reactor.core.publisher.FluxError.subscribe(FluxError.java:44)
  at reactor.core.publisher.Flux.subscribe(Flux.java:8888)
[...]

Issue #920, opened 2026-04-15
HttpServletStreamableServerTransportProvider orphans sessions on first transient SSE write failure (no grace period)


andersenleo:
## Summary

In `HttpServletStreamableServerTransportProvider`, a single failed SSE write
immediately removes the session from the in-memory map, so the client's next
POST gets `Session not found` even though the failure was transient (LB
response-timeout, NEG rebalance, pod eviction, laptop sleep, mobile network
blip). The client is forced into a full `initialize` round-trip and loses
any server-side session state.

Adding a short, configurable **grace period** before `sessions.remove(...)`
— during which a reconnect with the same session id reattaches — would solve
this without touching where sessions are stored or how resumability works.

A secondary, unrelated observation about a duplicate `asyncContext.complete()`
call is noted at the bottom.

## Relationship to existing work

I want to flag upfront how this differs from related open issues/PRs, so it's
easy to triage.

- **#914** (`McpSessionStore` SPI) — changes *where* sessions live
  (pluggable store: Redis, JDBC, Hazelcast). Great for restart/cluster
  scenarios. Orthogonal to this issue: even with a Redis-backed store, one
  failed write to one client still triggers `sessions.remove(sessionId)`
  and orphans **that** client's session.
- **#830** (`Last-Event-ID` resumability) — adds an event store and wires
  replay through `GET /mcp`. Complementary to this issue but requires the
  session to still exist when the client reconnects. With today's eager
  removal, replay is unreachable for the case described here.
- **#107** (server-restart `Session not found`) — same symptom, different
  trigger (process restart vs in-process transient write failure). #914
  addresses #107; neither addresses the scenario below.
- **#274 / #738 / #376 / #201** — persistence / clustering angles, all in
  the same family as #914.

The ask here is narrower than any of those: **don't drop the session on the
first transient write failure**.

## Version

- `io.modelcontextprotocol.sdk:mcp-core` 1.1.0
- JDK 25, Tomcat (Spring Boot 4), behind a GKE HTTP(S) LB

## Reproduction

1. Run any gateway built on `HttpServletStreamableServerTransportProvider`
   behind an L7 LB that caps per-response duration below the session lifetime
   (GKE `BackendConfig.timeoutSec` ≤ 60 s is a clean repro).
2. Connect an MCP client (seen with `claude-code`, but any streamable-HTTP
   client triggers it).
3. Wait for the LB to close the SSE stream (~60 s in our case).
4. Observe — server log:

   ```
[...]

PR #1028, opened 2026-06-14
Evict streamable HTTP sessions after failed keep-alive pings

### LINKED ISSUE #1022: [BUG] KeepAliveScheduler does not evict sessions after ping failure, causing unbounded CPU growth
## Description

When `keep-alive-interval` is configured on a Streamable HTTP MCP Server, `KeepAliveScheduler` periodically sends ping messages to all registered `McpStreamableServerSession` instances. When a session's stream is unavailable (client already disconnected), the ping fails and a WARN is logged, but **the session is never removed from the session map**. Over time, dead sessions accumulate indefinitely, causing increasing CPU usage as the scheduler iterates over a growing list of unreachable sessions every interval.

## Environment

- Spring AI: 1.1.4
- MCP Java SDK: (bundled with Spring AI 1.1.4)
- Java: JDK 25
- Server: Tomcat (embedded via Spring Boot 3.4.1)
- Transport: Streamable HTTP (`spring.ai.mcp.server.protocol=STREAMABLE`)

## Configuration

```yaml
spring:
  ai:
    mcp:
      server:
        type: SYNC
        protocol: STREAMABLE
        streamable-http:
          mcp-endpoint: /mcp
          keep-alive-interval: 30s
```

## Steps to Reproduce

1. Deploy an MCP Server with Streamable HTTP transport and `keep-alive-interval: 30s`
2. Have multiple MCP clients connect, call tools, then disconnect (close TCP)
3. Wait a few minutes and observe logs and CPU usage

## Observed Behavior

Logs fill with repeated warnings every 30 seconds, one per dead session:

```
WARN [io.modelcontextprotocol.util.KeepAliveScheduler]
Failed to send keep-alive ping to session McpStreamableServerSession@3656e950: Stream unavailable for session d0a2a5b9-d427-49f6-8f0b-6c37293e2e35

WARN [io.modelcontextprotocol.util.KeepAliveScheduler]
Failed to send keep-alive ping to session McpStreamableServerSession@917a226: Stream unavailable for session 5e87f163-f127-4793-9fec-cd35b6a59327

WARN [io.modelcontextprotocol.util.KeepAliveScheduler]
Failed to send keep-alive ping to session McpStreamableServerSession@5795b3a2: Stream unavailable for session ...
```

- Dead sessions are **never** removed from the internal map
- CPU usage grows linearly with the number of accumulated dead sessions
- The `boundedElastic` thread pool is consumed by failed ping attempts
- Eventually causes CPU alerts and degrades ability to serve new requests

In our production environment, CPU went from normal to sustained **135%+** within hours of deployment, triggered by ~50 disconnected clients accumulating dead sessions.

## Expected Behavior

When a ping fails (stream unavailable), `KeepAliveScheduler` should:

1. Increment a failure counter for that session
2. After N consecutive failures (e.g., 2-3), call `session.close()` and remove it from the session map
3. Log the eviction at INFO level

## Impact

- **Severity: High** — causes progressive CPU degradation in production
- Combined with the CLOSE-WAIT socket leak issue, can render a server completely unusable
- No recovery without restart — dead sessions accumulate forever

## Workaround

Disable keep-alive entirely:

```yaml
spring:
  ai:
    mcp:
      server:
        streamable-http:
          keep-alive-interval: 0s
```

This stops the CPU waste but also removes the ability to detect stale sessions proactively.

## Suggested Fix

In `KeepAliveScheduler`, after a failed ping:

```java
// Pseudocode
void sendKeepAlive() {
    Iterator<Map.Entry<String, McpStreamableServerSession>> it = sessions.entrySet().iterator();
    while (it.hasNext()) {
        Map.Entry<String, McpStreamableServerSession> entry = it.next();
        try {
            entry.getValue().sendPing();
            failureCounters.remove(entry.getKey());
        } catch (Exception e) {
            int failures = failureCounters.merge(entry.getKey(), 1, Integer::sum);
            if (failures >= MAX_FAILURES) {
                entry.getValue().close();
                it.remove();
                failureCounters.remove(entry.getKey());
                log.info("Evicted dead session: {}", entry.getKey());
            }
        }
    }
}
```

lxq19991111:
## Summary

Fixes #1022.

Evict dead Streamable HTTP sessions after repeated keep-alive ping failures.

When `keep-alive-interval` is enabled, `KeepAliveScheduler` currently detects dead sessions by logging failed pings, but the failed sessions remain in the Streamable HTTP provider's session registry. Over time, the scheduler keeps pinging the same unreachable sessions every interval, causing repeated WARN logs and CPU usage that grows with the number of accumulated dead sessions.

This PR adds keep-alive success/failure hooks and lets the Streamable HTTP servlet provider evict the logical session after consecutive keep-alive failures.

## Dependency / Review Order

Depends on #1027.

#1027 establishes the transport-level lifecycle behavior: servlet async lifecycle events and SSE write failures close only the current HTTP/SSE transport while keeping the logical MCP session registered.

This PR builds on that behavior by evicting the logical MCP session only after repeated keep-alive ping failures. After #1027 lands, this PR should be rebased on top of it so the keep-alive eviction behavior does not conflict with transport-level write-failure cleanup.

## Related Issues and PRs

- Fixes #1022.
- Depends on #1027 for servlet async lifecycle cleanup and write-failure semantics.
- Related to #952 and #972: this PR should not be interpreted as deleting the logical session on a single request-specific write failure. That behavior belongs to #1027 / #972.
- Related to #920: transient write failures should not immediately orphan a session, so this PR uses a consecutive-failure threshold rather than evicting on the first failure.

## Changes

- Add success and failure callbacks to `KeepAliveScheduler`.
- Keep the new scheduler callbacks backward compatible by defaulting them to no-op handlers.
- Track consecutive keep-alive ping failures in `HttpServletStreamableServerTransportProvider`.
- Reset the failure count after a successful keep-alive ping.
- Evict and close a Streamable HTTP session after 3 consecutive keep-alive ping failures.
- Clear failure counts on session deletion and provider shutdown.
- Guard failure-count reset and eviction with session identity checks so stale keep-alive results cannot affect a replaced session.
- Add focused tests for scheduler callbacks and Streamable HTTP keep-alive eviction behavior.

## Rationale

The issue suggested doing the failure counter and session removal directly in `KeepAliveScheduler`. This PR intentionally keeps `KeepAliveScheduler` transport-agnostic.

`KeepAliveScheduler` only receives a `Supplier<Flux<McpSession>>`. It does not own the Streamable HTTP provider's session registry, and it cannot safely remove entries from that registry directly.

The Streamable HTTP servlet provider owns the `sessions` map, so it also owns eviction. The scheduler reports ping success or failure through hooks, and the provider decides whether to reset the failure count, keep the session, or evict it.

This PR also avoids evicting on the first failure. A single failed ping can be transient, especially around load balancer timeouts, reconnects, request-specific streams, or temporary transport unavailability. Requiring 3 consecutive failures provides a conservative boundary between transient transport failure and a logical session that is no longer reachable.

Session TTL is intentionally left out of this PR. TTL may still be useful as a separate defense-in-depth policy, but this PR focuses only on sessions that keep-alive has already identified as repeatedly unreachable.

## Scope

This PR intentionally focuses on:

- `KeepAliveScheduler`
- Streamable HTTP servlet session eviction in `HttpServletStreamableServerTransportProvider`

Out of scope:

- Servlet async lifecycle cleanup and write-failure retention, handled in #1027
- Explicit session TTL
- Configurable keep-alive failure threshold
- Session persistence / distributed session storage
- Last-Event-ID resumability / event store support
- Deprecated `HttpServletSseServerTransportProvider`

## How Has This Been Tested?

```bash
mvn -pl mcp-core -Dtest=KeepAliveSchedulerTests,HttpServletStreamableServerTransportProviderTests test
mvn -pl mcp-core test
```

The added scheduler tests cover:

- null success/failure callback validation
- success callback invocation after successful ping
- failure callback invocation after failed ping
- callback exceptions do not stop the scheduler

The added provider tests cover:

- first keep-alive failure does not evict the session
- repeated keep-alive failures evict and close the session
- successful keep-alive resets the failure count
- stale failures from an old session instance do not evict a replacement session with the same id

## Types of changes

- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io/)

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
@@ -24,6 +24,7 @@ import io.modelcontextprotocol.server.McpTransportContextExtractor;
 import io.modelcontextprotocol.spec.HttpHeaders;
 import io.modelcontextprotocol.spec.McpError;
 import io.modelcontextprotocol.spec.McpSchema;
+import io.modelcontextprotocol.spec.McpSession;
 import io.modelcontextprotocol.spec.McpStreamableServerSession;
 import io.modelcontextprotocol.spec.McpStreamableServerTransport;
 import io.modelcontextprotocol.spec.McpStreamableServerTransportProvider;
@@ -87,6 +88,8 @@ public class HttpServletStreamableServerTransportProvider extends HttpServlet
 
 	public static final String FAILED_TO_SEND_ERROR_RESPONSE = "Failed to send error response: {}";
 
+	private static final int KEEP_ALIVE_FAILURE_THRESHOLD = 3;
+
 	/**
 	 * The endpoint URI where clients should send their JSON-RPC messages. Defaults to
 	 * "/mcp".
@@ -106,6 +109,8 @@ public class HttpServletStreamableServerTransportProvider extends HttpServlet
 	 * Map of active client sessions, keyed by mcp-session-id.
 	 */
 	private final ConcurrentHashMap<String, McpStreamableServerSession> sessions = new ConcurrentHashMap<>();
+
+	private final ConcurrentHashMap<String, Integer> keepAliveFailureCounts = new ConcurrentHashMap<>();
 
 	private McpTransportContextExtractor<HttpServletRequest> contextExtractor;
 
@@ -158,6 +163,8 @@ public class HttpServletStreamableServerTransportProvider extends HttpServlet
 				.builder(() -> (isClosing) ? Flux.empty() : Flux.fromIterable(sessions.values()))
 				.initialDelay(keepAliveInterval)
 				.interval(keepAliveInterval)
+				.onSuccess(this::resetKeepAliveFailures)
+				.onFailure(this::handleKeepAliveFailure)
 				.build();
 
 			this.keepAliveScheduler.start();
@@ -231,8 +238,10 @@ public class HttpServletStreamableServerTransportProvider extends HttpServlet
 			});
 
 			this.sessions.clear();
+			this.keepAliveFailureCounts.clear();
 		}).then().doOnSuccess(v -> {
 			sessions.clear();
+			keepAliveFailureCounts.clear();
 			logger.debug("Graceful shutdown completed");
 			if (this.keepAliveScheduler != null) {
 				this.keepAliveScheduler.shutdown();
@@ -445,6 +454,7 @@ public class HttpServletStreamableServerTransportProvider extends HttpServlet
 				McpStreamableServerSession.McpStreamableServerSessionInit init = this.sessionFactory
 					.startSession(initializeRequest);
 				this.sessions.put(init.session().getId(), init.session());
+				this.keepAliveFailureCounts.remove(init.session().getId());
 
 				try {
 					McpSchema.InitializeResult initResult = init.initResult().block();
@@ -614,6 +624,7 @@ public class HttpServletStreamableServerTransportProvider extends HttpServlet
 		try {
 			session.delete().contextWrite(ctx -> ctx.put(McpTransportContext.KEY, transportContext)).block();
 			this.sessions.remove(sessionId);
+			this.keepAliveFailureCounts.remove(sessionId);
 			response.setStatus(HttpServletResponse.SC_OK);
 		}
 		catch (Exception e) {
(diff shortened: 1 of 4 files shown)


ShemTovYosef:
Does it take into account this issue too? https://github.com/modelcontextprotocol/java-sdk/pull/972

lxq19991111:
Hi @ShemTovYosef, thanks for pointing this out.

Not directly. #972 addresses a transport-level write failure case: a request-specific SSE response write failure should not immediately remove the logical MCP session.

This PR is targeting a different layer: logical session eviction after repeated keep-alive ping failures, which is intended to identify sessions that are actually no longer reachable.

The layering I’m aiming for is:

- #1027 / #972: write failures or servlet async lifecycle events close only the current HTTP/SSE transport, while keeping the logical MCP session registered.
- #1028: after that transport-level behavior is in place, evict the logical session only after consecutive keep-alive ping failures.

So I’ve marked #1028 as depending on #1027. After #1027 lands, I can rebase this PR on top of it so the keep-alive eviction behavior does not conflict with the transport lifecycle cleanup.

lxq19991111:
@tzolov @Kehrlann an API-design question before finalizing this PR.

A servlet-only hook doesn't help out-of-tree providers: Spring AI's WebMVC and WebFlux Streamable HTTP providers each own their session registry and construct their own `KeepAliveScheduler` (these transports were migrated out of this repo in #805), so they need a way to plug their own eviction into the scheduler.

One clarification on the visibility, since it constrains the options: `KeepAliveScheduler` is in `io.modelcontextprotocol.util`, while `HttpServletStreamableServerTransportProvider` is in `io.modelcontextprotocol.server.transport`. A package-private hook can't be reached even by the SDK's own servlet provider across that package boundary, so "keep it internal" isn't actually viable.

So the real choice is:

1. No new public surface — the provider stops using `KeepAliveScheduler` and runs its own ping loop + failure tracking + eviction. This duplicates the scheduler and diverges the servlet path from the other transports.
2. Expose the success/failure callback on the public `KeepAliveScheduler` builder (it already takes the session source via the public `builder(Supplier<Flux<McpSession>>)`), so the servlet provider — and the out-of-tree Spring AI WebMVC/WebFlux providers — can wire eviction through the same scheduler. The servlet provider would be the first consumer.

I'm inclined toward option 2 since it avoids duplication across the migrated Spring transports, but this is a public-surface decision, so I'd like your call before I lock the signature.

Since #1022 is tagged "ready for work", this PR is ready for review whenever you have time. It depends on #1027 and I'll rebase on top of it once that lands.

lxq19991111:
#1130 answers the open API question here by adding a public `KeepAliveScheduler.Builder#onPingFailure` hook, and goes further with idle-session eviction. Happy to close this in favor of #1130 once it lands.

PR #1037, opened 2026-06-24
fix: standardize resource not found error code

### LINKED ISSUE #1001: SEP-2164: Standardize resource not found error code (-32602)
Tracking implementation of [SEP-2164](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2164) for the 2026-07-28 MCP spec release.

CrypticCortex:
## Summary
- Align `McpError.RESOURCE_NOT_FOUND` with SEP-2164 by using JSON-RPC `INVALID_PARAMS` (`-32602`).
- Update the resource-not-found error code regression expectation.

## Test plan
- `./mvnw -pl mcp-core -Dtest=McpErrorTest test`
- `./mvnw -pl mcp-core test`
- `./mvnw -pl mcp-core spring-javaformat:validate`
- `git diff --check`

Closes #1001

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/spec/McpSchema.java
@@ -151,7 +151,7 @@ public final class McpSchema {
 		/**
 		 * Resource not found.
 		 */
-		public static final int RESOURCE_NOT_FOUND = -32002;
+		public static final int RESOURCE_NOT_FOUND = INVALID_PARAMS;
 
 		/**
 		 * URL elicitation is required before the request can proceed.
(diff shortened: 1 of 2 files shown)


Kehrlann:
This is incorrect, now EVERY `RESOURCE_NOT_FOUND` usage will throw `INVALID_PARAMS`, even when it should be `RESOURCE_NOT_FOUND`, e.g. when a [Resource does not exist](https://modelcontextprotocol.io/specification/2025-11-25/server/resources#error-handling)

Additionally, this is a breaking change, unsure whether we will merge this before we create a 3.x branch.

CrypticCortex:
Thanks, you’re right. change is too broad because it changes every `RESOURCE_NOT_FOUND` caller, including cases that should keep the existing resource-not-found code.

I prepared a narrower local fix: keep `RESOURCE_NOT_FOUND = -32002`, and only have the SDK-owned missing `resources/read` fallback return SEP-2164 `INVALID_PARAMS` (`-32602`) with `data: { uri }`, with tests for both behaviors.

Would you prefer that update now, or should this wait for 3.x / protocol-versioned behavior?

Kehrlann:
Please apply it. I'll assign the original issue to you. Could you please comment on #1001 so I can assign you?

For when exactly we ship it, I'll keep you posted. Assigning to 3.x until we decide what to do.

PR #1046, opened 2026-06-29
Add initialize request metadata support

### LINKED ISSUE #940: McpClient should allow passing _meta in InitializeRequest
The MCP specification defines `_meta` as a valid field on every request, including `initialize`, via the `RequestParams` base type:

```typescript
// schema.ts (2025-11-25)
interface RequestParams {
  _meta?: {
    progressToken?: ProgressToken;
    [key: string]: unknown;
  };
}

interface InitializeRequestParams extends RequestParams {
  protocolVersion: string;
  capabilities: ClientCapabilities;
  clientInfo: Implementation;
}
```

`McpSchema.InitializeRequest` in the Java SDK correctly models this (the `_meta` field was added in #344). However, `McpSyncClient#initialize()` / `LifecycleInitializer#doInitialize` always constructs the request with the 3-arg convenience constructor, hardwiring `_meta` to `null`. There is no API to populate it.

### Impact

Servers that rely on `_meta` of the `initialize` request will always receive `null` metadata when called from a Java client, even though the spec explicitly allows it. This is inconsistent with `tools/call`, where `CallToolRequest._meta` is fully caller-controlled.

### Workaround (current)

The only workaround is to wrap the `McpClientTransport` and intercept the outgoing `initialize` `JSONRPCRequest` in `sendMessage`, rebuilding it with `_meta` populated. This works but couples application code to internal wire-format details (`JSONRPCRequest`, `InitializeRequest` record structure) that should be opaque to callers.

### Proposed solution

Expose `_meta` on the `McpClient` builder, consistent with the pattern used in PR #906 for paginated list methods:

```java
// Option A — builder-level (static meta per client lifetime)
McpClient.sync(transport)
    .initializeMeta(Map.of("server_id", "…", "invocation_id", "…"))
    .build();

// Option B — overloaded initialize() (dynamic meta per call)
client.initialize(Map<String, Object> meta);
```

### Related

- #344 — added `_meta` field to `McpSchema` classes (schema layer, no client API)
- #907 / PR #906 — added `_meta` to paginated list client methods (`listTools`, etc.)
- [MCP spec — General fields: `_meta`](https://modelcontextprotocol.io/specification/2025-11-25/basic#meta)
- [MCP spec schema (2025-11-25)](https://github.com/modelcontextprotocol/specification/blob/main/schema/2025-11-25/schema.ts)

smf-h:
## Summary
- add optional `_meta` support for MCP initialize requests
- expose initialize-request metadata on async and sync client builders
- propagate metadata through lifecycle initialization while defensively copying user-provided maps

## Validation
- `.\mvnw.cmd -pl mcp-core "-Dtest=LifecycleInitializerTests,McpAsyncClientTest,McpSyncClientTest" test`
- Result: `BUILD SUCCESS`; `Tests run: 37, Failures: 0, Errors: 0, Skipped: 0`

Fixes #940

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/client/LifecycleInitializer.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/client/LifecycleInitializer.java
@@ -7,7 +7,9 @@ package io.modelcontextprotocol.client;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 
@@ -89,6 +91,8 @@ class LifecycleInitializer {
 
 	private final McpSchema.Implementation clientInfo;
 
+	private final Map<String, Object> initializeRequestMeta;
+
 	private List<String> protocolVersions;
 
 	private final AtomicReference<DefaultInitialization> initializationRef = new AtomicReference<>();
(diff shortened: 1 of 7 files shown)


Kehrlann:
I'm wondering whether we should really implement #940 . If we do, we need to decide how to implement it.
Let's start the discussion there, and, if need be, we can open a PR later on.

Issue #1047, opened 2026-06-30
SSE client rejects valid `retry:` field and ignores reconnection timing (Streamable HTTP)


imankhadiv:
**Bug description**

The client SSE line parser in `ResponseSubscribers` rejects the standard SSE `retry:` field, throwing:

```
io.modelcontextprotocol.spec.McpTransportException: Invalid SSE response. Status code: 200 Line: retry: 500
```

Per the SSE specification, `retry:` sets the stream's reconnection time, and unknown fields MUST be ignored (never error the stream). Because of this, the Streamable HTTP client:
1. can error a live SSE stream when the server sends a `retry:` line, and
2. does not honor the server-provided reconnection delay, reconnecting immediately.

This is the MUST-level `client-sse-retry-timing` failure already noted as a known limitation in `conformance-tests/VALIDATION_RESULTS.md`.

**Environment**
- java-sdk `main` (2.0.1-SNAPSHOT)
- Java 17+
- Transport: `HttpClientStreamableHttpTransport` (client), SSE / Streamable HTTP

**Steps to reproduce**
Run the official MCP conformance `sse-retry` client scenario:

```bash
./mvnw clean package -DskipTests -pl conformance-tests/client-jdk-http-client -am
npx -y @modelcontextprotocol/conformance client \
  --command "java -jar conformance-tests/client-jdk-http-client/target/client-jdk-http-client-2.0.1-SNAPSHOT.jar" \
  --scenario sse-retry
```

Observed:
```
Error: Invalid SSE response. Status code: 200 Line: retry: 500
[client-sse-retry-timing ] FAILURE  Client MUST respect the retry field (reconnected ~0ms instead of 500ms)
[client-sse-last-event-id] WARNING  Client SHOULD send Last-Event-ID on reconnection
OVERALL: FAILED
```

**Expected behavior**
- The SSE parser parses/ignores `retry:` (and any unknown SSE field) without erroring the stream.
- On reconnection after a drop, the client waits the server-specified `retry` interval before reconnecting.

**Minimal reproducible example**
Feeding the SSE lines `id: e1` / `retry: 500` / `data: hello` / (blank) to the SSE line subscriber currently throws `McpTransportException` instead of yielding a single event.

**Proposed scope (two parts)**
1. Parser robustness (small, self-contained): parse `retry:` and ignore unknown fields in `ResponseSubscribers`. *(Implemented locally with unit tests.)*
[...]

PR #1051, opened 2026-07-02
fix(client): ignore unknown SSE fields instead of erroring the stream

imankhadiv:
## Summary
The client SSE line parser in `ResponseSubscribers` errored the whole stream on any line it didn't recognize — e.g. the standard `retry:` field — throwing:

```
io.modelcontextprotocol.spec.McpTransportException: Invalid SSE response. Status code: 200 Line: retry: 500
```

Per the [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html), a client must **ignore fields it does not recognize** rather than treat them as an invalid response. This change makes the parser log unrecognized lines at `debug` and continue.

This addresses the "ignore unknown fields" ask in #1047. As discussed there, resumability / `retry:` reconnection timing is intentionally **not** implemented (it is being removed in the 2026-07-28 spec).

## Changes
- `ResponseSubscribers.SseLineSubscriber`: the fall-through `else` now ignores unrecognized SSE lines instead of emitting `McpTransportException`. Removed the now-unused import.
- No API change.

## Testing
- New `ResponseSubscribersTests` (2 tests): a `retry:` line no longer errors the stream and the event still parses; an arbitrary unknown field is ignored.
- `./mvnw test -pl mcp-core` → green.

## Notes
- The conformance `sse-retry` scenario stays listed in `conformance-tests/conformance-baseline.yml`: this fix removes the crash, but the scenario's retry-timing check is a resumability behavior that is intentionally out of scope.

Refs #1047

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/client/transport/ResponseSubscribers.java
@@ -15,7 +15,6 @@ import org.reactivestreams.Subscription;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import io.modelcontextprotocol.spec.McpTransportException;
 import reactor.core.publisher.BaseSubscriber;
 import reactor.core.publisher.FluxSink;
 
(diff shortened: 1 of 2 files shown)


naroonmaxai-web:
Reviewed and everything looks good to me. Approved!

imankhadiv:
Gentle nudge on this one, @Kehrlann — it's scoped down to just ignoring unrecognized SSE fields per the spec (resumability dropped), with focused tests. Happy to adjust anything. Thanks for taking a look when you have a moment! 🙏

PR #1076, opened 2026-08-01
fix McpTransportContext.create() mutability issue and add tests

### LINKED ISSUE #1077: `McpTransportContext.create` documents an unmodifiable context but stores the caller's map by reference
**Bug description**

`McpTransportContext.create(Map)` is documented to "create an unmodifiable context containing the given metadata":
https://www.javadocs.dev/io.modelcontextprotocol.sdk/mcp-core/2.0.0/io/modelcontextprotocol/common/McpTransportContext.html
It stores the caller's map by reference instead, with no copy and no unmodifiable wrapper. Mutating the map after the call changes the context, and because `equals` and `hashCode` delegate to the map, the context's identity moves with it.

**Environment**

- MCP Java SDK 2.0.0, and `main` at `fd00498` (`2.0.1-SNAPSHOT`). The two are identical here
- Java 17 target, reproduced on Java 21
- `mcp-core` only, no transport or vector store involved

**Steps to reproduce**

1. Build a mutable `Map<String, Object>` and put an entry in it.
2. Pass it to `McpTransportContext.create(...)`.
3. Mutate the map: overwrite the entry, add another, or clear it.
4. Read the context back with `get(...)`.

The context reflects every mutation.

**Expected behavior**

`create(...)` returns a context that is unmodifiable, as the javadoc says. One line does it:

```java
DefaultMcpTransportContext(Map<String, Object> metadata) {
    Assert.notNull(metadata, "The metadata cannot be null");
    this.metadata = Map.copyOf(metadata);
}
```

`Map.copyOf` copies and makes the result unmodifiable. It also rejects null keys and values, so it is a behavior change for anyone who puts a null value in the map today. If that needs to keep working, `Collections.unmodifiableMap(new HashMap<>(metadata))` does the same job without the restriction.

If storing by reference is deliberate, so a caller can update a context in place, then the javadoc is the thing to fix rather than the code. In that case the thread safety expectation is worth spelling out, because the context is read from transport worker threads and the field is neither `volatile` nor guarded.

Tests at https://github.com/modelcontextprotocol/java-sdk/pull/1076.

The SDK's test suite doesn't catch this because every caller of `create(...)` in the repo passes `Map.of(...)`, which is already immutable.

Practical case: The context is captured on one thread and read on the HTTP client's worker threads during the SSE reconnect and the follow-up POST. A caller who builds a `HashMap`, passes it to `create(...)` and then reuses or clears it has a cross thread data race on a field with no `volatile` and no synchronization.

iuliiasobolevska:
## Motivation and Context
Fixes https://github.com/modelcontextprotocol/java-sdk/issues/1077.

## How Has This Been Tested?
Add new tests that reproduce the issue.

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/mcp-core/src/test/java/io/modelcontextprotocol/common/McpTransportContextTests.java
+++ b/mcp-core/src/test/java/io/modelcontextprotocol/common/McpTransportContextTests.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2026-2026 the original author or authors.
+ */
+
+package io.modelcontextprotocol.common;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link McpTransportContext#create(Map)}, which is documented to return an
+ * unmodifiable context.
+ */
+class McpTransportContextTests {
+
+	@Test
+	void createdContextShouldNotSeeLaterWritesToTheSourceMap() {
+		Map<String, Object> metadata = new HashMap<>();
+		metadata.put("tenant", "acme");
+
+		McpTransportContext context = McpTransportContext.create(metadata);
+		metadata.put("tenant", "other");
+
+		assertThat(context.get("tenant")).isEqualTo("acme");
+	}
+
+	@Test
+	void createdContextShouldNotSeeLaterAdditionsToTheSourceMap() {
+		Map<String, Object> metadata = new HashMap<>();
+		metadata.put("tenant", "acme");
+
+		McpTransportContext context = McpTransportContext.create(metadata);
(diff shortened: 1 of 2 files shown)


mcp-core/src/test/java/io/modelcontextprotocol/common/McpTransportContextTests.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2024-2025 the original author or authors.
```
Kehrlann:
```suggestion
 * Copyright 2026-2026 the original author or authors.
```

Kehrlann (commented):
Excellent point. Please fix the header.

Kehrlann (approved):
Updated and ready to merge

PR #1064, opened 2026-07-17
feat: add extensions attribute to ClientCapabilities and ServerCapabilities

YousefHaggy:
Add the optional `extensions` field to `ClientCapabilities` and `ServerCapabilities`, mirroring the MCP extensions framework (SEP-2133) as declared in the draft schema and already shipped by the TypeScript, Python, C#, Go, Rust, and Kotlin SDKs.

Refs modelcontextprotocol/java-sdk#785

## Spec + cross-SDK precedent

- **Spec:** defined on `ClientCapabilities` (and `ServerCapabilities`) in the
  draft schema — [`schema/draft/schema.ts#L785`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/26897cc322f356487da89113451bd16b520b9288/schema/draft/schema.ts#L785)
  (server twin at [`#L882`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/26897cc322f356487da89113451bd16b520b9288/schema/draft/schema.ts#L882)) —
  introduced by [SEP-2133](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2133)
  (merged into the spec 2026-01-26). Type: `{ [key: string]: object }`. Keys are
  extension identifiers following `_meta` naming rules; an empty object means
  "supported, no settings". See also the
  [extensions overview docs](https://modelcontextprotocol.io/docs/extensions/overview).
- **Other SDKs already ship it** on `ClientCapabilities`. Java is the only major
  SDK without it:
  - [TypeScript](https://github.com/modelcontextprotocol/typescript-sdk/blob/1480241e2a2a7f0ceee8e7723b2adcf88579bb36/packages/core-internal/src/types/spec.types.2026-07-28.ts#L754) — `extensions?: { [key: string]: JSONObject };`
  - [Python](https://github.com/modelcontextprotocol/python-sdk/blob/3a6f2996cdd8358957479791e8b26198c07d6a75/src/mcp-types/mcp_types/_types.py#L400) — `extensions: dict[str, dict[str, Any]] | None = None`
  - [C#/.NET](https://github.com/modelcontextprotocol/csharp-sdk/blob/0d34048e5fc0a2e5a579a2d48a17308083a50028/src/ModelContextProtocol.Core/Protocol/ClientCapabilities.cs#L86) — `IDictionary<string, object>? Extensions`
[...]

PR #1062, opened 2026-07-16
Add ttlMs and cacheScope to cacheable result records (SEP-2549)

aboullaite:
Add ttlMs and cacheScope to cacheable result records (SEP-2549)

## Motivation and Context

SEP-2549 introduces TTL caching hints so clients and intermediaries (gateways, proxies) can cache list and read results without re-fetching on every request. The draft spec adds two fields to cacheable result types:

- `ttlMs` (integer, >= 0): how long in milliseconds the client may consider the response fresh, analogous to HTTP `Cache-Control: max-age`
- `cacheScope` (enum: `"private"` / `"public"`): whether the response may be shared across authorization contexts, analogous to HTTP `Cache-Control: private` vs `public`

This PR adds the schema fields and server-side default stamping. The server now emits `ttlMs=0` (immediately stale) and `cacheScope="public"` on all cacheable responses. Server-side configuration hooks for user-defined TTLs are intentionally left out following @Kehrlann's note that those belong with the repositories refactor in #578 (and the related #1034).

Towards #1009

## How Has This Been Tested?

UT + IT

## Breaking Changes

None. Both fields are optional (nullable), existing constructors are preserved as `@Deprecated` overloads.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

**Remaining work for full #1009 coverage** (follow-up PRs):
- Server configuration API for authors to set TTL per method/resource (dependent on #578)
- Client-side TTL response cache with notification-driven invalidation

PR #1048, opened 2026-06-30
Feature/distributed session storage

galm1007:
These are the exact changes in vlin529's [PR](https://github.com/modelcontextprotocol/java-sdk/pull/914)

I have simply resolved the conflict to allow to merge after approval.

PR #1089, opened 2026-08-13
MCP Spec 28-7-2026 Design

omarmahamid:
## Summary

Adds a design proposal — `docs/design/2026-07-28-spec-support.md` — for implementing the
[`2026-07-28`](https://modelcontextprotocol.io/specification/2026-07-28) MCP specification
revision in this SDK.

### Link for md viewer:

https://github.com/omarmahamid/java-sdk/blob/feat/28-7-mcp-spec/docs/design/2026-07-28-spec-support.md

**Docs-only.** No production code changes. The intent is to agree the shape of the work
*before* implementation, per `CONTRIBUTING.md`:

> For non-trivial changes, clarify scope with maintainers in an issue before investing in an
> implementation.

## Why a design document rather than PRs

The SDK currently implements `2024-11-05` through `2025-11-25`. Every one of those revisions
shares one structural assumption: **an MCP conversation is a session established by an
`initialize` handshake**, over which either party may originate JSON-RPC requests.

`2026-07-28` removes that assumption. Six of its nine *major* changes are removals or
inversions of mechanisms `mcp-core` models as load-bearing types — `McpServerSession`,
`McpStreamableServerSession`, `LifecycleInitializer`, `McpTransportSession`,
`McpTransportStream`, and the client-side request handlers for `roots/list`,
`sampling/createMessage` and `elicitation/create`:

| # | Change | SEP |
| --- | --- | --- |
| 1 | Remove protocol-level sessions and `Mcp-Session-Id` | SEP-2567 |
| 2 | Remove the `initialize` / `notifications/initialized` handshake; per-request `_meta` envelope | SEP-2575 |
| 3 | Add `server/discover` (servers **MUST** implement) | SEP-2575 |
| 4 | Replace HTTP `GET` + `resources/subscribe` with `subscriptions/listen` | SEP-2575 |
| 5 | Remove `ping`, `logging/setLevel`, `notifications/roots/list_changed` | SEP-2575 |
| 6 | Move tasks to the `io.modelcontextprotocol/tasks` extension | SEP-2663 |
| 7 | **MRTR** replaces server-initiated requests | SEP-2322 |
| 8 | Required `resultType` on every result | SEP-2322 |
| 9 | Remove SSE resumability (`Last-Event-ID`) | SEP-2575 |

Three properties make piecemeal delivery actively harmful:

- **The changes are coupled.** `resultType` is meaningless without MRTR; MRTR is
  unimplementable without the `_meta` envelope; the envelope is unverifiable without header
  mirroring. Landing them independently produces intermediate states that are on no protocol
  revision at all.
- **The era boundary must be decided once.** Whether "modern vs. legacy" is a runtime flag, a
  parallel type hierarchy, or separate modules determines the shape of every subsequent PR.
- **It is a major version.** Per `VERSIONING.md`, three independent triggers apply, so the
  removals must be batched into one release.

## The finding the design rests on

**This SDK already ships a stateless server.** `McpStatelessAsyncServer`'s own Javadoc:

> "It allows simple horizontal scalability since it does not maintain a session and does not
> require initialization. Each instance of the server can be reached with no prior knowledge
> and can serve the clients with the capabilities it supports."

That is close to a verbatim description of the `2026-07-28` server model. The proposal
therefore does **not** add a parallel server implementation — it promotes the existing
stateless server to be the canonical *modern* server, and leaves the session-based classes as
a deprecated *legacy* path. Two additive SPI extensions are needed: answer one request with a
**stream** (`subscriptions/listen`), and answer one request with **"I need more input"**
(MRTR).

Without that existing asset, `2026-07-28` support would be a ground-up rewrite. This is the
main reason the proposal is tractable.

## What the document contains

Structured as a design proposal (Motivation → Public Interfaces → Proposed Changes →
Security → Compatibility/Migration → Test Plan → Rejected Alternatives → FAQ):

- **Gap analysis** of `mcp-core` at `8ee8ccbc`, requirement by requirement, each classified
  *additive* / *structural* / *wire-format* / *removal*.
- **Full public API surface**: `ProtocolVersions.MCP_2026_07_28`, `McpMetaKeys`, new
  `ErrorCodes` (`-32020` / `-32021` / `-32022`), `DiscoverResult`, `InputRequiredResult`,
  `InputRequests` / `InputResponses`, `ListRootsRequest`, `SubscriptionsListenRequest`,
  `CacheableResult`, `McpRequestContext`, `McpSubscription`, `McpSubscriptionSink`,
  `McpRequestStateCodec`, `McpHeaderCodec`, `McpParamHeaderExtractor`, new exceptions.
- **Seven keyed design decisions** (D1–D7), each with the alternative it rejects.
- **Five implementation phases**, each independently reviewable and each leaving `main` on a
  coherent protocol revision. P1–P2 are non-breaking and could ship in a `2.x` minor.
- **Behaviour matrix** resolving the spec's client/server era matrix to concrete SDK
  behaviour.
- **Test plan** covering unit, integration and conformance, including per-record cases derived
  from the Case A / Case B rules in `CONTRIBUTING.md`.
- **Six rejected alternatives** with reasoning.

## Design decisions worth reviewer attention

**D3 — MRTR is resolved inside the client session layer.** Sampling / elicitation / roots
handler signatures do **not** change; the same handlers are invoked from a bounded retry loop
instead of from an inbound-request dispatcher. This keeps `callTool()` returning
`CallToolResult` rather than a union type, so existing callers recompile unchanged. RA-3
argues the alternative.

**D4 — `resultType` must be a real record component.** A default method on `Result` would not
serialize (Jackson emits record *components*), so every result the SDK emitted would be
missing a spec-required field. That means ~15 records under the `CONTRIBUTING.md` Case B
rules. The interface default is still added, for uniform *reading* across eras.

**D1 — era dispatch at the transport front door**, not a `boolean modern` flag on the existing
session types. The spec says a dual-era server "selects its behavior from how the client
opens" — that is a routing decision on the first message and belongs in one place.

**D6 — the modern client needs a tool-definition cache**, because `x-mcp-header` mirroring
requires the tool's `inputSchema` at `tools/call` time and `tools/call` does not carry it. The
same cache is what makes `ttlMs` / `cacheScope` actionable.

## Open questions for maintainers

1. **Is HTTP+SSE removal in scope for the major release?** SEP-2596 reclassifies it as
   *Deprecated*, not *Removed*, and the feature-lifecycle policy allows a twelve-month window.
   The document proposes removal (the release is already breaking) but explicitly flags
   deprecate-now / remove-next-major as the conservative option and does not insist.
2. **Is the dual-era default correct**, versus modern-only with legacy behind a flag?
3. **Should `mcp-ext-tasks` be a new module**, or should tasks stay in `mcp-core` behind the
   extension capability?
4. **Should an escape hatch expose `InputRequiredResult` directly?** The document defers it
   under `CONTRIBUTING.md`'s "concrete, not speculative" principle, but a real use case would
   change that.
5. **Phasing**: P1–P2 are non-breaking. Ship them in a `2.x` minor to shorten the major-release
   critical path, or hold everything for the major?

## Known behaviour change

One change cannot be made backward-compatible: on the modern path, `notifications/message` is
suppressed for requests that omit `io.modelcontextprotocol/logLevel`. The spec states servers
**MUST NOT** emit it in that case. It leads the migration guide, and `logLevel` is settable
once on the client builder to restore log flow with a one-line change.

## What this PR does not do

- No production code, no schema changes, no dependency changes.
- Does not commit the project to the proposed API — the naming and shape are a starting point
  for discussion.
- Does not supersede the spec-revision GitHub Project board referenced in `ROADMAP.md`; it is
  intended to complement it.

## Checklist

- [ ] Docs-only; `./mvnw clean compile -DskipTests` unaffected
- [ ] No formatting impact (`spring-javaformat` applies to Java sources only)
- [ ] `ROADMAP.md` update to retarget the focus area from `2025-11-25` to `2026-07-28` — folded
      into this PR or filed as a follow-up, reviewer's preference
- [ ] Spec references verified against `modelcontextprotocol.io/specification/2026-07-28`
      (changelog, versioning, MRTR, subscriptions, discover, Streamable HTTP)

### DIFF
--- a/docs/design/2026-07-28-spec-support.md
+++ b/docs/design/2026-07-28-spec-support.md
@@ -1,0 +1,1550 @@
+# JSDP-1: MCP `2026-07-28` Protocol Revision Support in the MCP Java SDK
+
+## Status
+
+| Field | Value |
+| --- | --- |
+| **Status** | Under Discussion |
+| **Proposal type** | Java SDK Design Proposal (JSDP) |
+| **Target SDK version** | `3.0.0` (current `main`: `2.0.1-SNAPSHOT`) |
+| **Spec revision** | [`2026-07-28`](https://modelcontextprotocol.io/specification/2026-07-28) |
+| **Highest revision implemented today** | [`2025-11-25`](https://modelcontextprotocol.io/specification/2025-11-25) |
+| **Breaking change** | Yes — see [Compatibility, Deprecation, and Migration Plan](#compatibility-deprecation-and-migration-plan) |
+| **Affected modules** | `mcp-core`, `mcp-test`, `conformance-tests`, `mcp-bom`, `docs` |
+| **Reference implementation** | LangChain4j client-side implementation, [langchain4j/langchain4j#5881](https://github.com/langchain4j/langchain4j/pull/5881) |
+| **Related SEPs** | SEP-2567, SEP-2575, SEP-2322, SEP-2663, SEP-2243, SEP-2549, SEP-2577, SEP-2596, SEP-2106, SEP-414 |
+
+> **Note on process.** This document is a design proposal prepared for discussion. Per
+> `AGENTS.md` in this repository, no issue, pull request, or discussion has been or will be
+> filed upstream on the basis of this document without an explicit, separately authorised
+> decision by a maintainer with standing in the project.
+
+---
+
+## Table of Contents
+
+1. [Motivation](#motivation)
+2. [Public Interfaces](#public-interfaces)
+3. [Proposed Changes](#proposed-changes)
+4. [Security Hardening](#security-hardening)
+5. [Compatibility, Deprecation, and Migration Plan](#compatibility-deprecation-and-migration-plan)
+6. [Test Plan](#test-plan)
+7. [Reference Implementation](#reference-implementation)
+8. [Documentation Plan](#documentation-plan)
+9. [Rejected Alternatives](#rejected-alternatives)
+10. [FAQ](#faq)
+
+---
+
+## Motivation
+
+### Background
+
+The MCP Java SDK (`io.modelcontextprotocol.sdk`) is the *formal*, specification-owned Java
+implementation of the Model Context Protocol. Its stated job, per `CONTRIBUTING.md`, is to
+implement the MCP specification — and per `ROADMAP.md`, the project's declared goal is
+**Tier 1 SDK support**: "fully support all the upcoming specification features on the day of
+its release."
+
+The SDK currently implements protocol revisions `2024-11-05` through `2025-11-25`. Every one
+of those revisions shares a single structural assumption: **an MCP conversation is a session
+established by an `initialize` handshake**, over which either party may originate JSON-RPC
+requests at any time.
+
+The `2026-07-28` revision deletes that assumption.
+
+### Motivating Question
+
+> How does an SDK whose entire client and server core is built around a stateful
+> `initialize` handshake and bidirectional request origination absorb a revision that removes
+> both — without abandoning the ~4 years of deployed servers and clients that still speak the
+> handshake?
+
+This is not a feature-addition question. Six of the nine *major* changes in `2026-07-28`
+are removals or inversions of mechanisms that `mcp-core` models as load-bearing types:
+`McpServerSession`, `McpStreamableServerSession`, `LifecycleInitializer`,
+`McpTransportSession`, `McpTransportStream`, and the client-side request handlers for
+`roots/list`, `sampling/createMessage`, and `elicitation/create`.
+
+### What the `2026-07-28` Revision Changes
+
+Summarised from the [official changelog](https://modelcontextprotocol.io/specification/2026-07-28/changelog).
+
+#### Major changes
+
+| # | Change | SEP | Structural impact on this SDK |
+| --- | --- | --- | --- |
+| 1 | Remove protocol-level sessions and the `Mcp-Session-Id` header from Streamable HTTP. List endpoints no longer vary per connection. Cross-call state moves to server-minted handles passed as ordinary tool arguments. | SEP-2567 | Removes the reason `McpStreamableServerSession` exists on the modern path. |
+| 2 | Make MCP stateless: remove the `initialize` / `notifications/initialized` handshake. Every request carries its protocol version and client capabilities in `_meta`. Version mismatch ⇒ `UnsupportedProtocolVersionError`. | SEP-2575 | Deletes `LifecycleInitializer` from the modern path; every request gains a mandatory envelope. |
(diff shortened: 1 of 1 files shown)


slachiewicz:
Thanks for the thorough write-up — this is the first end-to-end design on the table, and promoting `McpStatelessAsyncServer` matches what's actually in the Java SDK code today.

Three flags from a Spring Boot-behind-reverse-proxy deployer's perspective:

1. **D1 dual-era default (your Q2):** we'd strongly prefer one endpoint serving both eras, selected by how the client opens (`initialize` → legacy path; `server/discover` or a request carrying `_meta` + `MCP-Protocol-Version: 2026-07-28` with no session → modern path), with a builder switch to run modern-only. Separate endpoints or a transport-level flag turns this into a deployment change for every existing user behind Tomcat/reverse proxies.

2. **SEP-2567 state:** please keep the document explicit that cross-call state travels as server-minted handles passed as ordinary tool arguments, with no SDK-side store. The open `McpSessionStore` PRs (#914, #1048) would then be clearly scoped as 2.x-only under this design — worth stating so contributors stop building against a model 3.x removes.

3. **Q5 phasing:** ship P1–P2 (extensions #1064, headers SEP-2243, TTL #1062, deprecations #1053, plus any validator gaps on SEP-2106) in a 2.x minor now. It shortens the 3.x critical path and gives clients something concrete to test against while the core team does SEP-2575/2322.

Two checks while reviewing: the `_meta` keys should be the namespaced forms (`io.modelcontextprotocol/protocolVersion`, `io.modelcontextprotocol/clientCapabilities`) per the changelog, and the Tasks section should reflect the final extension shape — polling `tasks/get`, `tasks/update`, unsolicited task handles, no `tasks/list`.

Happy to help with any of the P1–P2 items if maintainers want to open them up.

omarmahamid:
@slachiewicz 


Thanks — all three flags land.

D1/Q2: Agreed — one endpoint, era chosen by how the client opens. The modern-only switch is currently an interface override; I'll move it to a builder option and make the header ↔ _meta check (-32020) explicit in D1.

SEP-2567: Agreed. I'll add a subsection: state travels as server-minted handles in tool arguments, no SDK-side store, and McpSessionStore (#914, #1048) is legacy/2.x-only.

Q5: Yes, ship P1–P2 as a 2.x minor. Happy to take you up on the offer — if you pick up P1–P2 (#1064, #1062, #1053, SEP-2243, SEP-2106), I'll focus on P3–P4. I have a local P1–P2 spike I can share so we don't duplicate.

Until a maintainer confirms the direction and 3.0.0 scope — so let's treat the split as proposed until they weigh in.


Will post the revised sections once updated.

@tzolov , @Kehrlann can you take a look into the propsal ?

Thanks

Kehrlann:
Hey @omarmahamid , thanks for your proposal. Unfortunately, we are not open to design documents at this time. We don't have the bandwidth to review 1600 lines of design.

Issue #990, opened 2026-05-29
SEP-2243: HTTP Header Standardization for Streamable HTTP Transport


caomengxuan666:
## Description

Per [SEP-2243 (HTTP Standard Headers)](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/2243-http-standard-headers.md), MCP clients MUST send the following HTTP headers on POST requests:

- `Mcp-Method`: Set to the JSON-RPC `method` value from the request body (REQUIRED on every POST)
- `Mcp-Name`: Set from `params.name` (for tools/call, prompts/get) or `params.uri` (for resources/read, resources/subscribe, resources/unsubscribe) when present

## Context

The cxxmcp (C++ SDK) conformance testing revealed that the official conformance test suite's reference client (TypeScript SDK) does not send these headers, making the conformance suite self-contradictory. See:
- [typescript-sdk#2176](https://github.com/modelcontextprotocol/typescript-sdk/issues/2176) — TypeScript SDK bug report
- [conformance#323](https://github.com/modelcontextprotocol/conformance/issues/323) — Conformance suite self-contradiction
- [typescript-sdk#2178](https://github.com/modelcontextprotocol/typescript-sdk/pull/2178) — Fix PR for TypeScript SDK

## Impact

Servers that enforce SEP-2243 (reject missing `Mcp-Method` with HTTP 400) are incompatible with clients that don't send these headers. This affects cross-SDK interoperability.

## Request

Please implement `Mcp-Method` and `Mcp-Name` header support per SEP-2243 to ensure cross-SDK compatibility.

nikita-kibitkin:
I can take the Mcp-Method client-side part of this.

#994 already covers Mcp-Name for Streamable HTTP, but Mcp-Method is still not emitted by HttpClientStreamableHttpTransport. I would add the HttpHeaders constant and set Mcp-Method on every Streamable HTTP POST from the JSON-RPC request/notification method, next to the existing MCP-Protocol-Version handling, with tests around initialize and a regular request/notification.

I would keep server-side validation out of this PR and leave it as a follow-up, since #994 already starts touching that area for Mcp-Name and I do not want to mix client emission with server enforcement in the same small change. If #994 merges first, I will rebase on top of it.

Does that split work, or would you prefer Mcp-Method to be folded into #994 instead?

Kehrlann:
[...]

PR #1092, opened 2026-08-16
fix: validate MCP-Protocol-Version header in servlet transports

ez-lbz:
The streamable HTTP spec requires the server to answer with 400 Bad Request when the client sends an unsupported MCP-Protocol-Version header, and to accept the request when the header is missing (falling back to the negotiated version). The Servlet-based providers never read this header.

This change validates the MCP-Protocol-Version header in doGet, doPost, and doDelete of `HttpServletStreamableServerTransportProvider` and in doPost of `HttpServletStatelessServerTransport`, rejecting unsupported versions with a 400 JSON error response consistent with the existing error style.

Tests cover the version-header matrix (supported version accepted, unsupported version rejected, missing header accepted) across both providers.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStatelessServerTransport.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStatelessServerTransport.java
@@ -19,6 +19,7 @@ import io.modelcontextprotocol.json.McpJsonMapper;
 import io.modelcontextprotocol.common.McpTransportContext;
 import io.modelcontextprotocol.server.McpStatelessServerHandler;
 import io.modelcontextprotocol.server.McpTransportContextExtractor;
+import io.modelcontextprotocol.spec.HttpHeaders;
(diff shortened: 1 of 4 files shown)


slachiewicz:
Heads-up on overlap: #1112 implements SEP-2243 end to end for the servlet transports, including the `MCP-Protocol-Version` validation this PR adds, following the ask in #990 to land the SEP as one piece. It exempts `initialize` and returns `INVALID_REQUEST` on an unsupported version. If you would rather carry the protocol-version part yourself, say so and I will drop that slice from #1112.

ez-lbz:
Thanks for the heads-up — understood. If #1112's SEP-2243 implementation covers the `MCP-Protocol-Version` validation for the servlet transports with equivalent rejection behavior, this PR is redundant and I'm happy to close it in favor of landing the SEP as one piece. If you'd rather keep a narrower stopgap until #1112 merges, the diff here is self-contained (servlet transport providers + tests only) — your call either way.

Issue #1085, opened 2026-08-11
Backport server/discover / METHOD_NOT_FOUND fix (-32601) to pre-2.0 lines — HTTP 500 has no remedy without Spring Boot 4 migration


eashwar-mp:
### Summary

The `server/discover` → HTTP 500 regression (unknown JSON-RPC method throws instead of returning `-32601 Method not found`) currently has **no remedy for users on the pre-2.0 release lines**. The known fixes all target 2.0.x:

- #784 — "Stateless server handler throws exception instead of returning JSON-RPC -32601" — closed, milestoned to **2.0.1**.
- #1072 — "[2026-07-28] Java SDK 2.0.0 returns HTTP 500 for OpenAI `server/discover` requests" — open, against **2.0.0**; PR #1083 fixes the 2.0.x stateless transport.

This issue asks for the fix to also be made available to users still on `0.x` / `1.x`.

### Why this matters for pre-2.0 users

OpenAI's hosted MCP connector now sends `server/discover` during its handshake (before `tools/list`). On the pre-2.0 SDK the stateless transport responds:

```
Missing handler for request type: server/discover   → HTTP 500
```

OpenAI surfaces that 500 as **HTTP 424 `external_connector_error`**, which aborts the entire response — even for a plain "hi" that needs no tool call. In practice this takes an MCP server completely offline for the OpenAI Responses API connector the moment the client starts probing, with no application-side cause. The documented backward-compat behavior is to return `-32601` so the client falls back to the legacy `initialize` flow, and that fallback lives in the SDK's request handling, not in application code — so downstream apps cannot fix it without patching or replacing the SDK.

The only forward path today is upgrading to 2.0.x. For Spring AI consumers that is not a version bump: `spring-ai` 2.x requires **Spring Boot 4 / Spring Framework 7**, a framework-wide migration. Concretely, we are on `spring-ai 1.1.4` (MCP Java SDK `0.17.0`) on Spring Boot 3.x, and a trial bump to `spring-ai` 2.0 compiles and passes unit tests but fails to boot (`NoClassDefFoundError: org.springframework.core.Nullness`, a Framework 7 class). So a spec-compatibility regression triggered by a **client-side** change is effectively gating a large, unrelated framework migration.

### The ask

Either of the following would resolve it for pre-2.0 users:

[...]

PR #1083, opened 2026-08-09
fix: return HTTP 404 for METHOD_NOT_FOUND in stateless transport (#1072)

elang2:
## Summary

The 2026-07-28 Streamable HTTP spec requires that an unsupported RPC method return HTTP 404 with JSON-RPC error -32601, not HTTP 500. Right now `HttpServletStatelessServerTransport` returns 200 for everything, including unrecognized methods like `server/discover`. OpenAI's hosted MCP client hits this path and surfaces a 424 `external_connector_error` to users.

This PR adds a `mapJsonRpcErrorToHttpStatus()` helper that maps -32601 to 404. It's a small, isolated change that unblocks production traffic without waiting for full 2026-07-28 lifecycle support tracked in #1011.

## Test plan

- Updated `testMissingHandlerReturnsMethodNotFoundError` to verify HTTP 404 status
- Added `testUnknownMethodReturnsHttp404WithMethodNotFoundError` for `server/discover` specifically
- All 340 mcp-core tests pass
- All 937 mcp-test integration tests pass (BUILD SUCCESS)

Fixes #1072

Issue #1072, opened 2026-07-30
[2026-07-28] Java SDK 2.0.0 returns HTTP 500 for OpenAI server/discover requests


morad3741:
Starting on 2026-07-30 at approximately 07:10 UTC, we observed OpenAI's hosted MCP client begin sending `server/discover` before `tools/list` when using a remote MCP server through the
  Responses API.

  Our server uses the latest released MCP Java SDK, `2.0.0`, which supports protocol revision `2025-11-25` but does not currently implement `server/discover`.

  When the request reaches `HttpServletStatelessServerTransport`, the SDK returns HTTP 500 with an error similar to:

  `Missing handler for request type: server/discover`

  OpenAI then returns HTTP 424 `external_connector_error` from the Responses API. This prevents the entire response from being generated, even when the user sends a message such as "hi" that
  does not require an MCP tool call.

  We understand that support for the new stateless lifecycle is already tracked in:

  - https://github.com/modelcontextprotocol/java-sdk/issues/1011

  We also found that the general unknown-handler behavior was fixed by:

  - https://github.com/modelcontextprotocol/java-sdk/pull/800

  However, that change was merged after the latest `2.0.0` release and does not appear to be available in a published version.

  Could you please clarify:

  1. Which Java SDK version is expected to support the `2026-07-28` protocol revision?
  2. Is there an estimated release timeline?
  3. Will there be an interim `2.x` release containing the unknown-handler fix from #800?
  4. Until full support is available, what is the recommended server behavior for `server/discover` so modern clients can safely fall back to the legacy `initialize` flow?

  The `2026-07-28` Streamable HTTP specification requires an unsupported RPC to return HTTP 404 with JSON-RPC error `-32601`, rather than HTTP 500:

  https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http

  **Environment**

  - MCP Java SDK: `2.0.0`
  - Maven artifacts:
    - `io.modelcontextprotocol.sdk:mcp-core`
    - `io.modelcontextprotocol.sdk:mcp-json-jackson2`
  - Server transport: `HttpServletStatelessServerTransport`
  - Server type: `McpStatelessSyncServer`
  - Java: Amazon Corretto 25
  - Spring Boot: `3.5.16`
  - Transport: Streamable HTTP
  - Deployment: Stateless Kubernetes service
  - MCP capabilities: Tools only
  - Vector store: Not applicable
[...]

PR #800, opened 2026-02-16
Fix error behavior on unregistered handlers in stateless server handler

maff:
When a client requests a resource without a registered handler despite a stateless server not advertising it, the server now returns with a JSON-RPC method not found error (-32601), aligned with the stateful implementation.

## Motivation and Context

Closes issue #784

## How Has This Been Tested?

Integration test covering unregistered handler response with stateless transport.

## Breaking Changes

Behavioral change to return a proper JSON-RPC error instead of a HTTP 500 response. Breaking only in terms of changed behavior in case of an error.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #1104, opened 2026-08-20
Backport #800: fix error behavior on unregistered handlers in stateless server handler

olsavmic:
Backport of #800 to the 0.18.x line. See #1085.

## Why

`DefaultMcpStatelessServerHandler.handleRequest` returns `Mono.error(...)` for a method with no registered handler. That is an early return, so the error never reaches the `onErrorResume` a few lines below it that maps errors onto a JSON-RPC error response — it escapes to the transport instead, and every stateless transport maps an escaping handler error to **HTTP 500**:

- `HttpServletStatelessServerTransport`
- `WebMvcStatelessServerTransport` (`io.modelcontextprotocol.sdk:mcp-spring-webmvc`)
- `WebFluxStatelessServerTransport` (`io.modelcontextprotocol.sdk:mcp-spring-webflux`)

Requests for *registered* methods are unaffected — `tools/call` with an unknown tool name already returns a proper `-32602`, because that path does run through `onErrorResume`.

The practical impact is the one described in #1072: OpenAI's hosted MCP connector sends `server/discover` before `tools/list`, so a healthy server answers a routine capability probe with a 5xx. On our side that made `POST /mcp` the only 5xx source in the service and burned the availability SLO budget; OpenAI surfaces the same 500 as HTTP 424 `external_connector_error`, which aborts the whole response.

## Why a backport

The fix is on `main` and shipped in `mcp-core` 2.0.1, but 2.0.x is not reachable for Spring AI users on Spring Boot 3:

- `io.modelcontextprotocol.sdk:mcp-spring-webmvc` does not exist at 2.0.x — the Spring transports moved to `org.springframework.ai` — so `mcp-core` cannot be bumped on its own.
- `spring-ai` 2.0's `spring-ai-starter-mcp-server-webmvc` pulls `spring-boot-starter-web:4.1.0`, making the upgrade a Spring Framework 7 migration rather than a version bump.
- `spring-ai` 1.1.x pins the 0.18.x line, and 0.18.4 (the latest published) still has the early return.

That is the situation #1085 describes, with two other reporters confirming it on 0.17.0 and 0.18.3.

## Change

Three lines in `DefaultMcpStatelessServerHandler`, identical in effect to #800: return a `JSONRPCResponse` carrying `-32601 Method not found: <method>` instead of failing the `Mono`.

## Tests

- `DefaultMcpStatelessServerHandlerTests` — the unit test from #800, verbatim.
- `HttpServletStatelessIntegrationTests#testMissingHandlerReturnsMethodNotFoundError` — asserts the **HTTP status is 200** as well as the JSON-RPC error body. #800's integration test goes through an MCP client and so only sees the JSON-RPC layer; since the regression people actually hit is the status code, this one drives the transport directly with `MockHttpServletRequest` (the same style as the neighbouring `testThrownMcpErrorAndJsonRpcError`) so a future transport-level regression is caught too.

`./mvnw -pl mcp-core test -Dtest=DefaultMcpStatelessServerHandlerTests` → 1/1 green.
`./mvnw -pl mcp-test test -Dtest=HttpServletStatelessIntegrationTests` → 13/13 green.

## Note on HTTP status

This restores parity with the stateful session handler and with 2.0.x: unknown method → **HTTP 200** with a JSON-RPC `-32601` body. The 2026-07-28 Streamable HTTP revision asks for HTTP 404 with `-32601`, which is what #1083 adds for `HttpServletStatelessServerTransport` on 2.0.x. That status change is deliberately left out here — it is a separate concern from the 500, the Spring transports each carry their own copy of the status logic and would need the same treatment, and 200 + `-32601` is already enough for clients to fall back to the legacy `initialize` flow.

Happy to open the equivalent PRs against `1.0.x` and `1.1.x` — both carry the same early return (via `McpError.builder(...)`), and I have the same change prepared and green on both.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/DefaultMcpStatelessServerHandler.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/DefaultMcpStatelessServerHandler.java
@@ -32,7 +32,9 @@ class DefaultMcpStatelessServerHandler implements McpStatelessServerHandler {
 			McpSchema.JSONRPCRequest request) {
 		McpStatelessRequestHandler<?> requestHandler = this.requestHandlers.get(request.method());
 		if (requestHandler == null) {
-			return Mono.error(new McpError("Missing handler for request type: " + request.method()));
+			return Mono.just(new McpSchema.JSONRPCResponse(McpSchema.JSONRPC_VERSION, request.id(), null,
+					new McpSchema.JSONRPCResponse.JSONRPCError(McpSchema.ErrorCodes.METHOD_NOT_FOUND,
+							"Method not found: " + request.method(), null)));
 		}
 		return requestHandler.handle(transportContext, request.params())
 			.map(result -> new McpSchema.JSONRPCResponse(McpSchema.JSONRPC_VERSION, request.id(), result, null))
--- a/mcp-core/src/test/java/io/modelcontextprotocol/server/DefaultMcpStatelessServerHandlerTests.java
+++ b/mcp-core/src/test/java/io/modelcontextprotocol/server/DefaultMcpStatelessServerHandlerTests.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2026-2026 the original author or authors.
+ */
+
+package io.modelcontextprotocol.server;
+
+import io.modelcontextprotocol.common.McpTransportContext;
+import io.modelcontextprotocol.spec.McpSchema;
+import org.junit.jupiter.api.Test;
+import reactor.test.StepVerifier;
+
+import java.util.Collections;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class DefaultMcpStatelessServerHandlerTests {
+
+	@Test
+	void testHandleRequestWithUnregisteredMethod() {
+		// no request/initialization handlers
+		DefaultMcpStatelessServerHandler handler = new DefaultMcpStatelessServerHandler(Collections.emptyMap(),
+				Collections.emptyMap());
+
+		// unregistered method
+		McpSchema.JSONRPCRequest request = new McpSchema.JSONRPCRequest(McpSchema.JSONRPC_VERSION, "resources/list",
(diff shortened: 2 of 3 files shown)


eashwar-mp (approved):
Confirmed the underlying bug from the reporter side in #1085 (0.17.0 via spring-ai 1.1.4, WebMvc stateless transport). I also verified this exact change applies byte-for-byte to 0.17.x: the early-return and all `McpSchema` APIs it touches are identical there.

The fix reads correctly, it restores parity with the stateful handler (unknown method returns HTTP 200 with a `-32601` body) and unblocks the OpenAI connector's `server/discover` probe. Leaving the HTTP 404 status change to #1083 as a separate concern is the right call for a backport. LGTM.

JHTosas:
@eashwar-mp It seems the MR is missing workflows being approved, to then merge, and it seems only Maintainers can do that. Any chance you're one, or know who we can ping? This is a much needed change for my team. Thanks!

eashwar-mp:
@JHTosas I'm not a maintainer, my review is a community approval with read-only permissions, so it won't satisfy the merge gate or approve the CI workflows. This needs someone with write access.

Looking at who merges here, @Kehrlann is the most active maintainer (nearly all recent merges are his), and @tzolov triaged this one on #1085 (labeled it P1 and tagged it for confirmation). Either is probably a good ping to approve the workflows and take it from there. The bug is now confirmed on three lines (0.17.0, 0.18.3, 0.18.4) and the change is three lines plus tests, so hopefully a quick one.

cc: @Kehrlann @tzolov FYI


-----------------------
@JHTosas If it's breaking your prod today or this week, I would recommend adding a servlet filter which intercepts only a server/discover request (carrying an id) and returns a clean -32601 at HTTP 200. That should be a good intermediary fix, while the maintainers push this one out and publish the artifact.

Kehrlann:
Thank you for the backport.

The 0.17.0 line is completely out of support, we won't be backporting this further.

PR #1106, opened 2026-08-22
fix: reject duplicate streamable HTTP initialize requests

### LINKED ISSUE #962: Duplicate initialize is accepted after initialization and returns a new session ID
**Bug description**

After a normal Streamable HTTP initialization flow, the Java SDK server accepts later `initialize` requests instead of rejecting them as duplicate initialization for an already-initialized session. The duplicate request returns HTTP 200 and a normal `InitializeResult`; when the duplicate request uses an older protocol version, that older version is echoed in the result.

The behavior I observed is not a same-session overwrite in the Streamable HTTP transport. Instead, it behaves like fresh session creation: each duplicate `initialize` creates a new `McpStreamableServerSession` and returns a new `Mcp-Session-Id`, even though the client already has an active session.

From a spec perspective, the lifecycle text says initialization is the first interaction and that the operation phase should use the negotiated protocol version and capabilities. The Streamable HTTP transport also defines a session as logically related interactions beginning with initialization, and says a client should start a new session after receiving HTTP 404 for an existing session ID. I do not see text that explicitly defines how a same-endpoint duplicate `initialize` should be handled, so I am filing this as a lifecycle/session-boundary ambiguity rather than claiming that the spec explicitly requires rejecting duplicate `initialize`.

I checked the Java SDK implementation path directly:

- `HttpServletStreamableServerTransportProvider` handles any JSON-RPC request whose method is `initialize` by calling `sessionFactory.startSession(initializeRequest)` and then storing the returned session under `init.session().getId()`.
- `DefaultMcpStreamableServerSessionFactory#startSession` creates a new UUID and constructs a new `McpStreamableServerSession` from that `InitializeRequest`.
- `McpStreamableServerSession` stores `clientCapabilities` and `clientInfo` from that request, and later creates `McpAsyncServerExchange` objects using those values.

So the observed new session IDs are consistent with the current implementation path.

**Environment**

- Stable release: `v1.1.2` (`e9e1a2f34dedb72008d90e9919052d46eb2b701c`)
- `main` snapshot: from 2026-05-15 (`c09ee67f60260bd258b1a1aab9315a647a239d86`)
- Transport: Streamable HTTP server, stateful mode
- Java runtime used for repro: OpenJDK `21.0.10`
- SDK build target: Java 17 (`java.version`, `maven.compiler.source`, and `maven.compiler.target` are set to `17` in the SDK `pom.xml`)

**Steps to reproduce**

1. Start a Java SDK Streamable HTTP server.
2. Send an initial `initialize` request and save the returned `Mcp-Session-Id`.
3. Send `notifications/initialized` with that session ID.
4. Send a normal request such as `ping` with that session ID and observe a successful response.
5. Send a second `initialize` request after the server is already operational. Include the existing `Mcp-Session-Id` header to make clear that this is not a client intentionally starting a fresh session without a session ID.
6. Repeat with changed `clientInfo`, changed `capabilities`, or an older `protocolVersion` such as `2024-11-05`.
7. Observe that each later `initialize` returns HTTP 200, a normal `InitializeResult`, and a new `Mcp-Session-Id`.

In my logs, both the stable release and the 2026-05-15 `main` snapshot showed:

```text
Second initialize after normal initialization: HTTP 200, InitializeResult returned
Duplicate initialize with changed capabilities/clientInfo: HTTP 200, InitializeResult returned
Duplicate initialize with protocolVersion 2024-11-05: HTTP 200, InitializeResult returned with protocolVersion 2024-11-05
Five repeated later initialize requests: all returned HTTP 200
A later ping still succeeded
````

**Expected behavior**

If a client already has an active Streamable HTTP session, I would expect a later `initialize` sent with that active `Mcp-Session-Id` to be rejected, for example with a JSON-RPC `Invalid Request` error or another clear 4xx/JSON-RPC error.

If this behavior is intentional, I would expect it to be documented explicitly, including whether servers should ignore an existing `Mcp-Session-Id` on an `initialize` request and issue a new session ID.

**Minimal reproducible example**

Set `ENDPOINT` to a Java SDK Streamable HTTP endpoint:

```bash
ENDPOINT=http://127.0.0.1:8080/mcp
```

Initialize and copy the returned `Mcp-Session-Id` header into `SID`:

```bash
curl -i -sS --http1.1 -X POST "$ENDPOINT" \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -H 'MCP-Protocol-Version: 2025-11-25' \
  --data '{"jsonrpc":"2.0","id":"init-1","method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{"roots":{"listChanged":true},"sampling":{}},"clientInfo":{"name":"repro-initial","version":"1.0.0"}}}'
```

Send the initialized notification:

```bash
curl -i -sS --http1.1 -X POST "$ENDPOINT" \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -H 'MCP-Protocol-Version: 2025-11-25' \
  -H "Mcp-Session-Id: $SID" \
  --data '{"jsonrpc":"2.0","method":"notifications/initialized","params":{}}'
```

Confirm that the session is operational:

```bash
curl -i -sS --http1.1 -X POST "$ENDPOINT" \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -H 'MCP-Protocol-Version: 2025-11-25' \
  -H "Mcp-Session-Id: $SID" \
  --data '{"jsonrpc":"2.0","id":"ping-1","method":"ping","params":{}}'
```

Now send a second `initialize` with changed parameters on the already-operational session:

```bash
curl -i -sS --http1.1 -X POST "$ENDPOINT" \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -H 'MCP-Protocol-Version: 2025-11-25' \
  -H "Mcp-Session-Id: $SID" \
  --data '{"jsonrpc":"2.0","id":"init-2","method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{"roots":{"listChanged":false}},"clientInfo":{"name":"repro-duplicate","version":"2.0.0"}}}'
```

Observed result:

```text
HTTP/1.1 200 OK
Mcp-Session-Id: <new-session-id>

{"jsonrpc":"2.0","id":"init-2","result":{"protocolVersion":"2024-11-05", ...}}
```

I observed that the new Streamable HTTP session's `clientInfo` and `clientCapabilities` reflected the duplicate request. I did not observe a stored negotiated-protocol-version field in the Streamable HTTP session object. My concern here is the new session boundary created by a duplicate `initialize` after the client already had an active session.

steven-ji:
Reject duplicate Streamable HTTP `initialize` requests that carry an active
`Mcp-Session-Id`, while preserving fresh initialization without a session ID.

The rejection returns HTTP 400 with a JSON-RPC `INVALID_REQUEST` response that
retains the original request ID. It does not invoke the session factory or replace
the active session.

## Motivation and Context

The servlet Streamable HTTP transport currently handles `initialize` before checking
the supplied session ID. A client that already has an active session can therefore
send another `initialize` request and receive a new session instead of a clear
rejection.

This keeps the change at the transport boundary: only duplicate initialization for
an active session is rejected. The normal initialization path remains unchanged.

Fixes #962

## How Has This Been Tested?

The new regression test was first run against the current `main` implementation and
failed because the duplicate request returned HTTP 200. It passes after the fix.

- `./mvnw -pl mcp-core -Dtest=HttpServletStreamableServerDuplicateInitializeTests test`
  - 1 test passed
  - 0 failures
  - 0 errors
- `./mvnw -pl mcp-core clean test`
  - 386 tests passed
  - 0 failures
  - 0 errors
- `./mvnw -pl mcp-test -am -Dtest=HttpServletStreamableIntegrationTests -Dsurefire.failIfNoSpecifiedTests=false test`
  - 43 tests passed
  - 0 failures
  - 0 errors
- `git diff --check origin/main...HEAD`

No successful full-reactor test result is claimed.

## Breaking Changes

None. Fresh initialization without an active session ID retains its existing
behavior.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

- The earlier PR #969 is closed; this implementation was rebuilt and verified from
  the current `main` branch.
- `disclosure.txt` is included as required by the repository contribution policy.

### DIFF
--- a/disclosure.txt
+++ b/disclosure.txt
@@ -1,0 +1,1 @@
+This change was submitted despite me reading the rules and understanding AI contribution guidelines.
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
@@ -449,6 +449,14 @@ public class HttpServletStreamableServerTransportProvider extends HttpServlet
 					return;
 				}
 
+				String sessionId = request.getHeader(HttpHeaders.MCP_SESSION_ID);
+				if (sessionId != null && this.sessions.containsKey(sessionId)) {
+					this.responseJsonRpcError(response, HttpServletResponse.SC_BAD_REQUEST, jsonrpcRequest.id(),
+							McpSchema.ErrorCodes.INVALID_REQUEST,
+							"Duplicate initialize request for active session: " + sessionId);
+					return;
+				}
+
 				McpSchema.InitializeRequest initializeRequest = jsonMapper.convertValue(jsonrpcRequest.params(),
 						new TypeRef<McpSchema.InitializeRequest>() {
 						});
@@ -568,6 +576,18 @@ public class HttpServletStreamableServerTransportProvider extends HttpServlet
 				response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Error processing message");
 			}
 		}
+	}
+
+	private void responseJsonRpcError(HttpServletResponse response, int httpCode, Object requestId, int errorCode,
+			String message) throws IOException {
+		response.setContentType(APPLICATION_JSON);
+		response.setCharacterEncoding(UTF_8);
+		response.setStatus(httpCode);
+		McpSchema.JSONRPCResponse errorResponse = McpSchema.JSONRPCResponse.error(requestId,
+				new McpSchema.JSONRPCResponse.JSONRPCError(errorCode, message));
+		PrintWriter writer = response.getWriter();
+		writer.write(this.jsonMapper.writeValueAsString(errorResponse));
+		writer.flush();
 	}
 
 	/**
--- a/mcp-core/src/test/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerDuplicateInitializeTests.java
+++ b/mcp-core/src/test/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerDuplicateInitializeTests.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ */
+
+package io.modelcontextprotocol.server.transport;
+
+import java.io.ByteArrayInputStream;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import jakarta.servlet.ReadListener;
+import jakarta.servlet.ServletInputStream;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import org.junit.jupiter.api.Test;
+import reactor.core.publisher.Mono;
+
+import io.modelcontextprotocol.json.McpJsonMapper;
+import io.modelcontextprotocol.spec.HttpHeaders;
+import io.modelcontextprotocol.spec.McpSchema;
+import io.modelcontextprotocol.spec.McpStreamableServerSession;
+import io.modelcontextprotocol.spec.json.gson.GsonMcpJsonMapper;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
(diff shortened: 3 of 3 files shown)


steven-ji:
Thanks for the clarification on #962. I understand that when an `initialize` request includes an existing session ID, the server ignores that header and creates a separate new session, so it does not override the existing session or cause the session leak described in the issue.

Given that, this change is not needed, so I am closing this PR. Thanks for the explanation.

Issue #746, opened 2026-01-13
Support ToolsRepository for Dynamic Tool Management & Cursor-based Pagination Status


ultramancode:
# Support ToolsRepository for Dynamic Tool Management & Cursor-based Pagination

## Related Issue
#740
#578  

## Motivation

The current MCP server implementation exposes every registered tool to all connected clients without exception. In a production environment, this "all-or-nothing" approach presents several challenges:

- **Multi-tenant isolation**: The ability to serve different toolsets to different clients or environments.
- **Granular Access Control**: Protecting administrative or sensitive tools by hiding them from unauthorized users.
- **Dynamic Flexibility**: Loading tool definitions on-the-fly from external sources (Databases, Config Servers, etc.) rather than being limited to static registration.
- **Enterprise Scalability**: Efficiently managing large-scale toolsets without overwhelming the client or the network, ensuring resource efficiency through Cursor-based Pagination.
## Problem

The current architecture of `McpAsyncServer` is rigid:
- Tools are stored in a static `ConcurrentHashMap<String, AsyncToolSpecification>`.
- The `tools/list` request returns the entire map, regardless of the client's identity or needs.
- There is no way to access the client's context (e.g., client ID or role) during the tool discovery process.

## Proposed Solution

We introduce the `ToolsRepository` interface to decouple tool storage from the server engine, providing a pluggable architecture for tool management:

```java
public interface ToolsRepository {
    // Context-aware filtering and pagination support
    Mono<ToolsListResult> listTools(McpAsyncServerExchange exchange, String cursor);
    
    // Security validation during execution
    Mono<AsyncToolSpecification> resolveToolForCall(String name, McpAsyncServerExchange exchange);
    
    void addTool(AsyncToolSpecification tool);
    void removeTool(String name);
}
```

**Key Advantages**:
- **Exchange Integration**: Access client metadata and attributes via `McpAsyncServerExchange`.
- **Pre-filtering**: Prevent tool exposure at the listing stage, reducing the client-side attack surface.
- **Spec-aligned Pagination**: Support for opaque cursors, following the MCP specification.

## Practical Applications

1. **Role-Based Visibility (RBAC)**: Automatically hide admin tools from standard users by checking roles in the exchange.
2. **Tenant-Specific Ecosystems**: Dynamically provide a unique set of tools based on the tenant ID associated with the connection.
[...]

Issue #608, opened 2025-10-08
Support filtering tools depends on request data (http-headers, etc)


linarkou:
Please do a quick search on GitHub issues first, the feature you are about to request might have already been requested.

**Expected Behavior**

There is no way to filter available tools in [`toolsListRequestHandler()`](https://github.com/modelcontextprotocol/java-sdk/blob/19a8c00a475d38c6104887d9df1ab503b4a51839/mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java#L383) and [`toolsCallRequestHandler()`](https://github.com/modelcontextprotocol/java-sdk/blob/19a8c00a475d38c6104887d9df1ab503b4a51839/mcp-core/src/main/java/io/modelcontextprotocol/server/McpStatelessAsyncServer.java#L392)
I would like to filter tools to specific clients depending on some data in http-headers.

**Current Behavior**

There is no way to filter tools dynamically. The only way to do it - inherit `McpAsyncServer`, and use same package `io.modelcontextprotocol.server` - because `McpServerFeatures.Async` used in constructor is package-private. So maybe also make `McpServerFeatures.Async` public?

**Context**

I would like to implement authorization to be able to restrict tool calls to specific clients.

bcersows:
Did you have any progress on implementing this yourself? I'm also striving for the same aim.

yellowdog01:
I think this is a relatively important feature, but currently the SDK does not support it.

chemicL:
Let's tackle this in #578 if that is ok for you.

Issue #1073, opened 2026-08-01
Allow composing request customizers on the HTTP client transport builders


iuliiasobolevska:
**Expected Behavior**

A caller should be able to add a request customizer to a transport builder without discarding the ones already registered on it.

Two shapes would work. The smaller one is a getter, so a caller can compose by hand:

```java
var existing = builder.getAsyncHttpRequestCustomizer();
builder.asyncHttpRequestCustomizer(new DelegatingMcpAsyncHttpClientRequestCustomizer(List.of(existing, mine)));
```

The better one is an additive setter alongside the existing replace-all one:

```java
public Builder addAsyncHttpRequestCustomizer(McpAsyncHttpClientRequestCustomizer customizer) {
    Assert.notNull(customizer, "customizer must not be null");
    this.httpRequestCustomizers.add(customizer);
    return this;
}
```

with `build()` collapsing the list through `DelegatingMcpAsyncHttpClientRequestCustomizer`, plus the sync twin for `McpSyncHttpClientRequestCustomizer`.

Either would apply to both `HttpClientStreamableHttpTransport.Builder` and `HttpClientSseClientTransport.Builder`.

Making the existing setter additive would be the cleanest API, but it would change behavior for anyone who calls it twice today and expects a replacement, so it probably belongs in a major version.

**Current Behavior**

Both builders hold exactly one customizer, and the setter assigns it. On `main` at `fd00498`:

```java
// HttpClientStreamableHttpTransport
private final McpAsyncHttpClientRequestCustomizer httpRequestCustomizer;                                        // 129
private McpAsyncHttpClientRequestCustomizer httpRequestCustomizer = McpAsyncHttpClientRequestCustomizer.NOOP;   // 721
this.httpRequestCustomizer = asyncHttpRequestCustomizer;                                                        // 852
```

```java
// HttpClientSseClientTransport
private final McpAsyncHttpClientRequestCustomizer httpRequestCustomizer;                                        // 124
private McpAsyncHttpClientRequestCustomizer httpRequestCustomizer = McpAsyncHttpClientRequestCustomizer.NOOP;   // 192
this.httpRequestCustomizer = asyncHttpRequestCustomizer;                                                        // 301
```

The sync overload routes through `McpAsyncHttpClientRequestCustomizer.fromSync(...)` into the same field, so a sync customizer and an async one overwrite each other as well.

[...]

PR #1108, opened 2026-08-26
Add filtering for MCP tool list

### LINKED ISSUE #525: Add a way to customize default request handlers for Stateless server
**Expected Behavior**

Currently using the Stateless server. We need a way to override the behavior of a single request handler, for instance the `tools/list` request to provide custom tool descriptions.

**Current Behavior**

The `McpStatelessAsyncServer` constructor currently creates a `DefaultMcpStatelessServerHandler` as mcpHandler for the ServerTransport with private methods as handlers with private variables.
We can currently override the mcpHandler after-the-fact using `McpStatelessServerTransport::setMcpHandler`, but it would be nice to override only a single handler, for example adding a couple of getters:
```java

var serverTransport = WebMvcStatelessServerTransport.builder() [...] .build();
...
var server = McpServer.sync(serverTransport).tools(tools) [...] .build();

// Override default request handler
serverTransport.getMcpHandler()
               .getRequestHandlers()
               .put(McpSchema.METHOD_TOOLS_LIST, myCustomToolsList(tools));
```

What do you think?

**Context**

We want customizable tool descriptions per-user with minimal code duplication of the SDK.

### LINKED ISSUE #593: Support for accessing and customizing request handlers in McpServerSession
Expected Behavior

  As a developer extending the MCP Java SDK, I should be able to access and customize the request handlers after the
   McpServerSession is initialized. This would enable dynamic replacement of default handlers with custom
  implementations.

  // Example use case: Replacing default handlers with custom implementations
  McpServerSession session = // ... initialized session
  Map<String, McpRequestHandler<?>> handlers = session.getRequestHandlers();
  handlers.put(McpSchema.METHOD_TOOLS_LIST, new MyCustomListToolsRequestHandler());
  handlers.put(McpSchema.METHOD_TOOLS_CALL, new MyCustomToolsCallRequestHandler());

  Current Behavior

  Currently, the McpServerSession class does not provide any way to access or modify the requestHandlers map after
  initialization. The request handlers are set in the constructor and remain private with no getter method.

  Without a proper getter method, developers are forced to use reflection-based workarounds. For example, using
  Spring's ReflectionUtils:

  // Workaround: Using reflection to access private requestHandlers field
  Field field = ReflectionUtils.findField(McpServerSession.class, "requestHandlers");
  ReflectionUtils.makeAccessible(field);
  @SuppressWarnings("unchecked")
  Map<String, McpRequestHandler<?>> handlers = (Map<String, McpRequestHandler<?>>) ReflectionUtils.getField(field,
  session);

  // Replace with custom handlers
  handlers.put(McpSchema.METHOD_TOOLS_LIST, new MyCustomListToolsRequestHandler());
  handlers.put(McpSchema.METHOD_TOOLS_CALL, new MyCustomToolsCallRequestHandler());

  Or to replace the entire map:

  // Workaround: Using reflection to replace the requestHandlers field
  Field field = ReflectionUtils.findField(McpServerSession.class, "requestHandlers");
  ReflectionUtils.makeAccessible(field);

  Map<String, McpRequestHandler<?>> customHandlers = new ConcurrentHashMap<>();
  customHandlers.put(McpSchema.METHOD_TOOLS_LIST, new MyCustomListToolsRequestHandler());
  customHandlers.put(McpSchema.METHOD_TOOLS_CALL, new MyCustomToolsCallRequestHandler());
  // ... add other default handlers

  ReflectionUtils.setField(field, session, customHandlers);

  This approach is fragile, breaks encapsulation, and makes the code harder to maintain.

  Context

  How has this issue affected you?
  When building custom MCP server implementations, I need to replace default tool handlers with custom logic.
  Currently, I'm forced to use reflection hacks to access and modify the private requestHandlers field, which is not
   a recommended practice.

  What are you trying to accomplish?
  I want to implement custom McpRequestHandler instances for standard MCP methods (like tools/list and tools/call)
  and register them with an existing McpServerSession without using reflection.

  My proposed solution:
  1. Make the requestHandlers field a ConcurrentHashMap (for thread-safety during modifications)
  2. Add a public getter getRequestHandlers() to expose the handlers map

  This allows for safe runtime modification while maintaining backward compatibility and eliminating the need for
  reflection-based workarounds.

  What other alternatives have you considered?
  - Creating a wrapper class around McpServerSession - not feasible due to private field access
  - Rebuilding the session with new handlers - inefficient and loses session state
  - Using reflection (current workaround) - fragile and breaks encapsulation

  Are you aware of any workarounds?
  Yes, using Spring's ReflectionUtils to access and modify the private field (as shown above), but this is not a
  proper solution.

### LINKED ISSUE #997: Streamable HTTP: support per-request tools/list filtering hooks for large tool catalogs
## Summary
When servers expose a large number of tools, clients often need a filtered `tools/list` response based on request-scoped security and policy context (for example tenant, user permissions, and metadata headers).

In streamable HTTP mode, we need a supported way to customize `tools/list` results per request while still using the SDK-managed transport/session pipeline.

## Current limitation
Built-in `tools/list` handling does not provide a clear per-request customization hook for filtering the returned tool set.

As a result, applications may need to bypass SDK handling and implement manual interception/response logic.

## Why this matters
- Security: hide tools the caller is not authorized to see.
- Scalability/UX: reduce very large tool catalogs to relevant subsets.
- Multi-tenant correctness: tenant/user-specific visibility.
- Consistency: keep all methods on the same streamable transport path.

## Requested capability
Please add one of the following supported extension points:
1. A pluggable handler override for built-in `tools/list`.
2. Middleware/interceptor API that can transform `tools/list` results per request.
3. A dedicated `tools/list` filter callback with access to request/exchange context and full tool list.

## Expected outcome
Applications can apply per-request security/policy filtering for `tools/list` without bypassing streamable transport behavior.

## Compatibility
No breaking change required. Default behavior should remain unchanged if no custom filter/handler is registered.

## Related issues
- https://github.com/modelcontextprotocol/java-sdk/issues/525
- https://github.com/modelcontextprotocol/java-sdk/issues/593
- https://github.com/modelcontextprotocol/java-sdk/issues/720
- https://github.com/modelcontextprotocol/java-sdk/issues/412

Kehrlann:
## Context

The SDK lacks capabilities for dynamically returning tools, resources, prompts, etc (see #578).

A full implementation in the current state (2.x, 2025-11-25) spec would be complex, as it'd need to target [sync | async] x [stateful | stateless] = 4 variants. We'd need to address the bridges between sync and async, make it coexist with current server's tool management (array list of tools, `.addTool` method).

We are currently focusing on the implementation of the 2026-07-28 spec, which will bring an entirely new, breaking API. It would be a perfect opportunity to bring in repositories as described in #578 .

In the meantime, we recognize users have been asking for capabilities for a long time. The majority of the asks are centered around filtering the list of tools (#997 , #525, examples in #593, #746, #608, etc). While we recognize there are other asks (dynamic tool generation, other primitives than tools), we think they'd be better addressed by a full #578 implementation. In the meantime, we decided exposing filter capabilities for tools, in this PR.

## Scope

This PR is focused on tool list filtering only, while opening the door for filtering other resources if necessary. It addresses filtering in every variant mentioned above. 

⚠️⚠️  The filtering only applies to tools listing. It does **NOT** apply to tool calling, as this can already be achieved in the tool handler. Similarly, it does **NOT** apply to `notifications/tools/list_changed` notifications, as these notifications are session-scoped, and not request-scoped. It is recommended that you DO NOT use notifications when using filters.

Closes #997 
Closes #525 (with caveats, the main discussion is in #578)
Closes #593 (with caveats, the main discussion is in #578)

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -441,6 +441,88 @@ var syncToolSpecification = SyncToolSpecification.builder()
 
 `ImageContent.builder(data, mimeType)` and `AudioContent.builder(data, mimeType)` both take base64-encoded binary data. `EmbeddedResource.builder(resourceContents)` wraps either a `TextResourceContents` (for text data) or a `BlobResourceContents` (for base64-encoded binary data) — see [Reading Binary Resources](#reading-binary-resources) for the `BlobResourceContents` shape.
 
+### Filtering the Tool Listing per Request
+
+By default every registered tool is advertised to every caller. Over an HTTP transport you can
+vary the `tools/list` response per request — to hide tools the caller is not authorized to see,
+or to trim a large catalog down to a relevant subset — by registering one or more tool filters.
+
+The filter receives the `McpTransportContext` extracted from the current request, so it can key
+on HTTP headers, a token, a resolved principal, or anything else your
+`contextExtractor` puts there.
+
+=== "Sync"
+
+    ```java
+    McpServer.sync(transportProvider)
+        .tools(publicTool, adminTool)
+        .addToolFilter((transportContext, tool) ->
+            !tool.name().startsWith("admin-") || isAdmin(transportContext))
+        .build();
+    ```
+
+=== "Async"
+
+    ```java
+    McpServer.async(transportProvider)
+        .tools(publicTool, adminTool)
+        .addToolFilter((transportContext, tool) -> {
+            if (!tool.name().startsWith("admin-")) {
+                return Mono.just(true);
+            }
+            return isAdmin(transportContext); // Mono<Boolean>
+        })
+        .build();
+    ```
+
+The same `addToolFilter(...)` method is available on the stateless builders.
+
+!!! warning "Hiding a tool does not make it unreachable"
+
+    The filter controls **advertisement only**. A hidden tool called by name still executes:
+    you MUST enforce permissions in the tool's call handler. Use the filter to control what a
+    caller is told about, not what they are allowed to do.
+
+**Evaluation semantics**
+
+- The filter is consulted on **every** listing request and never cached, so the same session may
+  legitimately see different results for two successive requests carrying different credentials.
+- Registration order is preserved; only omissions happen.
+- Returning `Mono.empty()` from an async filter omits the tool. An error fails the whole listing
+  request rather than silently hiding tools.
+- Filters accumulate as a boolean **AND**: a tool is listed only when every registered filter accepts it, so a
+  later `addToolFilter(...)` can never widen access. Evaluation follows registration order and
+  short-circuits on the first filter that hides a tool.
+- `toolFilters(Consumer<List<...>>)` hands you the list of filters registered so far, so you can
+  inspect, reorder or clear them before building — useful when filters come from several places:
+
+    ```java
+    McpServer.sync(transportProvider)
+        .addToolFilter(tenantFilter)
+        .toolFilters(filters -> filters.add(0, cheapDenyAllForAnonymousFilter))
+        .build();
+    ```
+
+- Tools are tested one at a time, so a filter that performs I/O per tool costs one round trip per
+  tool. Resolve per-request state **once** in your `contextExtractor` and read it in the filter:
+
+    ```java
+    // one authorization lookup, shared by every tool tested in this request
+    .contextExtractor(request -> McpTransportContext.create(
+            Map.of("perms", introspect(request.getHeader("Authorization")))))
+
+    .addToolFilter((context, tool) -> 
+            ((Set<String>) context.get("perms")).contains(tool.name())
+    )
+    ```
+
+- `notifications/tools/list_changed` is **not** filtered. It is a server-initiated broadcast with
+  no request in flight, so there is no context to evaluate. A client may be told something changed
+  when its own visible set did not; it gets the correct view on its next `tools/list`. Consider disabling
+  this notification entirely when using tool filters.
+- With STDIO there is no per-request metadata, so the filter receives `McpTransportContext.EMPTY` and has nothing to key
+  on.
+
 ### Resource Specification
 
 Specification of a resource with its handler function.
--- a/docs/server.md
+++ b/docs/server.md
@@ -489,7 +489,11 @@ The same `addToolFilter(...)` method is available on the stateless builders.
   legitimately see different results for two successive requests carrying different credentials.
 - Registration order is preserved; only omissions happen.
 - Returning `Mono.empty()` from an async filter omits the tool. An error fails the whole listing
-  request rather than silently hiding tools.
+  request rather than silently hiding tools: a client cannot tell a filtered-down listing from a
+  partial one, and MCP has no way to signal "this listing was incomplete, retry".
+- A filter that errors is logged server-side and reported to the client as an opaque
+  `-32603 Internal error` with no `data`. If you want the client to see a specific error, throw an
+  `McpError`, those are passed through.
 - Filters accumulate as a boolean **AND**: a tool is listed only when every registered filter accepts it, so a
   later `addToolFilter(...)` can never widen access. Evaluation follows registration order and
   short-circuits on the first filter that hides a tool.
@@ -504,16 +508,26 @@ The same `addToolFilter(...)` method is available on the stateless builders.
     ```
 
 - Tools are tested one at a time, so a filter that performs I/O per tool costs one round trip per
-  tool. Resolve per-request state **once** in your `contextExtractor` and read it in the filter:
+  tool. Sync filters also run on a shared scheduler thread — not the request thread — unless
+  `immediateExecution(true)` is set, so thread-bound request state (Spring Security's
+  `SecurityContextHolder`, MDC, custom `ThreadLocal` holders) is **not visible** inside the filter.
+  For both reasons, resolve per-request state **once** in the transport's `contextExtractor`,
+  which does run on the request thread, and read only the extracted context in the filter:
 
     ```java
-    // one authorization lookup, shared by every tool tested in this request
-    .contextExtractor(request -> McpTransportContext.create(
-            Map.of("perms", introspect(request.getHeader("Authorization")))))
+    // transport builder: one authorization lookup, on the request thread,
+    // shared by every tool tested in this request
+    var transportProvider = HttpServletStreamableServerTransportProvider.builder()
+        .contextExtractor(request -> McpTransportContext.create(
+                Map.of("perms", introspect(request.getHeader("Authorization")))))
+        // ...
+        .build();
 
-    .addToolFilter((context, tool) -> 
-            ((Set<String>) context.get("perms")).contains(tool.name())
-    )
+    // server builder: the filter reads only the extracted context
+    McpServer.sync(transportProvider)
+        .addToolFilter((context, tool) ->
+                ((Set<String>) context.get("perms")).contains(tool.name()))
+        .build();
     ```
 
 - `notifications/tools/list_changed` is **not** filtered. It is a server-initiated broadcast with
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncListFilter.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncListFilter.java
@@ -65,10 +65,10 @@ public interface McpAsyncListFilter<T> {
 	 * contain {@code null} elements.
 	 */
 	static <T> McpAsyncListFilter<T> and(List<McpAsyncListFilter<T>> filters) {
-		Assert.noNullElements(filters, "filters must not contain null elements");
 		if (filters == null || filters.isEmpty()) {
 			return (transportContext, primitive) -> Mono.just(Boolean.TRUE);
 		}
+		Assert.noNullElements(filters, "filters must not contain null elements");
 		if (filters.size() == 1) {
 			return filters.get(0);
 		}
--- a/mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncListFilter.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncListFilter.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2026-2026 the original author or authors.
+ */
+
+package io.modelcontextprotocol.server;
+
+import java.util.List;
+
+import io.modelcontextprotocol.common.McpTransportContext;
+import io.modelcontextprotocol.spec.McpSchema.Tool;
+import io.modelcontextprotocol.util.Assert;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+
+/**
+ * Decide per request whether a primitive is advertised in the corresponding listing, such
+ * as {@code tools/list}.
+ * <p>
+ * A primitive hidden by this filter is omitted from listings ONLY. It remains reachable
+ * through its own endpoint: a hidden tool called by name still executes. Permissions MUST
+ * be enforced in the primitive's handler.
+ *
+ * @author Daniel Garnier-Moiroux
+ * @see McpSyncListFilter
+ * @see McpTransportContextExtractor
+ */
+@FunctionalInterface
+public interface McpAsyncListFilter<T> {
+
+	/**
+	 * Whether the given primitive is visible to the caller of the current request.
+	 * @param transportContext transport context containing, for example, HTTP headers or
+	 * a resolved principal. Should never be {@code null}, but may
+	 * {@link McpTransportContext#EMPTY} for transports that carry no per-request
+	 * metadata, such as STDIO.
+	 * @param primitive the primitive that is a candidate for inclusion in the listing,
+	 * such as {@link Tool}.
+	 * @return a publisher emitting {@code true} to include the primitive in the listing,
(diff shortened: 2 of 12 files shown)


mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
```diff
@@ -440,6 +443,43 @@ public AsyncSpecification<S> validateToolInputs(boolean validate) {
...
+
+		/**
+		 * Applies the given consumer to the list of registered tool filters, allowing
+		 * them to be inspected, reordered or cleared before the server is built.
+		 * @param toolFiltersConsumer consumer of the mutable list of registered filters,
+		 * must not be null
+		 * @return This builder instance for method chaining
+		 * @see #addToolFilter(McpAsyncListFilter)
+		 */
+		public AsyncSpecification<S> toolFilters(
```
tzolov, line 476: Not convinced about the practicality of such flexibility. It exposes the builder's internal mutable list.  Perhaps a plain `toolFilters(List<McpSyncListFilter<Tool>>)`,  or deferring the method until demand appears, would be the safer?

mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
```diff
@@ -1043,6 +1085,44 @@ public SyncSpecification<S> validateToolInputs(boolean validate) {
...
+
+		/**
+		 * Applies the given consumer to the list of registered tool filters, allowing
+		 * them to be inspected, reordered or cleared before the server is built.
+		 * @param toolFiltersConsumer consumer of the mutable list of registered filters,
+		 * must not be null
+		 * @return This builder instance for method chaining
+		 * @see #addToolFilter(McpSyncListFilter)
+		 */
+		public SyncSpecification<S> toolFilters(Consumer<List<McpSyncListFilter<McpSchema.Tool>>> toolFiltersConsumer) {
```
tzolov, line 1120: Same like the AsyncSpecification.toolFilters

mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
```diff
@@ -1582,6 +1664,43 @@ public StatelessAsyncSpecification validateToolInputs(boolean validate) {
...
+
+		/**
+		 * Applies the given consumer to the list of registered tool filters, allowing
+		 * them to be inspected, reordered or cleared before the server is built.
+		 * @param toolFiltersConsumer consumer of the mutable list of registered filters,
+		 * must not be null
+		 * @return This builder instance for method chaining
+		 * @see #addToolFilter(McpAsyncListFilter)
+		 */
+		public StatelessAsyncSpecification toolFilters(
```
tzolov, line 1694: Same like the AsyncSpecification.toolFilters

mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
```diff
@@ -2082,6 +2204,45 @@ public StatelessSyncSpecification validateToolInputs(boolean validate) {
...
+
+		/**
+		 * Applies the given consumer to the list of registered tool filters, allowing
+		 * them to be inspected, reordered or cleared before the server is built.
+		 * @param toolFiltersConsumer consumer of the mutable list of registered filters,
+		 * must not be null
+		 * @return This builder instance for method chaining
+		 * @see #addToolFilter(McpSyncListFilter)
+		 */
+		public StatelessSyncSpecification toolFilters(
```
tzolov, line 2233: Same like the AsyncSpecification.toolFilters

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncListFilter.java
```diff
@@ -0,0 +1,81 @@
...
+
+	/**
+	 * Combine multiple filters in a single AND-filter. An empty or {@code null} list
+	 * makes everything visible, keeping listing on a single code path when nothing is
+	 * configured.
+	 * @param filters the filters to combine. May be {@code null} or empty, but MUST NOT
+	 * contain {@code null} elements.
+	 */
+	static <T> McpAsyncListFilter<T> and(List<McpAsyncListFilter<T>> filters) {
+		Assert.noNullElements(filters, "filters must not contain null elements");
```
tzolov: this assert can go after the if (filters == null || filters.isEmpty()) check.

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncListFilter.java
```diff
@@ -0,0 +1,81 @@
...
+		if (filters == null || filters.isEmpty()) {
+			return (transportContext, primitive) -> Mono.just(Boolean.TRUE);
+		}
+		if (filters.size() == 1) {
```
tzolov, line 72: With 1 filter, Mono.empty() stays empty; with 2+, empty is coerced to false: (e.g..defaultIfEmpty(Boolean.FALSE)).  Not sure what downstream implications this might have. It will be cleaner to either remove the if (filters.size() == 1) optimization or use the same defaultIfEmpty strategy

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -537,9 +541,13 @@ public Mono<Void> notifyToolsListChanged() {
...
 	private McpRequestHandler<McpSchema.ListToolsResult> toolsListRequestHandler() {
 		return (exchange, params) -> {
-			List<Tool> tools = this.tools.stream().map(McpServerFeatures.AsyncToolSpecification::tool).toList();
-
-			return Mono.just(McpSchema.ListToolsResult.builder(tools).build());
+			// TODO: Implement pagination. Cursors must be computed over the filtered
+			// view, otherwise page offsets leak the number of hidden tools.
+			return Flux.fromIterable(this.tools)
+				.map(McpServerFeatures.AsyncToolSpecification::tool)
+				.filterWhen(tool -> this.toolFilter.isVisible(exchange.transportContext(), tool))
```
tzolov: The filterWhen doesn't catch any Mono.error or exceptions thrown in the filters and this error is propagated through the McpServerSession as McpError back to the MCP client. Not sure if it is safe to let the clients see those type of filter errors?

docs/server.md
```diff
@@ -441,6 +441,88 @@ var syncToolSpecification = SyncToolSpecification.builder()
...
+        .toolFilters(filters -> filters.add(0, cheapDenyAllForAnonymousFilter))
+        .build();
+    ```
+
+- Tools are tested one at a time, so a filter that performs I/O per tool costs one round trip per
+  tool. Resolve per-request state **once** in your `contextExtractor` and read it in the filter:
+
+    ```java
+    // one authorization lookup, shared by every tool tested in this request
+    .contextExtractor(request -> McpTransportContext.create(
```
tzolov: I guess this is from the transport while the addToolFilter is from the server builder. Snippet gives the wrong impression that they belong to a common builder. 
AI suggested a re-write like:

Sync filters also run on a shared scheduler thread — not the request thread — unless
  `immediateExecution(true)` is set, so thread-bound request state (Spring Security's
  `SecurityContextHolder`, MDC, custom `ThreadLocal` holders) is **not visible** inside the filter.
  For both reasons, resolve per-request state **once** in the transport's `contextExtractor`,
  which does run on the request thread, and read only the extracted context in the filter:

    ```java
    // transport builder: one authorization lookup, on the request thread,
    // shared by every tool tested in this request
    var transportProvider = HttpServletStreamableServerTransportProvider.builder()
        .contextExtractor(request -> McpTransportContext.create(
                Map.of("perms", introspect(request.getHeader("Authorization")))))
        // ...
        .build();

    // server builder: the filter reads only the extracted context
    McpServer.sync(transportProvider)
        .addToolFilter((context, tool) ->
                ((Set<String>) context.get("perms")).contains(tool.name()))
        .build();
    ```

mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
```diff
@@ -1582,6 +1664,43 @@ public StatelessAsyncSpecification validateToolInputs(boolean validate) {
...
 
+		/**
+		 * Adds a per-request filter deciding which tools are advertised in
+		 * {@code tools/list}, for example to hide tools the caller is not authorized to
+		 * see. Filters accumulate: a tool is listed only when every registered filter
+		 * accepts it, so a later registration can never widen access.
+		 * <p>
+		 * A hidden tool is omitted from listings only. It remains callable by name, so
+		 * enforce permissions in the tool's call handler. Tools are NOT hidden from
+		 * {@code notifications/tools/list_changed}, as it is a per-client context rather
```
tzolov: Stateless servers don't support notifications

mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
```diff
@@ -2082,6 +2204,45 @@ public StatelessSyncSpecification validateToolInputs(boolean validate) {
...
```
tzolov: no notifications support for the stateless servers

mcp-core/src/main/java/io/modelcontextprotocol/server/McpSyncListFilter.java
```diff
@@ -0,0 +1,38 @@
...
+import io.modelcontextprotocol.spec.McpSchema;
+
+/**
+ * Decide per request whether a primitive is advertised in the corresponding listing, such
+ * as {@code tools/list}.
+ * <p>
+ * A primitive hidden by this filter is omitted from listings ONLY. It remains reachable
+ * through its own endpoint: a hidden tool called by name still executes. Permissions MUST
+ * be enforced in the primitive's handler.
+ *
```
tzolov, line 17: Add "Unless immediateExecution is enabled, this filter runs on a shared scheduler thread, not the request thread. Thread-bound request state (SecurityContextHolder, MDC, custom ThreadLocal holders) is not visible here — resolve it in your transport's contextExtractor and read it from the transportContext parameter." 
The same warning on the sync builders' addToolFilter.
(AI suggestion)

tzolov (changes requested):
Thanks @Kehrlann,  
please check my comments

mcp-core/src/main/java/io/modelcontextprotocol/server/McpServer.java
Kehrlann: We've been called out for this specifically in https://github.com/modelcontextprotocol/java-sdk/issues/1073 (for request customizers). If a framework (e.g. Spring AI + MCP Security) sets a customizer, they want to be able to add to it, or reorder the customizers. It's likely going to be the same problem here. 

(This pattern is common in Spring Security.)

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncListFilter.java
Kehrlann: empty is equivalent to false; there are no semantic differences
it's not designed for end user consumption but rather for the SDK, so I feel we're safe here.


Updated the comment on the `.isVisible` method to document this.

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
Kehrlann: Wrapped in MCP internal error, logged server side.

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java
```diff
@@ -537,12 +541,32 @@ public Mono<Void> notifyToolsListChanged() {
...
-			List<Tool> tools = this.tools.stream().map(McpServerFeatures.AsyncToolSpecification::tool).toList();
-
-			return Mono.just(McpSchema.ListToolsResult.builder(tools).build());
+			// TODO: Implement pagination. Cursors must be computed over the filtered
+			// view, otherwise page offsets leak the number of hidden tools.
+			return Flux.fromIterable(this.tools)
+				.map(McpServerFeatures.AsyncToolSpecification::tool)
+				.filterWhen(tool -> this.toolFilter.isVisible(exchange.transportContext(), tool)
+					.onErrorResume(error -> opaqueListFilterError(tool, error)))
+				.collectList()
```
tzolov, line 550: This is an existing conditions, but apparently the toolsListRequestHandler returns a mutable list of tools. 
Adding  `.map(List::copyOf)` should make it immutable. E.g. 

```java
...
.collectList()
.map(List::copyOf)
.map(tools -> McpSchema.ListToolsResult.builder(tools).build());
```

mcp-core/src/main/java/io/modelcontextprotocol/server/McpAsyncListFilter.java
tzolov: Perhaps something like: 
``` java
    if (filters.size() == 1) {
        McpAsyncListFilter<T> filter = filters.get(0);
        return (transportContext, primitive) -> filter.isVisible(transportContext, primitive)
            .defaultIfEmpty(Boolean.FALSE);
    }
```

tzolov (commented):
There some changes (like the defaultIfEmpty and copyOf) I can apply as part of the merge if this is ok with you?

tzolov (approved):
LGtM

Issue #1110, opened 2026-08-27
JsonSchemaValidator.validate parses String input as serialized JSON; document this contract


slachiewicz:
## Summary

The `validate` method of `JsonSchemaValidator` treats a `String` `structuredContent` argument as a serialized JSON document to parse before validation. It does not treat the argument as a literal JSON string value.
This behavior is intentional and consistent across both JSON modules, but it is undocumented, which makes the API easy to misuse when you validate schemas whose document root is a scalar.

## Steps to reproduce

Suppose you validate a root-level string against an `enum` schema:

```java
var validator = new DefaultJsonSchemaValidator();
var schema = Map.<String, Object>of("enum", List.of("red", "green"));
var response = validator.validate(schema, "red");
```

The validation fails even though `red` is a member of the enumeration:

```text
Error parsing tool JSON Schema: Unrecognized token 'red': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
```

Two things confuse readers here:

1. The failure comes from JSON parsing, not from the `enum` keyword. NetworkNT  evaluates root-level enums correctly when the instance reaches it as a  parsed node.
2. The message says "tool JSON Schema", although the schema parses fine and the  *content* is what fails to parse.

Passing quoted JSON text produces the expected results:

```java
validator.validate(schema, "\"red\"");   // valid=true
validator.validate(schema, "\"blue\"");  // valid=false
```

Non-string instances follow the other path. The following call validates
correctly because Jackson converts the value through `valueToTree`:

```java
var numericSchema = Map.<String, Object>of("enum", List.of(1, 2));
validator.validate(numericSchema, 1);    // valid=true
```

## Current implementation

Both default validators contain the same special case:

- `mcp-json-jackson2`: `io.modelcontextprotocol.json.schema.jackson2.DefaultJsonSchemaValidator`,
  the `instanceof String` branch at line 70
- `mcp-json-jackson3`: `io.modelcontextprotocol.json.schema.jackson3.DefaultJsonSchemaValidator`,
  the `instanceof String` branch at line 69

When the argument is a `String`, the validator calls `jsonMapper.readTree` on
it; otherwise it calls `jsonMapper.valueToTree`.

## Suggested resolution

This is a documentation gap, not a behavioral bug. Add the contract to the
`JsonSchemaValidator#validate` Javadoc: if `structuredContent` is a `String`,
the method parses it as serialized JSON; quote embedded strings accordingly
[...]

PR #1111, opened 2026-08-27
docs(core): document String structuredContent parsing in JsonSchemaValidator

slachiewicz:
## What this change does

Adds the input-conversion contract of the `validate` method to the
`JsonSchemaValidator` interface Javadoc.

The method treats a `String` argument as a serialized JSON document and parses
it with `readTree` before validation. It converts any other type directly with
`valueToTree`. Both consequences follow for callers:

- To validate a root-level string, pass quoted JSON text: `"\"red\""`
  represents the value `red`.
- A bare string such as `red` reaches the parser as an unquoted token, and
  parsing fails with a message that mentions the tool JSON Schema, even though
  the schema is valid.

This contract exists identically in both implementations
(`mcp-json-jackson2` at line 70, `mcp-json-jackson3` at line 69). The interface
Javadoc did not state it before this change.

## Why

During output-schema validation testing, a root-level `enum` check failed in a
way that pointed at the schema parser rather than at caller-side string
handling. Tracing both default validators revealed the undocumented branch.
Documentation-only changes to a shared SPI carry no migration risk, and the
wording mirrors what the code does so future refactors can update one place.

## Testing

No production code changed. Before opening this pull request, I verified:

- `mcp-json-jackson3`: `DefaultJsonSchemaValidatorTests` passes, 40 of 40 tests
- `mcp-json-jackson2`: `DefaultJsonSchemaValidatorTests` passes, 40 of 40 tests
- `spring-javaformat:validate` passes for the touched module

## Notes for reviewers

If you prefer stronger ergonomics over pure documentation (for example, a
dedicated overload or a clearer parse-failure message), this wording still
stands on its own. Follow-up issues can adjust either message or API without
conflict. Known sharp edge, intentionally out of scope here: the
"Error parsing tool JSON Schema" prefix also appears when the *content* fails
to parse; existing test assertions pin that string in both modules.

### DIFF
--- a/mcp-core/src/main/java/io/modelcontextprotocol/json/schema/JsonSchemaValidator.java
+++ b/mcp-core/src/main/java/io/modelcontextprotocol/json/schema/JsonSchemaValidator.java
@@ -57,6 +57,11 @@ public interface JsonSchemaValidator {
 
 	/**
 	 * Validates the structured content against the provided JSON schema.
+	 * <p>
+	 * If {@code structuredContent} is a {@link String}, it is treated as a serialized
+	 * JSON document and parsed before validation; quote embedded strings accordingly (for
+	 * example {@code "\"red\""} for the JSON string value {@code red}). Any other type is
+	 * converted to its JSON representation directly.
 	 * @param schema The JSON schema to validate against.
 	 * @param structuredContent The structured content to validate.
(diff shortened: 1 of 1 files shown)


Kehrlann:
Thanks for the docs update.

Fixes #1110

Kehrlann (approved):

PR #1026, opened 2026-06-12
Send Mcp-Method header on Streamable HTTP client requests (SEP-2243)

nikita-kibitkin:
## Motivation and Context

Contributes the client-side Mcp-Method part of #990.

SEP-2243 (final, https://modelcontextprotocol.io/seps/2243-http-standardization) requires Streamable HTTP clients to mirror the JSON-RPC method of every request and notification into the `Mcp-Method` HTTP header, so proxies and load balancers can route MCP traffic without parsing the body. The client transport currently does not send this header.

#994 covers the `Mcp-Name` header. This PR is limited to client-side emission of `Mcp-Method`; server-side validation of header/body consistency is left as a follow-up, since #994 already touches that area for `Mcp-Name`. If #994 merges first, I will rebase on top of it.

## Description

- New `HttpHeaders.MCP_METHOD` constant.
- `HttpClientStreamableHttpTransport#sendMessage` sets `Mcp-Method` from the message method for `JSONRPCRequest` and `JSONRPCNotification`. `JSONRPCResponse` has no method and is sent without the header.

## How Has This Been Tested?

Three new tests in `HttpClientStreamableHttpTransportTest` (request, notification, response cases), using the existing request-customizer hook to capture the outgoing request. The two header-presence tests were verified to fail without the production change. All streamable HTTP client suites pass locally:

```
./mvnw test -pl mcp-test -am -Dtest="HttpClientStreamableHttp*" -Dsurefire.failIfNoSpecifiedTests=false
129 tests, 0 failures
```

## Breaking Changes

None. The header is additive and limited to Streamable HTTP client POSTs.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #994, opened 2026-06-03
feat(mcp-core): support Mcp-Name header for Streamable HTTP (SEP-2243)

cooleditphoto:
Add MCP_NAME header constant and inject it in the streamable HTTP client transport (HttpClientStreamableHttpTransport) and HTTP server transport

## Motivation and Context
https://github.com/modelcontextprotocol/java-sdk/issues/990
To solve the Mcp_Name part of https://modelcontextprotocol.io/seps/2243-http-standardization
## How Has This Been Tested?
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  04:44 min
[INFO] Finished at: 2026-06-02T09:02:38-04:00
[INFO] ------------------------------------------------------------------------
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #1112, opened 2026-08-27
feat(transport): add SEP-2243 Mcp-Method / Mcp-Name header mirroring with MCP-Protocol-Version validation

### LINKED ISSUE #990: SEP-2243: HTTP Header Standardization for Streamable HTTP Transport

slachiewicz:
Closes #990. Supersedes #994, #1026, #1092 (with credit to @cooleditphoto, @nikita-kibitkin, and @ez-lbz — the consolidation proposed in #990; @nikita-kibitkin agreed to fold in #1026).

## Summary

Implements [SEP-2243](https://modelcontextprotocol.io/seps/2243-http-standardization) HTTP header standardization end-to-end:

- The JDK Streamable HTTP client emits an `Mcp-Method` header on every POST and an `Mcp-Name` header whenever the request targets a named artifact (`tools/call`, `prompts/get`, `resources/read`, `resources/subscribe`, `resources/unsubscribe`).
- Non-ASCII, control character, or sentinel-matching `Mcp-Name` values are safely encoded using the SEP-2243 Base64 sentinel format (`=?base64?...?=`), protecting against JDK `HttpClient` `IllegalArgumentException` on non-Latin-1 characters.
- Both servlet server transports validate present `Mcp-Method` / `Mcp-Name` headers against the deserialized request body (decoding Base64 sentinels where present), returning `HEADER_MISMATCH` (`-32020`) with HTTP 400 on mismatches.
- Servlet server transports validate the `MCP-Protocol-Version` header against the supported version set (exempting `initialize` requests, which negotiate version via payload).
- Includes the deflake for `HttpClientStreamableHttpVersionNegotiationIntegrationTests`.

## Semantics (the decision this PR stands on)

SEP-2243 requires servers to reject header/body mismatches, and discussion on #994 established that rejecting *absent* headers must not break legacy clients:

- A present-but-mismatched `Mcp-Method` or `Mcp-Name` is rejected with `400 Bad Request` and `HEADER_MISMATCH` (`-32020`).
- Absent or blank headers never trigger validation errors by default, so clients that do not send the new headers keep working unchanged.
- Rejecting absent headers is opt-in via `requireMcpHeaders(true)` on the servlet transport builders, default off.
- An absent `MCP-Protocol-Version` falls back to negotiated behavior, while a present-but-unsupported version on post-init requests is rejected with `INVALID_REQUEST`.

## Implementation notes

- New constants `HttpHeaders.MCP_METHOD`, `HttpHeaders.MCP_NAME`, `HttpHeaders.BASE64_SENTINEL_PREFIX`, `HttpHeaders.BASE64_SENTINEL_SUFFIX`, and helper methods `HttpHeaders.encodeHeaderValue` / `HttpHeaders.decodeHeaderValue`.
- New error code `McpSchema.ErrorCodes.HEADER_MISMATCH = -32020`.
- Client emission happens in `HttpClientStreamableHttpTransport.sendMessage` for both requests and notifications. The name/URI comes from typed binding of the params (`tools/call`, `prompts/get` use `.name()`; the resource operations use `.uri()`), so no regex-on-body parsing is involved. Extraction failures log at debug and omit the header rather than fail the request.
- Server checks run after JSON-RPC deserialization in the servlet transports, comparing header against the *parsed* body. Responses carry no method and are skipped.
- The server checks are shared through a package-private `Sep2243RequestValidator` that reads headers via `HeaderAccessor`. It sits beside `ServerHttpHeaderValidator` rather than inside it because the mirroring checks need the parsed body.

## Testing

- `Sep2243ClientRequestHeaderTests` asserts emitted headers through a real JDK HTTP server round-trip, including Base64 sentinel encoding for non-ASCII tool names.
- `Sep2243ServerHeaderValidationTests` covers the rejection matrix (bad protocol version, method mismatch, name mismatch with error code `-32020`), Base64 decoding acceptance, the tolerance rule (absent headers do not trigger SEP-2243 errors by default), and the `requireMcpHeaders` strict mode, for both the streamable servlet provider and the stateless transport.
- `HttpClientStreamableHttpVersionNegotiationIntegrationTests` verifies protocol version propagation and async stream readiness.

Co-authored-by: Nikita Kibitkin <nikita.n.kibitkin@gmail.com>


slachiewicz:
Pushed `67e89e70`, addressing three gaps.

- `Builder.requireMcpHeaders(boolean)` on both servlet transports, default `false`. When on, a request that omits `Mcp-Method`, or targets a tool, prompt or resource without `Mcp-Name`, is rejected with 400 and `HEADER_MISMATCH`. This is the opt-in toggle discussed in #990; a present but mismatched header is rejected regardless.
- The validation now lives in one package-private `Sep2243RequestValidator` shared by both transports, reading headers through the `HeaderAccessor` from #870. It sits beside `ServerHttpHeaderValidator` rather than inside it because the mirroring checks need the parsed JSON-RPC body. Happy to move it if you would rather widen that interface.
- An unsupported `MCP-Protocol-Version` returns `INVALID_REQUEST` instead of `METHOD_NOT_FOUND`.

Verified: `./mvnw -B -pl mcp-core -am test` → 408 tests, 0 failures.

PR #1115, opened 2026-08-27
fix: tolerate unregistered listening stream for server-initiated messages

slachiewicz:
Right after initialize, the server may push a server-initiated request (e.g. roots/list when the client declared the roots capability) before the client has opened its GET /mcp stream; the session then failed with Stream unavailable for session <id> (intermittent -32603/500, seen in HttpServletStreamableIntegrationTests CI). sendRequest/sendNotification now retry missing-stream failures with bounded backoff (~5s) and accept(notification) routes handler-initiated pushes through the retrying session facade. Related: #952, #1061. New unit tests cover buffer-until-register, correlation, grace exhaustion, and the handler-push path. Happy to adapt the approach per maintainer guidance.

Fixes #1114

PR #1119, opened 2026-08-31
Fix integration tests for `roots/list`

### LINKED ISSUE #1114: HttpServletStreamableServerTransportProvider: server-initiated requests race client stream registration (intermittent -32603 "Stream unavailable")
### Summary

`HttpServletStreamableServerTransportProvider` intermittently answers `-32603 "Stream unavailable for session <id>"` (wrapped as JSON-RPC internal error / HTTP 500) when it attempts to push a server-initiated request (e.g. `roots/list`) to a client whose `GET /mcp` SSE stream has not been registered yet.

### Observed failure

- Test: `HttpServletStreamableIntegrationTests.testRootsNotificationWithEmptyRootsList(String)[1]`
- CI run: https://github.com/modelcontextprotocol/java-sdk/actions/runs/33068920136 (Build and Test)
- Client-side surfacing: `java.lang.RuntimeException: Failed to send message: AggregateResponseEvent[...]` wrapping
  `jsonRpcError: {"code":-32603,"message":"Error processing message: Stream unavailable for session 82454020-4c10-4276-9c30-57c15d49810e"}`

### Repro shape

1. Client declares the `roots` capability
2. Client sends `initialize`; server accepts and creates the session
3. Server immediately pushes `roots/list` over the session's `GET /mcp` stream
4. POST traffic racing the first stream registration loses → `-32603`, stream unavailable

### Why we believe it is a race (not environment-specific)

- The same commit passed the parallel "Jackson 2 Integration Tests" job minutes later
- Full local suite (embedded Tomcat 11, JDK 17, macOS) passed repeatedly; only 1 of 43 parameterizations in the class failed on CI
- Related history: revert `fb029c8c` "Clean up AbstractMcpClientServerIntegrationTests", #952 (session removed on non-fatal failure), and #1061 (same roots-push seam in the stateful streamable provider, blocking variant)

### Suggested directions

1. Buffer server→client messages per session until a stream registers (short grace window), or
2. Single retry on "stream unavailable" for capability-prompted pushes

A test-side await cannot fix this, since the failure occurs inside the server's push path.

### Environment

- main @ `b31841ee` (via PR #1112 CI), JDK 17, CI ubuntu-latest + local macOS

Kehrlann:
There is a race condition where the GET /mcp SSE stream in Streamable HTTP is established _after_ the server tries to send notifications. This surfaces errors from the MCP client sending notifications and breaks some integration tests.

We add a utility showing to ensure the stream is established before sending server notifications.

Fixes #1114
Supersedes #1115

@slachiewicz would you like to review this fix of #1114?

### DIFF
--- a/mcp-test/src/test/java/io/modelcontextprotocol/server/HttpServletStreamableIntegrationTests.java
+++ b/mcp-test/src/test/java/io/modelcontextprotocol/server/HttpServletStreamableIntegrationTests.java
@@ -30,7 +30,6 @@ import jakarta.servlet.http.HttpServletResponse;
 import org.apache.catalina.LifecycleException;
 import org.apache.catalina.LifecycleState;
 import org.apache.catalina.startup.Tomcat;
-import org.awaitility.Awaitility;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
@@ -42,6 +41,7 @@ import reactor.core.publisher.Mono;
 import reactor.test.StepVerifier;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
 
 @Timeout(15)
 class HttpServletStreamableIntegrationTests extends AbstractMcpClientServerIntegrationTests {
@@ -56,6 +56,16 @@ class HttpServletStreamableIntegrationTests extends AbstractMcpClientServerInteg
 	private static Tomcat tomcat;
 
 	private HttpServletStreamableServerTransportProvider mcpServerTransportProvider;
+
+	@Override
+	protected void awaitClientStreamEstablished() {
+		var timeout = Duration.ofSeconds(5);
+		await().atMost(timeout).untilAsserted(() -> {
+			assertThat(MCP_SERVLET.isStreamEstablished())
+				.withFailMessage("[Failed to observe MCP Client connection within %s]", timeout)
+				.isTrue();
+		});
+	}
 
 	static Stream<Arguments> clientsForTesting() {
 		return Stream.of(Arguments.of("httpclient"));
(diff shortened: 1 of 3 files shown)


slachiewicz (approved):

mcp-test/src/test/java/io/modelcontextprotocol/server/HttpServletStreamableIntegrationTests.java
```diff
@@ -57,6 +57,16 @@ class HttpServletStreamableIntegrationTests extends AbstractMcpClientServerInteg
 
 	private HttpServletStreamableServerTransportProvider mcpServerTransportProvider;
 
+	@Override
+	protected void awaitClientStreamEstablished() {
+		var timeout = Duration.ofSeconds(1);
```
slachiewicz: Minor: 1s is tight next to the 5s awaits used elsewhere in this suite — on a loaded CI runner this could trade the `-32603` flake for a "Failed to observe MCP Client connection" timeout. Would you consider aligning with the 5s convention (or a shared constant)? Non-blocking either way since the failure message would at least be clearer.

mcp-test/src/test/java/io/modelcontextprotocol/server/transport/TomcatTestUtil.java
```diff
@@ -104,6 +110,9 @@ public void service(ServletRequest request, ServletResponse response) throws Ser
 				throw new IllegalStateException("No delegate servlet has been set");
 			}
 			current.service(request, response);
+			if (request instanceof HttpServletRequest req && req.getMethod().equals("GET")) {
+				sseStreamEstablished.set(true);
```
slachiewicz, line 118: Might be worth one extra line in the utility's comment: the observed signal is "a GET completed through `service()`", which equals "listening stream registered" only because registration is synchronous inside `doGet` today. A replay GET (`Last-Event-ID`) or a rejected one (e.g. 400/404 from Accept/session checks, or protocol-version validation as proposed in #1112) would set the flag without registering a stream. Current tests hit neither, so fine as-is; just so the hook isn't over-trusted if `doGet` ever changes.

slachiewicz (approved):
Verified the chain behind the barrier: `DelegatingServlet.service()` sets the flag only after the delegate returns, and a non-replay `GET /mcp` registers the listening stream synchronously inside `doGet` (`McpStreamableServerSession#listeningStream`), so a true flag implies the stream is registered in the session. `setDelegate()` resets the flag per test, and the await placement covers all four `roots` seams. LGTM, two minor nits below.

One thought on framing: since the server-side race intentionally stays as-is (consistent with the C#, TypeScript and Python SDKs), "Stabilizes tests for #1114" with the issue closed as not-planned would keep that decision discoverable, rather than "Fixes #1114".

*This review was created with AI assistance.*

Issue #1009, opened 2026-06-06
SEP-2549: TTL for List Results


localden:
Tracking implementation of [SEP-2549](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549) for the 2026-07-28 MCP spec release.

Kehrlann:
This may be tied to #578 . We need to provide appropriate hooks for users to define TTLs

jfli001:
I'd like to work on this. Planning to add a `CacheableResult` interface + optional `ttlMs`/`cacheScope` fields (Case A from CONTRIBUTING.md) to the five list result records (`ListToolsResult`, `ListPromptsResult`, `ListResourcesResult`, `ListResourceTemplatesResult`, `ReadResourceResult`). Is this scope aligned with what's expected?

PR #1123, opened 2026-09-02
Add client-side result cache for SEP-2549 TTL hints

slachiewicz:
Stacked on #1131; the first five commits belong to that PR and this one is reviewable from `1cb8f446` onward. Part of #1009.

Adds a transparent cache in `McpAsyncClient` for `tools/list`, `prompts/list`, `resources/list`, `resources/templates/list` and `resources/read`. An entry is dropped when its TTL lapses, when the matching `*_changed` or `resources/updated` notification arrives, on re-initialization and on close. Multi-page listings are never cached, because stitching a cached first page to a fresh later page mixes two server snapshots.

Caching is opt-in through `enableResultCaching(true)`. A client talking to a server that already emits positive TTLs would otherwise start serving stale listings on upgrade with no code change, and SEP-2549 says clients MAY cache. The bundled in-memory store is bounded at 512 entries, and `McpClientCacheStore` lets callers back it with a cache library without adding a dependency to `mcp-core`.

*This change was created with AI assistance.*


slachiewicz:
Pushed `07e8a0c2`, addressing a self-review of this branch.

The load-bearing fixes: cache lookups moved inside `Mono.defer` (an assembled `Mono` was replaying a snapshot past its TTL); a generation counter so a response already in flight when `*_changed` arrives is not stored; entries dropped on re-initialization and close; TTL clamped at 24h with saturating addition; `resources/read` keys include `_meta`; the store is bounded at 512 entries. A listing spanning several pages is no longer cached at all — stitching a cached first page to a freshly fetched later page mixes two server snapshots.

Listings are scoped `private`, not `public`: `tools/list` is filtered per caller since #1111, so a shared cache must never serve one principal's listing to another.

Two additions, because the caching was otherwise unreachable or unavoidable:

- `listCache(Duration, CacheScope)` on all four server builders. Listings hardcoded `ttlMs: 0`, so the server half of SEP-2549 could not be switched on.
- `enableResultCaching(boolean)`, `invalidateCache()`, and an `McpClientCacheStore` SPI on the client. The bundled in-memory store stays the default, so backing the cache with Caffeine is a user choice and `mcp-core` keeps its three compile dependencies.

`ttlMs` is now validated non-negative on the result builders, which the description already claimed. Deserialization stays lenient — a bad server hint should not make a whole listing unparseable.

Docs for both hooks are in `docs/server.md` and `docs/client.md`.

Verified: `./mvnw test` → BUILD SUCCESS, all 11 modules, 0 failures.

slachiewicz:
Split this PR to make review smaller: the schema fields, server stamping and the `listCache` builder option now live in #1131, with @aboullaite's commits kept as-is. This PR is rebased on top of it and holds only the client-side cache, which is now opt-in rather than on by default. The branch was force-pushed for the split; there were no review comments to outdate.

PR #1130, opened 2026-09-09
Streamable HTTP: Add session sweeping and close hanging streams

### LINKED ISSUE #1021: [BUG] McpStreamableServerSession does not close server-side socket when client disconnects, causing CLOSE-WAIT leak and thread pool exhaustion
## Description

When using the MCP Java SDK's Streamable HTTP server transport (via `spring-ai-starter-mcp-server-webmvc`), the server-side socket is not properly closed after the client disconnects (sends TCP FIN). This causes connections to remain in `CLOSE-WAIT` state indefinitely, each holding a Tomcat worker thread. Under moderate load, the entire Tomcat thread pool is exhausted within seconds, making the server completely unresponsive to any new requests including health checks.

## Environment

- Spring AI: 1.1.4
- MCP Java SDK: (bundled with Spring AI 1.1.4)
- Java: JDK 25
- Server: Tomcat 10.1.34 (WAR deployment via Spring Boot 3.4.1)
- Transport: Streamable HTTP (`spring.ai.mcp.server.protocol=STREAMABLE`)
- OS: Linux (Kubernetes pod, 2 CPU / 4GB RAM)

## Configuration

```yaml
spring:
  ai:
    mcp:
      server:
        type: SYNC
        protocol: STREAMABLE
        streamable-http:
          mcp-endpoint: /mcp
          keep-alive-interval: 0s
```

## Steps to Reproduce

1. Deploy an MCP Server with Streamable HTTP transport (WebMVC, SYNC mode)
2. Have an external MCP client send requests to `POST /mcp` (initialize + tools/call)
3. Client receives the tool response and closes the TCP connection (sends FIN)
4. Repeat with multiple clients (or a single client with retry logic)
5. Observe server-side socket states with `ss -tnp | grep 8080`

## Observed Behavior

After the client closes the connection:

- Server-side socket enters `CLOSE-WAIT` and is **never** closed
- The Tomcat worker thread handling that request is never released back to the pool
- Under load from a single upstream LB doing health-check retries, all 150 Tomcat threads are exhausted within ~30 seconds
- New connections (including K8s readiness probes) queue in the TCP backlog and time out

```
$ ss -tlnp | grep 8080
LISTEN 151    150    *:8080    *:*

$ ss -tnp | grep 8080 | head -5
CLOSE-WAIT 115  0  [::ffff:10.125.87.86]:8080  [::ffff:10.125.87.4]:47140
CLOSE-WAIT 115  0  [::ffff:10.125.87.86]:8080  [::ffff:10.125.87.4]:42756
CLOSE-WAIT 115  0  [::ffff:10.125.87.86]:8080  [::ffff:10.125.87.4]:47446
CLOSE-WAIT 115  0  [::ffff:10.125.87.86]:8080  [::ffff:10.125.87.4]:50138
CLOSE-WAIT 115  0  [::ffff:10.125.87.86]:8080  [::ffff:10.125.87.4]:43160

$ ss -tnp | grep 8080 | wc -l
150

$ curl --max-time 5 http://localhost:8080/health
curl: (28) Failed to connect to localhost port 8080: Connection timed out
```

All 150 connections are from the same upstream IP (load balancer), all in CLOSE-WAIT.

## Expected Behavior

When the client closes the TCP connection (sends FIN), the server should:

1. Detect the peer shutdown (e.g., via IOException on write, or checking `SocketChannel.read() == -1`)
2. Close the SSE stream / Reactor Sink associated with that session
3. Remove the session from the internal session map
4. Close the server-side socket
5. Release the Tomcat thread back to the pool

## Root Cause Analysis

The MCP Streamable HTTP transport opens an SSE stream for each session. When the client disconnects:
- The server-side `Sinks.Many` has no subscribers, but the stream is never terminated
- The Servlet async context is never completed
- The socket remains open on the server side (only client sent FIN)
- Tomcat's NIO connector holds the thread waiting for the async context to complete

## Impact

- **Severity: Critical** — renders the server completely unresponsive
- Makes rolling deployments impossible in production (new pods get flooded by retrying clients immediately after startup)
- K8s readiness probes fail → pod marked unhealthy → never enters service
- No automatic recovery — requires pod restart AND stopping upstream traffic simultaneously

## Workaround

Set Tomcat connection timeout to force-close idle connections:

```yaml
server:
  tomcat:
    connection-timeout: 30000
    keep-alive-timeout: 30000
    max-connections: 200
    threads:
      max: 200
```

This allows Tomcat to reclaim CLOSE-WAIT connections after 30 seconds, but is not a proper fix — it just limits the damage window.

## Suggested Fix

The Streamable HTTP transport provider should register a listener for client disconnect events. In the WebMVC integration:

```java
// When setting up the async response for SSE:
asyncContext.addListener(new AsyncListener() {
    @Override
    public void onComplete(AsyncEvent event) {
        cleanupSession(sessionId);
    }
    @Override
    public void onTimeout(AsyncEvent event) {
        cleanupSession(sessionId);
    }
    @Override
    public void onError(AsyncEvent event) {
        cleanupSession(sessionId);
    }
    // ...
});
```

Or detect write failures when attempting to send data to the client and trigger session cleanup.

### LINKED ISSUE #1022: [BUG] KeepAliveScheduler does not evict sessions after ping failure, causing unbounded CPU growth

Kehrlann:
Fixes #1021
Fixes #1022

## Context

Streamable HTTP sessions and the connections they run on had no server-side lifecycle management, and expected the client to explicitly call `DELETE` to remove the session.
This lead to the session map growing infinitely (see #1022)

Additionally, the KeepAliveScheduler, when enabled, added pressure by sending frequent pings when no stream existed (#1021).

## Implementation

This PR introduces fixes for various streams never being closed properly, as well as a session-sweeping mechanism which empties the session map when no activity is discovered.

Changes:

#### KeepAliveScheduler
- The keep-alive scheduler now terminates the `GET` SSE stream whenever a ping fails. The stream may later be re-established by the client.
- The keep-alive scheduler only tries to ping when an SSE stream is available for the given session.
- The keep-alive scheduler is now on by default, as it's the only way to reliably discover a stream has been terminated on the client side. Default interval is 30 minutes.

#### Session eviction

- Sessions are now periodically evicted from the server, based on a `sessionSweepInterval` parameter (defaults to 30 minutes).
- Sessions are marked "active" whenever they receive a client message through `POST`. 
- Active sessions, as well as sessions with a live SSE stream, are not evicted.

#### Session cleanup

- On close, sessions now clean up pending responses as well as open SSE streams awaiting responses (e.g. POST -> SSE stream with an elicitation request).

### Known limitations

- A handler (e.g. tool handler) that never completes still pins the container thread blocked in `doPost`; the unblock depends on a write failing


lxq19991111:
Nice to see the connection/session split made explicit — `releaseTransport()` vs `close()` is the distinction that was missing in #1021/#1022.

One semantics question while this is still in progress: I think the idle clock may not restart when the last open stream is released.

`lastActivityNanos` is updated when client activity happens, but the three places that do `openStreams.remove(this)` (`close`, `closeGracefully`, `releaseTransport`) leave it untouched. So the time a session spent legitimately holding a connection is counted as idle time once that connection goes away, and the reconnection window a client actually gets is `max(0, idleTimeout - connectionLifetime)`. A stream that stayed open longer than the timeout leaves no window at all.

That seems to conflict with the intent expressed in `onPingFailure`, where the session is kept precisely so the client may reconnect and the idle timeout only reclaims it if the client never does.

Ran this against bf508ca — it passes as written, i.e. the session is immediately evictable the moment the connection is released:

    @Test
    void idleClockCountsTimeSpentHoldingAnOpenStream() throws Exception {
        var idleTimeout = Duration.ofMillis(300);
        var session = session();
        var stream = session.listeningStream(new RecordingTransport());

        // The session holds a connection for longer than the idle timeout
        Thread.sleep(400);
        assertThat(session.isIdleFor(idleTimeout)).isFalse();

        // The connection is released right now, as onPingFailure would
        stream.releaseTransport();

        // ...and the session is already considered idle for 400ms
        assertThat(session.isIdleFor(idleTimeout)).isTrue();
    }

Should the idle timeout start when the last open stream is released, rather than at the last client activity? `sessionHoldingAnOpenStreamIsNeverIdle` uses `Duration.ZERO`, so it passes under either reading — a positive timeout with a stream held longer than it would pin down the intended one.

Happy to help verify whichever semantics you settle on.

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
```diff
@@ -913,6 +1024,21 @@ public Builder contextExtractor(McpTransportContextExtractor<HttpServletRequest>
 		 * keep-alive will be scheduled.
 		 * @return this builder instance
 		 */
+		/**
```
lxq19991111, line 1027: Minor: the new `sessionIdleTimeout` javadoc landed between the existing `keepAliveInterval` javadoc and its method, so there are now two consecutive javadoc blocks and `keepAliveInterval()` is left undocumented.

Kehrlann:
@lxq19991111 there is a spec-compliant case where no SSE stream is ever established - the client only issues POST requests and the server sends `application/json` HTTP responses (like our MCP Stateless server does). So the existence of a stream does not indicate the liveness of the session. I think it's better to track client _requests_ rather than streams. 

Now should we indeed count disconnect as client activity? It can be a client initiated action, but in that case the client is likely signaling that it _wants_ to disconnect, so I'm not sure it's worth keeping their session active for longer.

lxq19991111:
That makes sense. Client-originated traffic is a better liveness signal than stream presence, especially for POST-only clients, and I agree a disconnect shouldn't by itself refresh session activity. My test shows the current behavior but doesn't establish that a fresh idle window should start on disconnect — thanks for clarifying.

Separately, while tracing the cleanup paths I noticed an asymmetry between the GET and POST paths. `doGet` registers the wrapper's method (`listeningStream::releaseTransport`), which removes it from `openStreams`, whereas `doPost` registers the raw transport's method (`sessionTransport::close`). That marks the transport closed and completes the `AsyncContext`, but it doesn't remove the `McpStreamableServerSessionStream` created inside `responseStream()` from `openStreams`.

This normally self-heals when the response pipeline terminates and runs `stream.closeGracefully()`. However, if the handler never completes, the wrapper continues to block idle sweeping. With a `Mono.never()` handler, `session.isIdleFor(Duration.ZERO)` is still false after closing the raw transport.

My inclination would be to tie the cleanup to the session stream wrapper, since it owns the `openStreams` registration and closing only the lower-level transport leaves the higher-level lifecycle state inconsistent. Cancelling the pipeline alone wouldn't
cover it either: `.then(stream.closeGracefully())` isn't reached on cancel, so `isIdleFor(Duration.ZERO)` stays false after disposing the subscription. That route would need a `doFinally`, the way `sendRequest` now handles cancellation. Does that match the intended ownership here?

lxq19991111:
I pulled the latest branch (c1cc11c) and re-ran the two session-level cases I raised earlier. With a Mono.never() handler, the POST response stream now detaches from the session both when releaseTransport() is called and when the subscription is cancelled; hasOpenStream() becomes false in both cases. Thanks for fixing that cleanup gap. 

I also reproduced the remaining limitation more narrowly: after releaseTransport(), the stream is detached and hasOpenStream() is false, but a thread waiting in responseStream.handle(...).block() remains parked because releasing the transport does not terminate that subscription.

A focused follow-up could remove this explicit wait for asynchronously completing handlers. One possible approach would be to subscribe with explicit terminal-error handling, retain the subscription in a race-safe holder, and dispose it when the existing AsyncListener reports the end of the async lifecycle. Cancellation could then reuse the cleanup already centralized in usingWhen. 

One semantic point to confirm is pending requests: cancellation calls closeGracefully(), which fails server-initiated requests still pending on that stream, while releaseTransport() deliberately keeps them resolvable through a separate POST. 

This would not detect a silent peer disconnect when no further I/O occurs, especially with setTimeout(0), and it would not prevent a handler from blocking during subscription. Those cases would require a separate timeout or processing-deadline policy. 

Would you be open to a focused follow-up PR for this narrower change, including regression coverage for the Mono.never() path, synchronous-completion races, and terminal write errors?

lxq19991111:
I pulled the latest branch (c1cc11c) and re-ran the two session-level cases I raised earlier. The recent POST cleanup and disconnected-client changes work as expected: with a Mono.never() handler, the response stream now detaches from the session both when releaseTransport() is called and when the subscription is cancelled, and hasOpenStream() becomes false in both cases. Thanks for addressing that cleanup gap.

The remaining limitation I see is narrower than the disconnected-client case covered by the latest commits: after releaseTransport(), the stream is detached and hasOpenStream() is false, but a thread waiting in responseStream.handle(...).block() remains parked because the handler subscription itself is still active. The PR already documents the general case where a never-completing handler can keep the doPost thread blocked, so I see this as a possible follow-up rather than a blocker for this PR.

One possible approach would be to avoid the explicit wait for asynchronously completing handlers, subscribe with explicit terminal-error handling, retain the subscription in a race-safe holder, and dispose it when the existing AsyncListener reports the end of the async lifecycle. Cancellation could then reuse the cleanup already centralized in usingWhen.

One semantic point to confirm would be how this should interact with pending server-initiated requests: cancellation calls closeGracefully(), while releaseTransport() keeps the session available for a separate POST.

This would still not detect a silent peer disconnect when no further I/O occurs, especially with setTimeout(0), and it would not prevent a handler from blocking during subscription; those cases would require a separate timeout or processing-deadline policy.

Would you be open to a focused follow-up PR for this narrower limitation, with regression coverage for the Mono.never() path, synchronous-completion races, and terminal write errors?

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
```diff
@@ -165,15 +191,95 @@ private HttpServletStreamableServerTransportProvider(McpJsonMapper jsonMapper, S
...
+	 * Schedules the periodic eviction of idle sessions.
+	 * @param sessionSweepInterval the interval between two sweeps
+	 * @return the handle on the scheduled sweeps, to be disposed when the transport shuts
+	 * down
+	 */
+	private Disposable startSessionSweeper(Duration sessionSweepInterval) {
+		return Flux.interval(sessionSweepInterval, sessionSweepInterval, Schedulers.boundedElastic())
+			.concatMap(
+					tick -> sweepSessions().doOnError(e -> logger.error("Session sweep failed", e)).onErrorComplete())
+			.onErrorComplete(error -> {
```
chemicL: This can be passed as a subscribe(...) argument to avoid some nesting.

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
```diff
@@ -165,15 +191,95 @@ private HttpServletStreamableServerTransportProvider(McpJsonMapper jsonMapper, S
...
+	 * @return a Mono completing once every evicted session has been closed
+	 */
+	private Mono<Void> sweepSessions() {
+		if (this.isClosing) {
+			return Mono.empty();
+		}
+		Set<String> active = this.activeSessions.getAndSet(ConcurrentHashMap.newKeySet());
+		return Flux.fromIterable(this.sessions.values())
+			.filter(session -> !session.hasOpenStream() && !active.contains(session.getId()))
+			.filter(session -> this.sessions.remove(session.getId(), session))
```
chemicL: perhaps collapse these?

mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStreamableServerTransportProvider.java
```diff
@@ -165,15 +191,95 @@ private HttpServletStreamableServerTransportProvider(McpJsonMapper jsonMapper, S
...
+			.flatMap(session -> {
+				logger.debug("Evicting idle session {}", session.getId());
+				return session.closeGracefully()
```
chemicL: Add timeout?

mcp-test/src/test/java/io/modelcontextprotocol/server/HttpServletStreamableIntegrationTests.java
```diff
@@ -247,26 +271,305 @@ void replacedListeningStreamIsClosed() throws Exception {
...
+			.build();
+		MCP_SERVLET.setDelegate(mcpServerTransportProvider);
+		prepareAsyncServerBuilder().serverInfo("test-server", "1.0.0").build();
+
+		var httpClient = HttpClient.newHttpClient();
+		var sessionId = initializeSession(httpClient);
+
+		// The session holds no stream and stays silent for long enough that a scheduled
+		// sweeper would have evicted it several times over
+		Thread.sleep(SESSION_SWEEP_INTERVAL.multipliedBy(4).toMillis());
```
chemicL: I wish we didn't have to sleep at all in tests. Any chance to virtualize the scheduler and just move the time? Every test that does some sleeping even if it's 150ms adds up (and we do have plenty of tests that sleep) and eventually a suite that should finish in under a second takes long enough to encourage contributors to step away from the keyboard breaking the flow.

Kehrlann: Switched everything to VirtualTimeScheduler

PR #1139, opened 2026-09-20
feat: add X-Accel-Buffering: no header to SSE responses

### LINKED ISSUE #293: Add X-Accel-Buffering: no Header for SSE Responses
### Quick Search Confirmation
I searched existing issues in this repository and did not find a duplicate request for adding `X-Accel-Buffering` headers to SSE responses.

### Expected Behavior
When using the `spring-ai-starter-mcp-server-webmvc` module to handle Server-Sent Events (SSE), the framework should automatically inject the header `X-Accel-Buffering: no` into all SSE responses. This ensures proxy servers (e.g., Nginx) disable response buffering, allowing real-time streaming of AI-generated content (e.g., chat completions, streaming predictions) without latency.

**Example of Expected Headers**:
For an SSE endpoint like `/stream`, the response should include:
```
Content-Type: text/event-stream
X-Accel-Buffering: no
```

### Current Behavior
- **Missing Header**: SSE responses from `spring-ai-starter-mcp-server-webmvc` do not include `X-Accel-Buffering: no`.
- **Consequence**: Proxy servers (e.g., Nginx) buffer SSE responses by default, causing:
  - Delayed delivery of real-time data (e.g., chat messages appear in batches instead of streaming).
  - Inconsistent behavior in AI applications where low latency is critical.

Debasishhh:
## Summary

Fixes #293.

None of the servlet-based SSE responses set `X-Accel-Buffering: no`. Without it, reverse proxies like Nginx buffer the response by default, batching what should be a real-time stream of JSON-RPC messages/notifications instead of forwarding each one as it's written.

## Changes

Added the header alongside the existing `Cache-Control`/`Connection` headers at all three places an SSE response is opened:

- `HttpServletSseServerTransportProvider#doGet` (the legacy SSE transport's connection endpoint)
- `HttpServletStreamableServerTransportProvider#doGet` (the streamable transport's GET listening stream)
- `HttpServletStreamableServerTransportProvider#doPost` (the streamable transport's SSE response to a streamed tool call)

## Tests

Added one integration test per call site above, each opening a real connection through a real Tomcat instance and asserting `X-Accel-Buffering: no` is present on the response. I verified all three fail without the header and pass with it (reverted the header lines locally, confirmed the exact assertion failures, then restored the fix).

## Verification

- `./mvnw -pl mcp-core -am install && ./mvnw -pl mcp-test -Dtest=HttpServletSseIntegrationTests,HttpServletStreamableIntegrationTests test` — 43 + 55 tests, 0 failures.
- `./mvnw -pl mcp-core,mcp-test -am spring-javaformat:validate` — clean.

## Related PR

#848 proposed the same header addition (2 of the 3 sites) but appears to have gone stale without review. This PR covers all three SSE response sites and adds test coverage.


ShemTovYosef:
Without this fix stream is not working in AWS API Gateway where backend proxy is ngingx

